PR #106, opened 2014-11-04
[FEATURE] Add removeDisplayNone option + tests

timothyasp:
This adds an options parameter and some tests for a `removeDisplayNone` option
which specifies if the Emogrifier parser will remove elements with the style
"display: none".

Unfortunatly, that nice little optimization had some consequences.

When writing responsive email templates, "display: none" is used often to hide
certain sections of the template for desktop email clients and show them in
phone clients.

I can see how it'd be nice to clean up elements that will not be displayed in
some cases, so I made it an option.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -101,6 +101,11 @@ class Emogrifier {
     private $styleAttributesForNodes = array();
 
     /**
+     * Remove elements from the dom that have display: none set.
+     */
+    private $removeDisplayNone = TRUE;
+
+    /**
      * This attribute applies to the case where you want to preserve your original text encoding.
      *
      * By default, emogrifier translates your text into HTML entities for two reasons:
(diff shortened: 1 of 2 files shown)


oliverklee:
Ping … could you please rebase, re-push and then comment on this PR? Thanks!

Classes/Emogrifier.php
```diff
@@ -101,6 +101,11 @@ class Emogrifier {
     private $styleAttributesForNodes = array();
 
     /**
+     * Remove elements from the dom that have display: none set.
+     */
+    private $removeDisplayNone = TRUE;
```
erikhansen: `TRUE` should be changed to `true`

erikhansen:
@oliverklee Based on the recent pull request (#156) that added a couple of additional options, don't you think that this PR should be modified to follow the same pattern as the new options that have been added? Here is what I would recommend:

* Change `setOptions` method to `disableRemoveDisplayNone` and set the value of the property directly. See: https://github.com/jjriv/emogrifier/blob/master/Classes/Emogrifier.php#L325
* Update the [Options](https://github.com/jjriv/emogrifier#options) section of the `README.md` file to cover the new option

oliverklee:
Yes, absolutely.

(And we should flip the option to "keepDisplayNone" so true does not mean "no display: none" and false does not mean "keep remove: none". This will make the code easier to understand.)

timothyasp:
Closing this pull and opening a new one based off the new master with the suggested changes.

PR #156, opened 2015-05-09
[FEATURE] Allow disabling of "inline style" and "style block" parsing

erikhansen:
This pull request implements the functionality described in task #155.

Please review how I edited the `emogrify()` method to disable "inline style" parsing to ensure it was done properly.

I updated the readme and added unit tests to cover the new options that were added.

Once you merge in this pull request, would you mind tagging a 0.1.1 release?

PR #157, opened 2015-05-09
[FEATURE] Added enablePreserveEncoding option

erikhansen:
This `task/method-for-preserving-encoding` branch branches off of the `task/disable-style-parsing branch`, as it is a direct response to the methods that were added in that branch. Please merge in the pull request for that branch (#156) in order to see the diff of what this branches adds. Or, you can simply look at a diff between the two branches: https://github.com/classyllama/emogrifier/compare/task/disable-style-parsing...classyllama:task/method-for-preserving-encoding

Here is an explanation of the changes:

The public `preserveEncoding` property already existed, but since the
`isInlineStylesParsingEnabled and `isStyleBlocksParsingEnabled` methods
were recently added, I thought it made sense for all options to be set via
methods, rather than a mix of methods + properties.

A case could be made for changing the `preserveEncoding` method to a private
method, but it would break backwards compatibility. So maybe stick this in
as a TODO for a minor version.

Since the preservation of encoding can be enabled via either the method or
the property, I updated the unit tests to account for both options.

Updated documentation to reflect the usage of this method.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -9,6 +9,7 @@ namespace Pelago;
  * @author Cameron Brooks
  * @author Jaime Prado
  * @author Roman Ožana <ozana@omdesign.cz>
+ * @author Erik Pall Hansen
  */
 class Emogrifier
 {
@@ -335,6 +336,16 @@ class Emogrifier
     public function disableStyleBlocksParsing()
     {
         $this->isStyleBlocksParsingEnabled = false;
+    }
+
+    /**
+     * Enables the preservation of encoding
+     *
+     * return @void
+     */
+    public function enablePreserveEncoding()
+    {
+        $this->preserveEncoding = true;
     }
 
     /**
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ styles into inline style attributes in your HTML code.
 - [Usage](#usage)
 - [Installing with Composer](#installing-with-composer)
 - [Usage](#usage)
+- [Options](#options)
 - [Supported CSS selectors](#supported-css-selectors)
 - [Caveats](#caveats)
 - [Maintainer](#maintainer)
@@ -78,6 +79,16 @@ calling the `emogrify` method:
   preserves all of the "style" attributes on tags in the HTML you pass to it.
   However if you want to discard all existing inline styles in the HTML before
   the CSS is applied, you should use this option.
+* `$emogrifier->enablePreserveEncoding()` - By default, Emogrifier translates 
+  your text into HTML entities for two reasons: 
+  
+  * Because of client incompatibilities, it is better practice to send out HTML
+    entities.
+  
+  * It translates any illegal XML characters that DOMDocument cannot work with.
+  
+  If you would like to preserve your original encoding, you should use this 
+  option.
 
 
 ## Installing with Composer
--- a/Tests/Unit/EmogrifierTest.php
+++ b/Tests/Unit/EmogrifierTest.php
@@ -90,7 +90,24 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
     /**
      * @test
      */
-    public function emogrifyCanKeepEncodedUmlauts()
+    public function emogrifyCanKeepEncodedUmlautsUsingMethod()
+    {
+        $this->subject->enablePreserveEncoding();
+        $encodedString = 'Küss die Hand, schöne Frau.';
+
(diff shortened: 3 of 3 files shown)


oliverklee:
I've fixed the PHP_CodeSniffer warnings on Travis. Could you please rebase and re-push? Thanks!

oliverklee:
Could you also please squash the commits into one commit and use [FEATURE] instead of [TASK] (as this is a new feature)?

oliverklee:
(I'll have a look at the code after the re-push.)

oliverklee:
And if you like, you can also add yourself to the authors in the file (using the author annotation).

erikhansen:
@oliverklee - I've rebased my commit on to `task/disable-style-parsing`, so this is ready for your review. As mentioned initially, it would be best if you merged `task/disable-style-parsing` into master before reviewing this PR. I changed the commit tag to [FEATURE], although I left the "task/" label in the tag name, as I wasn't sure if I could change it.

Thanks, I added my name to the authors list.

oliverklee:
I've just merged #156. Could you please rebase?

erikhansen:
@oliverklee Thanks for merging #156 and for all of your excellent help.

I've rebased this commit as well. I also made some changes to the commit based on your feedback from #156.

Once you've merged this branch, would you mind tagging a 0.1.1 release? That would be really helpful, as we're using Composer to include Emogrifier and really need to be able to target a version, rather than @dev.

oliverklee:
Actually, I'd prefer Emogrifier to not change umlauts etc. to HTML entities (thus a new default behavior), and to not have any option to change this behavior. @erikhansen What do you think?

erikhansen:
I'm going to do some testing with Litmus to see how email clients handle each option and I'll get back with you.

oliverklee:
Great, thanks!

erikhansen:
I grabbed the umlaut text that you used in the unit test as well as some (popular unicode characters)[http://unicode-table.com/en/sets/popular/]. I then sent a test email: the first two lines are with the unicode characters converted to html entities, the last two lines are the unicode entities themselves:

    K&uuml;ss die Hand, sch&ouml;ne Frau.
    &#10084; &#9806; &#9728; &#9733; &#9730; &#9822; &#9775; &#9773; &#9762; &euro; &#9742; &#9873; &#10052; &#9835; &#9986;
    
    Küss die Hand, schöne Frau.
    ❤ ♎ ☀ ★ ☂ ♞ ☯ ☭ ☢ € ☎ ⚑ ❄ ♫ ✂

Here are the Litmus results: https://litmus.com/pub/743eb3e So as you can see, the tested email clients all supported the unicode characters. And in the case of the Blackberry 4 OS client, the unicode characters worked better, as Blackberry displayed the HTML email as plain text. Unicode characters will likely be better (supported on devices like the Apple watch)[https://litmus.com/blog/apple-watch-favors-text-version-breaks-links]).

So based on email client support, I agree with your suggestion to remove the character encoding.

However the docblock also mentioned this a benefit of character encoding: _It translates any illegal XML characters that DOMDocument cannot work with._ I've only spent a little bit of time researching this, but we may want to add a method that will remove any illegal characters. Something like this: http://stackoverflow.com/a/3466609/2726785 Let me know if you have thoughts on this, otherwise I'll dig into it more and recommend a solution.

oliverklee:
Thanks for testing!

I propose adding a separate PR for the feature of either removing illegal characters or throwing an exception if illegal characters are present. (I'd prefer the latter so potential problems get noticed early.)

erikhansen:
I agree with adding a new PR. I think we would want to have an option as to whether to throw an exception if illegal characters are determined. Because in a production environment, I would think you would want to just strip the illegal characters, especially since emails will commonly contain user-submitted information, such as their name. But in a development environment, I think you would definitely want to throw an exception.

I'm not sure that I'll have time to create a PR to remove illegal characters, at least not for a week or two. What do you want me to do with this PR? Just close it, since this option (and the related functionality) will be stripped out?

oliverklee:
Yes, let's close this PR and then have new PRs for other changes.

erikhansen:
Ok.

Any chance you could tag a 0.1.1 release?

oliverklee:
Can do. Would you like to have the tag now or with a to-come "don't encode" PR merged in?

erikhansen:
@oliverklee The to-come "don't encode" functionality is actually not immediately relevant to the project I'm working on, so a tag now would be ideal.

oliverklee:
Ok, I've just tagged and released V0.1.1.

PR #162, opened 2015-05-15
[FEATURE] Add isRemoveDisplayNoneEnabled option + tests

timothyasp:
This adds an options parameter and some tests for a `isRemoveDisplayNoneEnabled` option
which specifies if the Emogrifier parser will remove elements with the style
"display: none".

Unfortunatly, the nice little optimization has some consiquences.

When writing responsive email templates, "display: none" is used often to hide
certain sections of the template for desktop email clients and show them in
phone clients.

I can see how it'd be nice to clean up elements that will not be displayed in
some cases, so that's why I made it an options field.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -122,6 +122,14 @@ class Emogrifier
     private $isStyleBlocksParsingEnabled = true;
 
     /**
+     * Determines whether elements with the `display: none` property are
+     * removed from the DOM.
+     *
+     * @var bool
+     */
+    private $shouldKeepInvisibleNodes = true;
+
+    /**
      * This attribute applies to the case where you want to preserve your original text encoding.
      *
      * By default, emogrifier translates your text into HTML entities for two reasons:
@@ -289,23 +297,8 @@ class Emogrifier
             $this->fillStyleAttributesWithMergedStyles();
         }
 
-        // This removes styles from your email that contain display:none.
-        // We need to look for display:none, but we need to do a case-insensitive search. Since DOMDocument only
-        // supports XPath 1.0, lower-case() isn't available to us. We've thus far only set attributes to lowercase,
-        // not attribute values. Consequently, we need to translate() the letters that would be in 'NONE' ("NOE")
-        // to lowercase.
-        $nodesWithStyleDisplayNone = $xpath->query(
-            '//*[contains(translate(translate(@style," ",""),"NOE","noe"),"display:none")]'
-        );
-        // The checks on parentNode and is_callable below ensure that if we've deleted the parent node,
-        // we don't try to call removeChild on a nonexistent child node
-        if ($nodesWithStyleDisplayNone->length > 0) {
-            /** @var \DOMNode $node */
-            foreach ($nodesWithStyleDisplayNone as $node) {
-                if ($node->parentNode && is_callable(array($node->parentNode,'removeChild'))) {
-                    $node->parentNode->removeChild($node);
-                }
-            }
+        if ($this->shouldKeepInvisibleNodes) {
+            $this->removeInvisibleNodes($xpath);
         }
 
         $this->copyCssWithMediaToStyleNode($cssParts, $xmlDocument);
@@ -335,6 +328,16 @@ class Emogrifier
     public function disableStyleBlocksParsing()
     {
         $this->isStyleBlocksParsingEnabled = false;
+    }
+
+    /**
+     * Disables the removal of elements with `display: none` properties.
+     *
+     * @return void
+     */
+    public function disableInvisibleNodeRemoval()
+    {
+        $this->shouldKeepInvisibleNodes = false;
     }
 
     /**
@@ -415,6 +418,36 @@ class Emogrifier
         $key = array_search($tagName, $this->unprocessableHtmlTags, true);
         if ($key !== false) {
             unset($this->unprocessableHtmlTags[$key]);
+        }
+    }
+
+     /**
+      * This removes styles from your email that contain display:none.
+      * We need to look for display:none, but we need to do a case-insensitive search. Since DOMDocument only
+      * supports XPath 1.0, lower-case() isn't available to us. We've thus far only set attributes to lowercase,
+      * not attribute values. Consequently, we need to translate() the letters that would be in 'NONE' ("NOE")
+      * to lowercase.
+      *
+      * @param \DOMXPath $xpath
+      *
+      * @return void
+      */
+    private function removeInvisibleNodes(\DOMXPath $xpath)
+    {
+        $nodesWithStyleDisplayNone = $xpath->query(
+            '//*[contains(translate(translate(@style," ",""),"NOE","noe"),"display:none")]'
+        );
+        if ($nodesWithStyleDisplayNone->length === 0) {
+            return;
+        }
+
+        // The checks on parentNode and is_callable below ensure that if we've deleted the parent node,
+        // we don't try to call removeChild on a nonexistent child node
+        /** @var \DOMNode $node */
+        foreach ($nodesWithStyleDisplayNone as $node) {
+            if ($node->parentNode && is_callable(array($node->parentNode, 'removeChild'))) {
+                $node->parentNode->removeChild($node);
+            }
         }
     }
 
--- a/Tests/Unit/EmogrifierTest.php
+++ b/Tests/Unit/EmogrifierTest.php
@@ -1090,4 +1090,64 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
             $this->subject->emogrify()
         );
     }
+
+    /**
+     * @test
+     */
+    public function emogrifyByDefaultRemovesElementsWithDisplayNoneFromExternalCss()
+    {
+        $css = 'div.foo { display: none; }';
+        $html = $this->html5DocumentType . self::LF .
+            '<html><body><div class="bar"></div><div class="foo"></div></body></html>';
+
+        $expected = '<div class="bar"></div>';
+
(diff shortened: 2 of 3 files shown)


Classes/Emogrifier.php
```diff
@@ -122,6 +122,12 @@ class Emogrifier
     private $isStyleBlocksParsingEnabled = true;
 
     /**
+     * Determines whether elements with the `display: none` property are
+     * removed from the DOM.
+     */
+    private $isRemoveDisplayNoneEnabled = true;
```
oliverklee: Actually, the option name is a bit misleading. As it is about removing/keeping nodes that are display: none (and not about removing/keeping the display: none itself), I propose the name "shouldKeepInvisibleNodes". What do you think?

Classes/Emogrifier.php
```diff
@@ -122,6 +122,12 @@ class Emogrifier
     private $isStyleBlocksParsingEnabled = true;
 
     /**
+     * Determines whether elements with the `display: none` property are
+     * removed from the DOM.
+     */
```
oliverklee: Please add the missing @var bool.

Classes/Emogrifier.php
```diff
@@ -294,16 +300,18 @@ public function emogrify()
...
-        );
-        // The checks on parentNode and is_callable below ensure that if we've deleted the parent node,
-        // we don't try to call removeChild on a nonexistent child node
-        if ($nodesWithStyleDisplayNone->length > 0) {
-            /** @var \DOMNode $node */
-            foreach ($nodesWithStyleDisplayNone as $node) {
-                if ($node->parentNode && is_callable(array($node->parentNode,'removeChild'))) {
-                    $node->parentNode->removeChild($node);
+        if ($this->isRemoveDisplayNoneEnabled) {
+            $nodesWithStyleDisplayNone = $xpath->query(
```
oliverklee: Please let's move the code within the IF to a separate, well-named private method.

Classes/Emogrifier.php
```diff
@@ -294,16 +300,18 @@ public function emogrify()
...
+                '//*[contains(translate(translate(@style," ",""),"NOE","noe"),"display:none")]'
+            );
+            // The checks on parentNode and is_callable below ensure that if we've deleted the parent node,
+            // we don't try to call removeChild on a nonexistent child node
+            if ($nodesWithStyleDisplayNone->length > 0) {
+                /** @var $node \DOMNode */
+                foreach ($nodesWithStyleDisplayNone as $node) {
+                    if ($node->parentNode && is_callable(array($node->parentNode,'removeChild'))) {
```
oliverklee: Please add the missing space after the comma.

Tests/Unit/EmogrifierTest.php
```diff
@@ -1090,4 +1090,45 @@ public function emogrifyMergesCssWithMixedUnits()
             $this->subject->emogrify()
         );
     }
+
+    /**
+     * @test
+     */
+    public function emogrifyRemovesDisplayNoneElements()
```
oliverklee: Please add the precondition for this behavior to the test name (also for the other test): emogrifyFor...RemovesInvisibleElement

Tests/Unit/EmogrifierTest.php
```diff
@@ -1090,4 +1090,45 @@ public function emogrifyMergesCssWithMixedUnits()
...
+    {
+        $css = 'div.foo { display: none; }';
```
oliverklee: We also should have tests for both cases: When the display: none comes from a style node, and when it comes from a style attribute of the element.

Classes/Emogrifier.php
timothyasp: Sure, that sounds fine.

timothyasp:
@oliverklee I think i've made all the requested changes.  Thanks for the feedback!

Classes/Emogrifier.php
```diff
@@ -338,6 +331,16 @@ public function disableStyleBlocksParsing()
     }
 
     /**
+     * Disables the removal of elements with `display: none` properties
```
oliverklee: Please add the period at the end of the sentence.

Classes/Emogrifier.php
```diff
@@ -418,6 +421,34 @@ public function removeUnprocessableHtmlTag($tagName)
...
+      */
+    private function removeInvisibleNodes(\DOMXPath $xpath)
+    {
+        $nodesWithStyleDisplayNone = $xpath->query(
+            '//*[contains(translate(translate(@style," ",""),"NOE","noe"),"display:none")]'
+        );
+        // The checks on parentNode and is_callable below ensure that if we've deleted the parent node,
+        // we don't try to call removeChild on a nonexistent child node
+        if ($nodesWithStyleDisplayNone->length > 0) {
+            /** @var $node \DOMNode */
```
oliverklee: Please let's change the order of the elements within this @var comment: First the type, then the name. This will make the comment consistent with the other @var comments for local variables.

Classes/Emogrifier.php
```diff
@@ -418,6 +421,34 @@ public function removeUnprocessableHtmlTag($tagName)
...
+      * @return void
+      */
+    private function removeInvisibleNodes(\DOMXPath $xpath)
+    {
+        $nodesWithStyleDisplayNone = $xpath->query(
+            '//*[contains(translate(translate(@style," ",""),"NOE","noe"),"display:none")]'
+        );
+        // The checks on parentNode and is_callable below ensure that if we've deleted the parent node,
+        // we don't try to call removeChild on a nonexistent child node
+        if ($nodesWithStyleDisplayNone->length > 0) {
```
oliverklee: Now that this is within a separate nice method, we can change this to a guard clause:

if ... === 0 {
  return;
}

This will allow us to drop one level of indentation for the following code. Could you please do this?

Tests/Unit/EmogrifierTest.php
```diff
@@ -1090,4 +1090,46 @@ public function emogrifyMergesCssWithMixedUnits()
...
+        $this->assertContains(
+            $expected,
+            $this->subject->emogrify()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyPreservesInvisibleElements()
```
oliverklee: Please add the precondition to the test name:

emogrifyAfterDisableInvisibleNodeRemovalPreservesInvisibleElements

Tests/Unit/EmogrifierTest.php
```diff
@@ -1090,4 +1090,46 @@ public function emogrifyMergesCssWithMixedUnits()
...
     }
+
+    /**
+     * @test
+     */
+    public function emogrifyRemovesInvisibleElements()
+    {
+        $css = 'div.foo { display: none; }';
+        $html = $this->html5DocumentType . self::LF .
+            '<html><body><div class="bar"></div><div class="foobar" style="display: none;"></div>'
```
oliverklee: Actually, these should be two separate tests: one for the case that the display: none comes from the CSS via setCss, and one for the case that the display: none comes from the style attribute. I propose the following two test method names:

emogrifyByDefaultRemovesElementsWithDisplayNoneInStyleAttribute
emogrifyByDefaultRemovesElementsWithDisplayNoneFromExternalCss

What do you think?

oliverklee:
Could you also please write a short documentation for this new feature for the README?

oliverklee:
ping

timothyasp:
Made the requested changes.  The test failure looks Github related?  Travis couldn't connect to github.  I'm not sure how to requeue the test - @oliverklee could you do that?

oliverklee:
I've run the tests on this branch locally (as long as Travis does not play along). On your branch, I get the following failure:

    klee@gonzales:~/src/php/emogrifier$ vendor/bin/phpunit Tests/
    PHPUnit 4.6.7 by Sebastian Bergmann and contributors.
    
    ...............................................................  63 / 155 ( 40%)
    ............................................................... 126 / 155 ( 81%)
    ...........................F.
    
    Time: 74 ms, Memory: 5.00Mb
    
    There was 1 failure:
    
    1) Pelago\Tests\Unit\EmogrifierTest::emogrifyByDefaultRemovesElementsWithDisplayNoneInStyleAttribute
    Failed asserting that '<!DOCTYPE html>
    <html><body>
    <div class="bar"></div>
    <div class="foo"></div>
    </body></html>
    ' contains "<div class="bar"></div><div class="foo"></div>".
    
    /home/klee/src/php/emogrifier/Tests/Unit/EmogrifierTest.php:1130
    
    FAILURES!
    Tests: 155, Assertions: 155, Failures: 1.

Do the tests run fine on your local machine?

oliverklee:
And could you please prefix the commit message subject line with [FEATURE]?

Classes/Emogrifier.php
```diff
@@ -418,6 +421,37 @@ public function removeUnprocessableHtmlTag($tagName)
...
+      * @param \DOMXPath $xpath
+      *
+      * @return void
+      */
+    private function removeInvisibleNodes(\DOMXPath $xpath)
+    {
+        $nodesWithStyleDisplayNone = $xpath->query(
+            '//*[contains(translate(translate(@style," ",""),"NOE","noe"),"display:none")]'
+        );
+
```
oliverklee: Please remove this blank line so that the guard clause is a single visible block.

timothyasp:
I address all your concerns - still looks like Travis isn't working with github.

oliverklee:
I've tested the PR locally, fixed the unit tests (the DIVs were not nested, which made the assertContains not find any potential problems reliably), merged locally, and pushed.

Thanks for contributing, @timothyasp! :heart:

Issue #90, opened 2014-04-25
403 errors in Travis during composer install


oliverklee:

oliverklee:
We have this problem again. Sigh.

(Actually, we have this problem when there are more than a few builds per day.)

oliverklee:
https://translate.google.com/translate?sl=auto&tl=en&js=y&prev=_t&hl=de&ie=UTF-8&u=http%3A%2F%2Fblog.champs-libres.coop%2Fdev%2F2015%2F01%2F30%2Fauthenticate-against-github-travis-composer.html&edit-text=&act=url

oliverklee:
It looks like we actually need to create the oauth token (#76).

oliverklee:
I've opened a ticket for Travis CI: https://github.com/travis-ci/travis-ci/issues/4073

jjriv:
Thanks. Let me know if there is anything I can do to help.

____________________________________________________________________

John Reeve
jreeve@pelagodesign.com<mailto:jreeve@pelagodesign.com>
Intervals | Web-based Time, Task & Project Management | www.myintervals.com<http://www.myintervals.com>
____________________________________________________________________

oliverklee:
This should be fixed by #223.

PR #165, opened 2015-05-27
[BUGFIX] Avoid GitHub authentication failures on Travis

oliverklee:
By providing the --no-interaction option for composer install,
composer will do a git clone if the API limit is exhausted.

Fixes: #90


oliverklee:
It looks like this did not really fix the problem. *sigh*

Issue #172, opened 2015-06-07
Investigate whether we can get rid of any re-encoding altogether


oliverklee:
i.e., whether we can drop the ENCODING constant.

Renkas:
I'll add the same info here also:

```php
return mb_convert_encoding($xmlDocument->saveHTML(), self::ENCODING, 'HTML-ENTITIES');
```
This row is messing up my HTML. It double encodes the entities I already have in my (XHTML) document. If I return $xmlDocument->saveHTML() directly then everything seems to be fine. All entities that were not encoded in input are already encoded by DOMDocument.

Right now in my case the code is not working as expected.

PR #175, opened 2015-06-16
Properly handle UTF-8 characters and entities

Renkas:
This pull request will make handling of character entities work properly. As this is something that is dependant on PHPs DOMDocument then basically all that you can do is to add a note in the documentation that input (X)HTML needs to include meta tag with UTF-8 as charset. Otherwise DOMDocument will re-encode all the entities and characters in the document.

I removed the workaround put in to the code as it only made things worse. Added some test cases and added a note to README.md

Partially connected to #172

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -15,11 +15,6 @@ namespace Pelago;
 class Emogrifier
 {
     /**
-     * @var string
-     */
-    const ENCODING = 'UTF-8';
-
-    /**
      * @var int
      */
     const CACHE_KEY_CSS = 0;
@@ -130,6 +125,13 @@ class Emogrifier
      * @var bool
      */
     private $shouldKeepInvisibleNodes = true;
+
+    /**
+     * Determines whether Content-Type meta tag with UTF-( charset should be inserted to HTML if it is missing
+     *
+     * @var bool
+     */
+    private $shouldFixMissingCharsetDeclaration = true;
 
     /**
      * The constructor.
@@ -320,6 +322,16 @@ class Emogrifier
     public function disableInvisibleNodeRemoval()
     {
         $this->shouldKeepInvisibleNodes = false;
+    }
+
+    /**
+     * Disables the removal of elements with `display: none` properties.
+     *
+     * @return void
+     */
+    public function disableCharsetFix()
+    {
+        $this->shouldFixMissingCharsetDeclaration = false;
     }
 
     /**
@@ -649,7 +661,6 @@ class Emogrifier
     private function createXmlDocument()
     {
         $xmlDocument = new \DOMDocument;
-        $xmlDocument->encoding = self::ENCODING;
         $xmlDocument->strictErrorChecking = false;
         $xmlDocument->formatOutput = true;
         $libXmlState = libxml_use_internal_errors(true);
@@ -662,7 +673,8 @@ class Emogrifier
     }
 
     /**
-     * Returns the HTML with the unprocessable HTML tags removed.
+     * Returns the HTML with the unprocessable HTML tags removed and
+     * with added Content-Type meta tag if needed.
      *
      * @return string the unified HTML
      *
@@ -679,6 +691,20 @@ class Emogrifier
             );
         } else {
             $bodyWithoutUnprocessableTags = $this->html;
+        }
+        
+        if ($this->shouldFixMissingCharsetDeclaration) {
+            if (!stristr($bodyWithoutUnprocessableTags, 'Content-Type')) {
+
+                // We are trying to insert the meta tag to the right spot in the DOM as if we just prepend it to the HTML we will lose attributes set to the HTML tag.
+                if (stristr($bodyWithoutUnprocessableTags, '<head')) {
+                    $bodyWithoutUnprocessableTags = preg_replace('/<head(.*?)>/i', '<head$1><meta http-equiv="Content-Type" content="text/html; charset=utf-8">', $bodyWithoutUnprocessableTags);
+                } elseif (stristr($bodyWithoutUnprocessableTags, '<html')) {
+                    $bodyWithoutUnprocessableTags = preg_replace('/<html(.*?)>/i', '<html$1><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head>', $bodyWithoutUnprocessableTags);
+                } else {
+                    $bodyWithoutUnprocessableTags = '<meta http-equiv="Content-Type" content="text/html; charset=utf-8">' . $bodyWithoutUnprocessableTags;
+                }
+            }
         }
 
         return $bodyWithoutUnprocessableTags;
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -289,7 +289,7 @@ class Emogrifier
 
         $this->copyCssWithMediaToStyleNode($cssParts, $xmlDocument);
 
-        return mb_convert_encoding($xmlDocument->saveHTML(), self::ENCODING, 'HTML-ENTITIES');
+        return $xmlDocument->saveHTML();
     }
 
     /**
@@ -662,8 +662,7 @@ class Emogrifier
     }
 
     /**
-     * Returns the HTML with the non-ASCII characters converts into HTML entities and the unprocessable
-     * HTML tags removed.
+     * Returns the HTML with the unprocessable HTML tags removed.
      *
      * @return string the unified HTML
      *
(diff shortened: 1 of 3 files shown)


Classes/Emogrifier.php
```diff
@@ -289,7 +289,7 @@ public function emogrify()
 
         $this->copyCssWithMediaToStyleNode($cssParts, $xmlDocument);
 
-        return mb_convert_encoding($xmlDocument->saveHTML(), self::ENCODING, 'HTML-ENTITIES');
+        return $xmlDocument->saveHTML();
```
oliverklee: I think the constant ENCODING then is not used anymore either, and we can remove it as well.

README.md
```diff
@@ -63,6 +63,10 @@ After you have set the HTML and CSS, you can call the `emogrify` method to merge
 
     $mergedHtml = $emogrifier->emogrify();
 
+NB! for Emogrifier to properly deal with entities and UTF-8 characters you need to add Content-Type meta tag with UTF-8 charset declaration to your HTML.
+
+    $html = '<html><meta http-equiv="Content-Type" content="text/html; charset=utf-8"><h1>Hello world!</h1></html>';
```
oliverklee: Can't we force Emogrifier to always handle the HTML as UTF-8 instead, thus making the meta tag unnecessary?

Tests/Unit/EmogrifierTest.php
```diff
@@ -96,7 +96,39 @@ public function emogrifyKeepsUtf8Umlauts()
...
+        $this->assertContains(
+            $umlautString,
+            $this->subject->emogrify()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyKeepsUtf8UmlautsXhtml()
```
oliverklee: Could you please add an "In" before the "Xhtml"?

Tests/Unit/EmogrifierTest.php
```diff
@@ -96,7 +96,39 @@ public function emogrifyKeepsUtf8Umlauts()
...
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyKeepsUtf8UmlautsHtml4()
+    {
+        $umlautString = 'Öösel läks õunu täis ämber uhkelt ümber.';
+
+        $html = $this->html4TransitionalDocumentType . '<html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head><p>' . $umlautString . '</p></html>';
```
oliverklee: Could you please also add a test checking that required HTML entities (e.g. & amp ;) are kept?

Tests/Unit/EmogrifierTest.php
```diff
@@ -96,7 +96,39 @@ public function emogrifyKeepsUtf8Umlauts()
...
+        $this->assertContains(
+            $umlautString,
+            $this->subject->emogrify()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyKeepsUtf8UmlautsHtml4()
```
oliverklee: Could you please add an "In" before the "Html"?

oliverklee:
Could you please prefix the commit message subject line with "[BUGFIX]" and add the following line to the commit message description (so merging this PR will autoclose that ticket)?

Fixes #172

Renkas:
I did some changes.

I removed ENCODING constant
Emogrifier now tries to correct the missing meta tag automatically.
Also added option to disable this.
Added a test for &amp; also - and it passes like it should.

Classes/Emogrifier.php
```diff
@@ -649,7 +661,6 @@ function ($matches) use (&$media) {
     private function createXmlDocument()
     {
         $xmlDocument = new \DOMDocument;
-        $xmlDocument->encoding = self::ENCODING;
```
oliverklee: Doesn't the encoding field have any effect? Or could we avoid having the meta tag if we keep using $xmlDocument->encoding?

Classes/Emogrifier.php
```diff
@@ -681,8 +692,22 @@ private function getUnifiedHtml()
         } else {
             $bodyWithoutUnprocessableTags = $this->html;
         }
+        
+        if ($this->shouldFixMissingCharsetDeclaration) {
```
oliverklee: Please lets move this block (of the block in the inner or outer if) into a well-named, private method.

Classes/Emogrifier.php
```diff
@@ -681,8 +692,22 @@ private function getUnifiedHtml()
...
+            if (!stristr($bodyWithoutUnprocessableTags, 'Content-Type')) {
+
+                // We are trying to insert the meta tag to the right spot in the DOM as if we just prepend it to the HTML we will lose attributes set to the HTML tag.
+                if (stristr($bodyWithoutUnprocessableTags, '<head')) {
```
oliverklee: For stristr checks, please always use !== false as the return value isn't always a boolean.

Classes/Emogrifier.php
```diff
@@ -681,8 +692,22 @@ private function getUnifiedHtml()
...
+                    $bodyWithoutUnprocessableTags = preg_replace('/<head(.*?)>/i', '<head$1><meta http-equiv="Content-Type" content="text/html; charset=utf-8">', $bodyWithoutUnprocessableTags);
+                } elseif (stristr($bodyWithoutUnprocessableTags, '<html')) {
+                    $bodyWithoutUnprocessableTags = preg_replace('/<html(.*?)>/i', '<html$1><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head>', $bodyWithoutUnprocessableTags);
```
oliverklee: We have duplicate code here (at least the meta tag). Please move it to a local variable.

README.md
```diff
@@ -63,6 +63,11 @@ After you have set the HTML and CSS, you can call the `emogrify` method to merge
 
     $mergedHtml = $emogrifier->emogrify();
 
+For Emogrifier to properly deal with entities and UTF-8 characters you need to add Content-Type meta tag with UTF-8 charset
+declaration to your HTML. If you don't provide this tag in your HTML then Emogrifier adds it automatically.
```
oliverklee: Could you please reformat the lines in the README to be under 80 characters? This will avoid wrapping on GitHub.

README.md
```diff
@@ -63,6 +63,11 @@ After you have set the HTML and CSS, you can call the `emogrify` method to merge
 
     $mergedHtml = $emogrifier->emogrify();
 
+For Emogrifier to properly deal with entities and UTF-8 characters you need to add Content-Type meta tag with UTF-8 charset
```
oliverklee: I think there should be a comma after "characters".

README.md
oliverklee: I think there should be a comma after "HTML".

Classes/Emogrifier.php
```diff
@@ -323,6 +325,16 @@ public function disableInvisibleNodeRemoval()
     }
 
     /**
+     * Disables the removal of elements with `display: none` properties.
+     *
+     * @return void
+     */
+    public function disableCharsetFix()
```
oliverklee: We need unit tests for this method.

(Or, I'd prefer to always have this new behavior, and then we don't need that switch.)

oliverklee:
Could you please squash your commits?

oliverklee:
Could you also please rebase from master? I've merged some PRs.

Tests/Unit/EmogrifierTest.php
```diff
@@ -108,9 +108,41 @@ public function emogrifyKeepsUtf8Umlauts()
...
      * @test
      */
+    public function emogrifyKeepsUtf8UmlautsInXhtml()
+    {
+        $umlautString = 'Öösel läks õunu täis ämber uhkelt ümber. Also ampersand entity should remain unchanged: &amp;';
+
+        $html = $this->xhtml1StrictDocumentType . '<html style="padding: 1px;"><p>' . $umlautString . '</p></html>';
+        $this->subject->setHtml($html);
+
+        $this->assertContains(
```
oliverklee: Could you please change this to be a static call: self::assertContains

Tests/Unit/EmogrifierTest.php
```diff
@@ -108,9 +108,41 @@ public function emogrifyKeepsUtf8Umlauts()
...
+     * @test
+     */
+    public function emogrifyKeepsUtf8UmlautsInHtml4()
+    {
+        $umlautString = 'Öösel läks õunu täis ämber uhkelt ümber. Also ampersand entity should remain unchanged: &amp;';
+
+        $html = $this->html4TransitionalDocumentType . '<html><p>' . $umlautString . '</p></html>';
+        $this->subject->setHtml($html);
+
+        $this->assertContains(
```
oliverklee: Could you please change this to be a static call: self::assertContains

oliverklee:
Hi @Renkas, I've taken your PR, polished it an merged it as #189. Thanks a lot for contributing!

Issue #68, opened 2014-02-26
Add media print to the preserved media types


OzzyCzech:
Email tracking tools like https://litmus.com are using follow technique for tracking email prints:

    <style>
        @media print {
            #track { 
                background-image: url('http://www.letmeknow.com/print.jpg');
            }
        }
    </style>
    <div id="track"></div>

Current version from #62 PR strip this media query.

PR #178, opened 2015-06-19
[FEATURE] Allow user to define media types to keep and keep print type by default

Renkas:
I think print styles should be kept by default. But now user can manage himself what types to keep/drop.

Original #68

### DIFF
--- a/Tests/Unit/EmogrifierTest.php
+++ b/Tests/Unit/EmogrifierTest.php
@@ -634,7 +634,6 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
             'style in "braille" media type rule' => array('@media braille {p {color: #000;}}', '#000'),
             'style in "embossed" media type rule' => array('@media embossed {p {color: #000;}}', '#000'),
             'style in "handheld" media type rule' => array('@media handheld {p {color: #000;}}', '#000'),
-            'style in "print" media type rule' => array('@media print {p {color: #000;}}', '#000'),
             'style in "projection" media type rule' => array('@media projection {p {color: #000;}}', '#000'),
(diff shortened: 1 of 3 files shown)


oliverklee:
Hi @Renkas , thanks for the PR! Could you please squash the two commits and then rebase? I've just merged some PRs. Thanks!

Tests/Unit/EmogrifierTest.php
```diff
@@ -700,6 +700,42 @@ public function emogrifyKeepsMediaRules($css)
...
+    public function emogrifyRemoveAllowedMediaQueryType()
+    {
+        $css = '@media print { html {} }';
+
+        $html = $this->html5DocumentType . self::LF . '<html></html>';
+        $this->subject->setHtml($html);
+        $this->subject->setCss($css);
+        $this->subject->removeAllowedMediaType('print');
+
+        $this->assertNotContains(
```
oliverklee: Could you please make the assert call static (using self::)?

Tests/Unit/EmogrifierTest.php
```diff
@@ -700,6 +700,42 @@ public function emogrifyKeepsMediaRules($css)
...
+    public function emogrifyAddAllowedMediaQueryType()
+    {
+        $css = '@media braille { html {} }';
+
+        $html = $this->html5DocumentType . self::LF . '<html></html>';
+        $this->subject->setHtml($html);
+        $this->subject->setCss($css);
+        $this->subject->addAllowedMediaType('braille');
+
+        $this->assertContains(
```
oliverklee: Could you please make the assert call static (using self::)?

oliverklee:
@Renkas, I've polished your PR and merged it as #200. Thanks for contributing!

Issue #179, opened 2015-06-19
Errors with malformed CSS


Synchro:
I ran into two errors when processing some malformed style tags.

The style tag contained some stray HTML injected by ckeditor:

``` css
table.columns .right-text-pad {
  padding-right: 10px;
}
<style data-cke-temp="1">html{cursor:text;*cursor:auto}
img,input,textarea{cursor:default}
```

Trying to process this produced two errors. The first is caused by it choosing `<style data-cke-temp="1">html` as a selector, which will not convert to a valid xpath query:

```
2015-06-19 14:02:03 DOMXPath::query(): Invalid expression (error type 2 in ./vendor/pelago/emogrifier/Classes/Emogrifier.php on line 279)
```

The second is a consequence of this, attempting to iterate over the failed result of the query because the return value is not checked:

```
2015-06-19 14:02:03 Invalid argument supplied for foreach() (error type 2 in ./vendor/pelago/emogrifier/Classes/Emogrifier.php on line 282)
```

Obviously it would be better if bad CSS was not passed in in the first place, but it would be better to ignore it than throw errors.

oliverklee:
As for the cause of the error: I'm for having a whitelist of characters (i.e., a regular expression) which we can use to validate those selectors before using them.

Synchro:
I had a little look into matching selectors - it's not simple at all! I think the only invalid char in that selector is `<` - everything else is valid in other contexts.

oliverklee:
So I propose discarding any selectors that contain a "<". What do you think about this?

And would you be willing to create a PR for this?

Synchro:
Yes, I think that's reasonable. Where would I find the code that identifies selectors?

oliverklee:
That should be https://github.com/jjriv/emogrifier/blob/master/Classes/Emogrifier.php#L337 in the method parseSelectors.

Synchro:
As I've been reading, it's very hard to match CSS selectors with a regex as they are not a regular grammar. I'm thinking that it would be easier to apply a simple filter to remove stray tags before trying to parse the styles, for example:

    $css = preg_replace('/<[^>]*>/', '', $css);

While this is something of an edge case, it strikes me that this should be both harmless and effective as I can't think of any legitimate reason there should be HTML tags inside CSS.

JakeQZ:
[...]

PR #180, opened 2015-06-19
Ignore invalid selectors, see #179

Synchro:
This is a partial fix for #179. It simply adds a check on the xpath query return value before attempting to iterate over the result set. If the xpath has failed, it just issues a continue to go to the next iteration.

This fixes the second of the two errors mentioned in the ticket, so it's effectively treating the symptom rather than the cause, but it's better than nothing! I don't know how it would be best to try to prevent the first error - somehow validate the CSS selector fragment before trying to pass it to the xpath converter?

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -261,6 +261,10 @@ class Emogrifier
         foreach ($this->caches[self::CACHE_KEY_CSS][$cssKey] as $value) {
             // query the body for the xpath selector
             $nodesMatchingCssSelectors = $xpath->query($this->translateCssToXpath($value['selector']));
+            // Ignore invalid selectors
+            if (false === $nodesMatchingCssSelectors) {
+                continue;
+            }
 
             /** @var \DOMElement $node */
             foreach ($nodesMatchingCssSelectors as $node) {
--- a/Tests/Unit/EmogrifierTest.php
+++ b/Tests/Unit/EmogrifierTest.php
@@ -621,6 +621,29 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @test
+     */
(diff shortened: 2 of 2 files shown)


oliverklee:
Hi @Synchro, thanks for the PR!

Could you please add a regression test (i.e., a minimal unit test that fails without the fix and that passes with the fix)?

Could you also please prefix the commit message subject with [BUGFIX] (and, if you have more than one commit, squash those commits)? Thanks!

Classes/Emogrifier.php
```diff
@@ -262,6 +262,10 @@ public function emogrify()
             // query the body for the xpath selector
             $nodesMatchingCssSelectors = $xpath->query($this->translateCssToXpath($value['selector']));
 
+            // Ignore invalid selectors
+            if (!$nodesMatchingCssSelectors) {
```
oliverklee: This is not always a boolean. So we should not treat it as a boolean, but do an explicit === false instead.

Classes/Emogrifier.php
```diff
@@ -262,6 +262,10 @@ public function emogrify()
             // query the body for the xpath selector
             $nodesMatchingCssSelectors = $xpath->query($this->translateCssToXpath($value['selector']));
```
oliverklee: The blank line should be below the check, not above it.

oliverklee:
As for #179: Could you please add "Relates #179" to the commit message description part? Then #179 will  be autolinked (without closing it).

Synchro:
Done all those things, rebased, pushed. The test case was a little tricky - needed to create an error handler to trap the warnings. I've arranged it so that it eats the first error type, but lets the second go through. That can be removed when you need a test case for the other error.

oliverklee:
Hi @Synchro, I've polished your PR a bit in #194 and just merged it. Thanks for contributing!

PR #205, opened 2015-07-08
Handle important

vanderlee:
These changes allow Emogrifier to support `!important` like how browsers handle it.

Whenever a CSS expression in `<style>` tag ends with `!important`, it will overwrite any non-`!important` expressions in inline `style=""` attributes.

Unittests included to test both old and new behaviour.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -438,17 +438,17 @@ class Emogrifier
         }
     }
 
-     /**
-      * This removes styles from your email that contain display:none.
-      * We need to look for display:none, but we need to do a case-insensitive search. Since DOMDocument only
-      * supports XPath 1.0, lower-case() isn't available to us. We've thus far only set attributes to lowercase,
-      * not attribute values. Consequently, we need to translate() the letters that would be in 'NONE' ("NOE")
-      * to lowercase.
-      *
-      * @param \DOMXPath $xpath
-      *
-      * @return void
-      */
+    /**
+     * This removes styles from your email that contain display:none.
+     * We need to look for display:none, but we need to do a case-insensitive search. Since DOMDocument only
+     * supports XPath 1.0, lower-case() isn't available to us. We've thus far only set attributes to lowercase,
+     * not attribute values. Consequently, we need to translate() the letters that would be in 'NONE' ("NOE")
+     * to lowercase.
+     *
+     * @param \DOMXPath $xpath
+     *
+     * @return void
+     */
     private function removeInvisibleNodes(\DOMXPath $xpath)
     {
         $nodesWithStyleDisplayNone = $xpath->query(
@@ -530,6 +530,13 @@ class Emogrifier
     private function generateStyleStringFromDeclarationsArrays(array $oldStyles, array $newStyles)
     {
         $combinedStyles = array_merge($oldStyles, $newStyles);
+
+        foreach ($oldStyles as $attribute => $expression) {
+            if (isset($newStyles[$attribute]) && strtolower(substr($expression, -10)) === '!important') {
+                $combinedStyles[$attribute] = $expression;
+            }
+        }
+
         $style = '';
         foreach ($combinedStyles as $attributeName => $attributeValue) {
             $style .= (strtolower(trim($attributeName)) . ': ' . trim($attributeValue) . '; ');
--- a/Tests/Unit/EmogrifierTest.php
+++ b/Tests/Unit/EmogrifierTest.php
@@ -1519,4 +1519,77 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
             $this->subject->emogrify()
         );
     }
+
+    /**
+     * @test
+     */
+    public function emogrifyHandleNoImportant()
+    {
+        $css = 'p { margin: 1px; padding: 1px;}';
+        $html = $this->html5DocumentType . self::LF .
+            '<html><head</head><body><p style="margin: 2px; text-align: center;">some content</p></body></html>';
+        $expected = '<p style="margin: 2px; text-align: center; padding: 1px;">';
+        $this->subject->setHtml($html);
+        $this->subject->setCss($css);
+
+        self::assertContains(
+            $expected,
+            $this->subject->emogrify()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyHandleImportantStyleTag()
+    {
+        $css = 'p { margin: 1px !important; padding: 1px;}';
+        $html = $this->html5DocumentType . self::LF .
+            '<html><head</head><body><p style="margin: 2px; text-align: center;">some content</p></body></html>';
+        $expected = '<p style="margin: 1px !important; text-align: center; padding: 1px;">';
+        $this->subject->setHtml($html);
+        $this->subject->setCss($css);
+
+        self::assertContains(
+            $expected,
+            $this->subject->emogrify()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyHandleImportantStyleTagCaseInsensitive()
+    {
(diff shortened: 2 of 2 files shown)


oliverklee:
Hi @vanderlee, thanks for this PR! Could you please squash your commits into one, prefix the commit message subject with [FEATURE], and split the commit message into a subject line (no longer than 50 characters), a blank line and an optional description (word-wrapped at 79 characters)?

vanderlee:
How do I go about squashing the commits into one?

oliverklee:
Here are some instructions on squashing commits:

http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

vanderlee:
I'll look into it later. Ditto for my other pulls. I don't particularly enjoy all this bookkeeping work.

vanderlee:
Squashed!

vanderlee:
I must say, this squashing is a very error-prone endeavor.

oliverklee:
It is indeed as it includes changing history. ;-)

Classes/Emogrifier.php
```diff
@@ -438,17 +438,17 @@ public function removeAllowedMediaType($mediaName)
...
+     * This removes styles from your email that contain display:none.
+     * We need to look for display:none, but we need to do a case-insensitive search. Since DOMDocument only
+     * supports XPath 1.0, lower-case() isn't available to us. We've thus far only set attributes to lowercase,
+     * not attribute values. Consequently, we need to translate() the letters that would be in 'NONE' ("NOE")
+     * to lowercase.
+     *
+     * @param \DOMXPath $xpath
+     *
+     * @return void
+     */
```
oliverklee: Thanks for the cleanup! :heart:

Classes/Emogrifier.php
```diff
@@ -530,6 +530,13 @@ private function fillStyleAttributesWithMergedStyles()
     private function generateStyleStringFromDeclarationsArrays(array $oldStyles, array $newStyles)
     {
         $combinedStyles = array_merge($oldStyles, $newStyles);
+
+        foreach ($oldStyles as $attribute => $expression) {
+            if (isset($newStyles[$attribute]) && strtolower(substr($expression, -10)) === '!important') {
```
oliverklee: Do we need to use substr here, or would it suffice to use stripos to check whether !important is part of the string?

Classes/Emogrifier.php
```diff
@@ -530,6 +530,13 @@ private function fillStyleAttributesWithMergedStyles()
     private function generateStyleStringFromDeclarationsArrays(array $oldStyles, array $newStyles)
     {
         $combinedStyles = array_merge($oldStyles, $newStyles);
+
+        foreach ($oldStyles as $attribute => $expression) {
```
oliverklee: These should be named $attributeName and $attributeValue, if I understand those two variables correctly.

oliverklee: For better readability, please let's move this to a well-named private method. I propose the name "applyImportantStyles".

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1519,77 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
             $this->subject->emogrify()
         );
     }
+
+    /**
+     * @test
+     */
+    public function emogrifyHandleNoImportant()
```
oliverklee: Is this test any different from the existing tests? If not, it should be removed.

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1519,77 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
...
+        self::assertContains(
+            $expected,
+            $this->subject->emogrify()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyHandleImportantStyleTag()
```
oliverklee: I propose the name "importandInExternalCssOverwritesInlineCss".

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1519,77 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
...
+    {
+        $css = 'p { margin: 1px !important; padding: 1px;}';
+        $html = $this->html5DocumentType . self::LF .
+            '<html><head</head><body><p style="margin: 2px; text-align: center;">some content</p></body></html>';
```
oliverklee: Let's move the second rule (text-align: center) to a separate test that checks that important rules leaves the other rules intact.

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1519,77 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
...
+
+    /**
+     * @test
+     */
+    public function emogrifyHandleImportantStyleTagCaseInsensitive()
+    {
+        $css = 'p { margin: 1px !ImPorTant; padding: 1px;}';
+        $html = $this->html5DocumentType . self::LF .
+            '<html><head</head><body><p style="margin: 2px; text-align: center;">some content</p></body></html>';
+        $expected = '<p style="margin: 1px !ImPorTant; text-align: center; padding: 1px;">';
```
oliverklee: For this test, we don't need the text-align: center.

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1519,77 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
...
+        self::assertContains(
+            $expected,
+            $this->subject->emogrify()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyHandleImportantStyleTagCaseInsensitive()
```
oliverklee: handles (third person), please

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1519,77 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
...
+        self::assertContains(
+            $expected,
+            $this->subject->emogrify()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyHandleImportantStyleTagAndAttribute()
```
oliverklee: handles (third person), please

oliverklee: I'd prefer the name importandStyleRuleFromInlineCssOverwritesImportantStyleRuleFromExternalCss

Classes/Emogrifier.php
vanderlee: The string "!important" could be part of a URL or `content` value. It must be at the end. I do need to check for any possible whitespaces though.

oliverklee: Then we'll probably need to either use a regular expression, or we'd need another trim (or make sure that the elements in $oldStyles are trimmed when setting).

vanderlee:
For some reason I can't fix the build issue that I didn't even cause in the first place.

oliverklee:
The build issue is due to your copy of Emogrifier using Windows line endings (0x0d 0x0a) instead of Unix line endings (0x0a).

Tools like dos2unix can fix this for you, and Eclipse/Zend Studio also has a function for this. (PhpStorm does not, AFAIK).

oliverklee:
As for the second commit (the merge commit): Please always use rebase master on your branch, not merge master, to avoid these merge commits.

vanderlee:
I'm done.

This, including all the tweaks, could have been done many hours of work ago and I no longer feel my time is best spent continuing on.

Feel free to take over this and my other changes.

oliverklee:
Hi @vanderlee , I've just created a new PR #214 for this, did the remaining changes and merged it. Thanks a lot for contributing!

PR #206, opened 2015-07-08
emogrifyBodyContent() method, returning only the HTML within <body> tag.

vanderlee:
Included a `returnBodyContentOnly ()` function which, after called to enable the feature, will return the content of the `<body>` tag only, instead of the entire HTML.

For instance, `<!DOCTYPE html><html><head>...</head><body><p>Hello</p></body><html>` will return just `<p>Hello</p>`. This allows Emogrifier to be used on fragments of HTML code instead of entire files.

Unittests are included to test both default and feature behaviour.

If you prefer different function names, please let me know.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -181,7 +181,8 @@ class Emogrifier
     }
 
     /**
-     * Applies the CSS you submit to the HTML you submit.
+     * Applies $this->css to $this->html and returns the HTML with the CSS
+     * applied.
      *
      * This method places the CSS inline.
      *
@@ -196,6 +197,49 @@ class Emogrifier
         }
 
         $xmlDocument = $this->createXmlDocument();
+        $this->process($xmlDocument);
+
+        return $xmlDocument->saveHTML();
+    }
+
+    /**
+     * Applies $this->css to $this->html and returns only the HTML content
+     * within the <body> tag.
+     *
+     * This method places the CSS inline.
+     *
+     * @return string
+     *
+     * @throws \BadMethodCallException
+     */
+    public function emogrifyBodyContent()
+    {
+        if ($this->html === '') {
+            throw new \BadMethodCallException('Please set some HTML first before calling emogrify.', 1390393096);
+        }
+
+        $xmlDocument = $this->createXmlDocument();
+        $this->process($xmlDocument);
+
+        $innerDocument = new \DOMDocument();
+        foreach ($xmlDocument->documentElement->getElementsByTagName('body')->item(0)->childNodes as $childNode) {
+            $innerDocument->appendChild($innerDocument->importNode($childNode, true));
+        }
+
+        return $innerDocument->saveHTML();
+    }
+
+    /**
+     * Applies $this->css to $xmlDocument.
+     *
+     * This method places the CSS inline.
+     *
+     * @param \DOMDocument $xmlDocument
+     *
+     * @return void
+     */
+    protected function process(\DOMDocument $xmlDocument)
+    {
         $xpath = new \DOMXPath($xmlDocument);
         $this->clearAllCaches();
 
@@ -297,8 +341,6 @@ class Emogrifier
         }
 
         $this->copyCssWithMediaToStyleNode($cssParts, $xmlDocument);
-
-        return $xmlDocument->saveHTML();
     }
 
     /**
--- a/Tests/Unit/EmogrifierTest.php
+++ b/Tests/Unit/EmogrifierTest.php
@@ -75,6 +75,29 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
 
     /**
      * @test
+     *
+     * @expectedException \BadMethodCallException
+     */
+    public function emogrifyBodyContentForNoDataSetReturnsThrowsException()
+    {
+        $this->subject->emogrifyBodyContent();
+    }
+
+    /**
+     * @test
+     *
+     * @expectedException \BadMethodCallException
+     */
+    public function emogrifyBodyContentForEmptyHtmlAndEmptyCssThrowsException()
+    {
+        $this->subject->setHtml('');
+        $this->subject->setCss('');
+
+        $this->subject->emogrifyBodyContent();
+    }
+
+    /**
+     * @test
      */
     public function emogrifyAddsHtmlTagIfNoHtmlTagAndNoHeadTagAreProvided()
     {
@@ -1517,6 +1540,47 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
         self::assertContains(
             '<body><p></p></body>',
             $this->subject->emogrify()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyReturnsCompleteHtmlDocument()
+    {
+        $this->subject->setHtml($this->html5DocumentType . '<html><body><p></p></body></html>');
+
+        self::assertSame(
+            $this->html5DocumentType . self::LF .
+            '<html>' . self::LF .
+            '<head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head>' . self::LF .
+            '<body><p></p></body>' . self::LF .
+            '</html>' . self::LF,
+            $this->subject->emogrify()
+        );
+    }
+    
+    /**
+     * @test
+     */
+    public function emogrifyBodyContentReturnsBodyContentFromHtml()
+    {
+        $this->subject->setHtml($this->html5DocumentType . '<html><body><p></p></body></html>');
+        self::assertSame(
+            '<p></p>' . self::LF,
+            $this->subject->emogrifyBodyContent()
+        );
+    }
+    
+    /**
+     * @test
+     */
+    public function emogrifyBodyContentReturnsBodyContentFromContent()
+    {
+        $this->subject->setHtml('<p></p>');
+        self::assertSame(
+            '<p></p>' . self::LF,
+            $this->subject->emogrifyBodyContent()
         );
     }
 }


oliverklee:
Hi @vanderlee, thanks for this PR! Could you please squash your commits into one, prefix the commit message subject with [FEATURE], and split the commit message into a subject line (no longer than 50 characters), a blank line and an optional description (word-wrapped at 79 characters)?

Classes/Emogrifier.php
```diff
@@ -298,7 +306,17 @@ public function emogrify()
 
         $this->copyCssWithMediaToStyleNode($cssParts, $xmlDocument);
 
-        return $xmlDocument->saveHTML();
+        if ($this->returnBodyContentOnly) {
```
oliverklee: Instead of having a behavior switch (which is a bad smell), I'd prefer a separate method named something like "emogrifyBody" or "emogrifyBodyOnly".

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1519,47 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
...
+    }
+    
+    /**
+     * @test
+     */
+    public function emogrifyReturnBodyContentOnlyFromHtml()
+    {
+        $this->subject->setHtml($this->html5DocumentType . '<html><body><p></body></html>');
+        $this->subject->returnBodyContentOnly();
+        self::assertEquals(
```
oliverklee: Could you generally please use the type-safe variant assertSame?

oliverklee:
By the way, you can also run the code sniffer and the unit tests locally before pushing. The CONTRIBUTING.md file lists the necessary commands.

Classes/Emogrifier.php
vanderlee: The only way to do that without incurring excessive performance cost would be to refactor the vast majority of the `emogrify()` method into a new (internal) method --taking and returning a DOMDocument-- and have both the "output" functions `emogrify()` and `emogrifyBody()` (or whatever it'll be called) call that new method.

This approach would allow more extendability too, as it separates the processing from the interface.

This would be fine by me, but would it suit you?

oliverklee: Yes, I love that approach.

I see two possible approaches (which are both completely fine with me):

a) create a separate PR for the refactoring and then continue with this PR here after I've merged the refactoring PR
b) do the refactoring as part of this PR

vanderlee:
Please check my proposed changes.

Classes/Emogrifier.php
```diff
@@ -181,7 +181,7 @@ public function setCss($css)
     }
 
     /**
-     * Applies the CSS you submit to the HTML you submit.
+     * Applies the CSS you submit to the HTML you submit and returns the HTML.
```
oliverklee: I think the original "the CSS you submit" was not the best wording to begin with.

What about "Applies $this->css to $this->html and returns the HTML with the CSS applied."?

Classes/Emogrifier.php
```diff
@@ -195,7 +195,50 @@ public function emogrify()
...
 
-        $xmlDocument = $this->createXmlDocument();
+        $xmlDocument = $this->process($this->createXmlDocument());
+        
+        return $xmlDocument->saveHTML();
+    }
+                
+    /**
+     * Applies the CSS you submit to the HTML you submit and returns only the
+     * HTML content within the <body> tag.
```
oliverklee: The same about the wording.

Classes/Emogrifier.php
```diff
@@ -195,7 +195,50 @@ public function emogrify()
...
+     *
+     * @return string
+     *
+     * @throws \BadMethodCallException
+     */
+    public function emogrifyBodyContent()
+    {
+        if ($this->html === '') {
+            throw new \BadMethodCallException('Please set some HTML first before calling emogrify.', 1390393096);
+        }
```
oliverklee: Could you please add a unit test for this exception?

Classes/Emogrifier.php
```diff
@@ -195,7 +195,50 @@ public function emogrify()
...
+        $innerDocument = new \DOMDocument();
+        foreach ($xmlDocument->documentElement->getElementsByTagName('body')->item(0)->childNodes as $childNode) {
+            $innerDocument->appendChild($innerDocument->importNode($childNode, true));
+        }
+        
+        return $innerDocument->saveHTML();
+    }
+    
+    /**
+     * Applies the submitted CSS to the submitted HTML.
```
oliverklee: Ditto about the wording.

Classes/Emogrifier.php
```diff
@@ -195,7 +195,50 @@ public function emogrify()
...
+     *
+     * This method places the CSS inline.
+     *
+     * @param \DOMDocument $xmlDocument
+     *
+     * @return \DOMDocument
+     *
+     * @throws \BadMethodCallException
```
oliverklee: This annotation is not true as this method does not throw the exception. Could you please remove it?

Classes/Emogrifier.php
```diff
@@ -195,7 +195,50 @@ public function emogrify()
...
+    }
+    
+    /**
+     * Applies the submitted CSS to the submitted HTML.
+     *
+     * This method places the CSS inline.
+     *
+     * @param \DOMDocument $xmlDocument
+     *
+     * @return \DOMDocument
```
oliverklee: We don't need to return $xmlDocument as it is passed by kind-of-reference and gets modified in the method. So the caller can just work on their reference after the call.

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1519,45 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
             $this->subject->emogrify()
         );
     }
+
+    /**
+     * @test
+     */
+    public function emogrifyDefaultReturn()
+    {
+        $this->subject->setHtml($this->html5DocumentType . '<html><body><p></body></html>');
```
oliverklee: Could you please add the closing p tag in the new tests (as the tests are not about invalid XML)?

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1519,45 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
             $this->subject->emogrify()
         );
     }
+
+    /**
+     * @test
+     */
+    public function emogrifyDefaultReturn()
```
oliverklee: Generally, test names should be complete sentences (though maybe a git Pidgin-English style).

Better name:

emogrifyReturnsCompleteHtmlDocument

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1519,45 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
...
+            '<body><p></p></body>' . self::LF .
+            '</html>' . self::LF,
+            $this->subject->emogrify()
+        );
+    }
+    
+    /**
+     * @test
+     */
+    public function emogrifyReturnBodyContentFromHtml()
```
oliverklee: Third person, please: Return -> Returns

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1519,45 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
...
+        self::assertSame(
+            '<p></p>' . self::LF,
+            $this->subject->emogrifyBodyContent()
+        );
+    }
+    
+    /**
+     * @test
+     */
+    public function emogrifyReturnBodyContentFromContent()
```
oliverklee: Third person, please: Return -> Returns

Tests/Unit/EmogrifierTest.php
oliverklee: This should be the new method name: emogrify -> emogrifyBodyContent

Tests/Unit/EmogrifierTest.php
oliverklee: This should be the new method name: emogrify -> emogrifyBodyContent

oliverklee:
Great PR! I just needs some polish.

vanderlee:
Polish applied!

Let me know if this is good. There's no point squashing until it's good.

oliverklee:
@vanderlee Thanks! I'll have a look at it.

oliverklee:
@vanderlee Concerning the squashing: You can avoid having more than one commit per commit ( = per branch = per PR) if you use commit --amend for any changes to the first commit:

https://nathanhoad.net/git-amend-your-last-commit

Tests/Unit/EmogrifierTest.php
```diff
@@ -75,6 +75,29 @@ public function emogrifyForEmptyHtmlAndEmptyCssThrowsException()
 
     /**
      * @test
+     *
+     * @expectedException \BadMethodCallException
+     */
+    public function emogrifyBodyContentForNoDataSetReturnsThrowsException()
```
oliverklee: :heart:

Classes/Emogrifier.php
```diff
@@ -196,6 +197,47 @@ public function emogrify()
...
+        $innerDocument = new \DOMDocument();
+        foreach ($xmlDocument->documentElement->getElementsByTagName('body')->item(0)->childNodes as $childNode) {
+            $innerDocument->appendChild($innerDocument->importNode($childNode, true));
+        }
+
+        return $innerDocument->saveHTML();
+    }
+
+    /**
+     * Applies $this->css to a \DOMDocument.
```
oliverklee: Or even better: "… to $xmlDocument."

Classes/Emogrifier.php
```diff
@@ -196,6 +197,47 @@ public function emogrify()
...
+     *
+     * This method places the CSS inline.
+     *
+     * @param \DOMDocument $xmlDocument
+     */
```
oliverklee: Could you please add the mandatory @return void?

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1542,45 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
...
+            '<body><p></p></body>' . self::LF .
+            '</html>' . self::LF,
+            $this->subject->emogrify()
+        );
+    }
+    
+    /**
+     * @test
+     */
+    public function emogrifyBodyContentFromHtml()
```
oliverklee: Could you please add a verb ("returns") to the test name so that it reads like a full English sentence?

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1542,45 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
...
+        self::assertSame(
+            '<p></p>' . self::LF,
+            $this->subject->emogrifyBodyContent()
+        );
+    }
+    
+    /**
+     * @test
+     */
+    public function emogrifyBodyContentFromContent()
```
oliverklee: Could you please add a verb ("returns") to the test name so that it reads like a full English sentence?

oliverklee:
I've reviewed your changes. Almost there!

vanderlee:
Latest changes applied and squashed.

Tests/Unit/EmogrifierTest.php
```diff
@@ -1519,4 +1542,45 @@ public function emogrifyAutomaticallyClosesUnclosedTag()
...
+            '<body><p></p></body>' . self::LF .
+            '</html>' . self::LF,
+            $this->subject->emogrify()
+        );
+    }
+    
+    /**
+     * @test
+     */
+    public function emogrifyReturnsBodyContentFromHtml()
```
oliverklee: Sorry for the to and fro.

Actually, the test names usually should start with the name of the tested method (i.e., emogrifyBodyContent) and also be a complete sentence. So the tests should be named emogrifyBodyContentReturnsBodyContentFromHtml and Content.

oliverklee:
I'd like to merge this one next. @vanderlee Could you please rebase and update the two test names? Then I'll merge it. Thanks! :heart:

vanderlee:
Function names changed

oliverklee:
Thanks! Merged.

PR #208, opened 2015-07-09
[FEATURE] Cached generateStyleStringFromDeclarationsArrays method

vanderlee:
Adding a cache to `generateStyleStringFromDeclarationsArrays` improves performance on my realworld tests by roughly 30%.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -33,6 +33,11 @@ class Emogrifier
      * @var int
      */
     const CACHE_KEY_CSS_DECLARATION_BLOCK = 3;
+    
+    /**
+     * @var int
+     */
+    const CACHE_KEY_COMBINED_STYLES = 4;
 
     /**
      * for calculating nth-of-type and nth-child selectors
@@ -91,6 +96,7 @@ class Emogrifier
         self::CACHE_KEY_SELECTOR => array(),
         self::CACHE_KEY_XPATH => array(),
         self::CACHE_KEY_CSS_DECLARATION_BLOCK => array(),
+        self::CACHE_KEY_COMBINED_STYLES => [],
     );
 
     /**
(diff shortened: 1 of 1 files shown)


vanderlee:
I don't see what the merge conflicts might be.

oliverklee:
I think the issue is that the master on your fork https://github.com/vanderlee/emogrifier is behind this repository's master: "his branch is 7 commits behind jjriv:master"

This is how you sync your for to this repository:

1. (once) add this repo as another remote:

    git add remote upstream git@github.com:jjriv/emogrifier.git

2. every now an then (i.e., when the official repo has new commits):

git checkout master
git fetch upstream
git merge upstream/master

3. after that, for every of your PRs:

git checkout your-pr-branch
git rebase master
(then resolve any rebase conflicts)
git push -f


Here's more documentation on this:
https://help.github.com/articles/configuring-a-remote-for-a-fork/
https://help.github.com/articles/syncing-a-fork/
https://help.github.com/articles/resolving-merge-conflicts-after-a-git-rebase/

Hope this helps! :smile:

Classes/Emogrifier.php
```diff
@@ -530,11 +536,21 @@ private function fillStyleAttributesWithMergedStyles()
     private function generateStyleStringFromDeclarationsArrays(array $oldStyles, array $newStyles)
     {
         $combinedStyles = array_merge($oldStyles, $newStyles);
+
+        $cacheKey = serialize($combinedStyles);
```
oliverklee: As far as I know, serialize is quite slow. Have you checked whether json_encode or var_export on $combinedStyles might be faster?

Classes/Emogrifier.php
```diff
@@ -530,11 +536,21 @@ private function fillStyleAttributesWithMergedStyles()
...
+        if (isset($this->caches[self::CACHE_KEY_COMBINED_STYLES][$cacheKey])) {
+            return $this->caches[self::CACHE_KEY_COMBINED_STYLES][$cacheKey];
+        }
+
         $style = '';
         foreach ($combinedStyles as $attributeName => $attributeValue) {
             $style .= (strtolower(trim($attributeName)) . ': ' . trim($attributeValue) . '; ');
         }
-        return trim($style);
+        $trimmedStyle = substr($style, 0, -1);
```
oliverklee: You could just do a rtrim (as the rightmost character is a space). This probably is faster (and more straightforward to read).

oliverklee:
The PR looks very good! :heart: Just on thing to check performance-wise and one idea to simplify the code.

Classes/Emogrifier.php
vanderlee: I have tested `json_encode` in the past for similar purposes; `serialize` is consistantly faster with less memory overhead. Don't know about `var_export`, though I doubt all the extra formatting would help.

Classes/Emogrifier.php
vanderlee: The `substr` seems to be marginally faster than the `trim` in my tests. Using `rtrim` isn't any better or worse than `trim`, even when specifying only the space character needs to be trimmed.

I think rewriting the whole thing to use an array and `join` may be faster; typically letting the low-level code deal with things like lots of concatenation is faster in PHP than doing it manually. This also goes for some of the other code.

We're talking about marginal differences in this case, though.

vanderlee:
I'll dive into performance some more later. My gut feeling tells me there is a lot of waste going on with excessive `strtolower` calls and manual string handling in general. There are also some `in_array` and `strlen` calls, which are typically easy to replace with noticably faster (though not necessarily pretty) alternatives.

Classes/Emogrifier.php
oliverklee: Ah, okay. Then please let's use rtrim (for better clarity).

vanderlee:
Replaced with #211

PR #211, opened 2015-07-10
[FEATURE] Cache combinedStyles

vanderlee:
Same as #208, but on current master.

The current master with the short array fix caused so much changes that it was easier to just start anew with this feature than try to resolve all the conflicts.

I also added the clearing of the cache, which decreases performance gains somewhat: back to ~20% from ~30%.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -35,6 +35,11 @@ class Emogrifier
     const CACHE_KEY_CSS_DECLARATION_BLOCK = 3;
 
     /**
+     * @var int
+     */
+    const CACHE_KEY_COMBINED_STYLES = 4;
+
+    /**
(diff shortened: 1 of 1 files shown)


oliverklee:
Looks good. If you remove the trailing space (thus making the build pass), I'll directly merge it. :heart: 

(I would need to create an new PR if I wanted to edit a PR from a fork. So it's less work for me if you do the change.)

oliverklee:
By the way, you can configure most IDEs (including PhpStorm) to automatically drop any trailing whitespace on saving a file.

vanderlee:
Squashed and somehow had to fix a totally new /r/n error on a line I never touched (the very first).

oliverklee:
Merged. @vanderlee Thanks a lot for this PR!

PR #232, opened 2015-09-24
[FEATURE] Add CSS to HTML attribute mapping

sanderkruger:
This addition enables you to put more styles in the CSS and dismisses the need to put for example width and height attributes inline. Outlook 2007, for example will ignore width and height CSS properties on images so you really need the width and height as HTML attributes.

mapStyles2HTMLAttrs takes CSS properties that have HTML equivalents and adds these HTML attributes to the DOM nodes.

The biggest advantage I see is that you can have cleaner HTML markup and more easy building a system with email messages that can be branded dynamically with different stylesheets.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -148,6 +148,14 @@ class Emogrifier
     private $shouldKeepInvisibleNodes = true;
 
     /**
+     * Determines whether CSS styles that have an equivalent HTML attribute
+     * should be mapped and attached to those elements.
+     *
+     * @var bool
+     */
+    private $shouldMapCss2Html = false;
+
+    /**
      * The constructor.
      *
      * @param string $html the HTML to emogrify, must be UTF-8-encoded
@@ -304,6 +312,9 @@ class Emogrifier
                     $oldStyleDeclarations = [];
                 }
                 $newStyleDeclarations = $this->parseCssDeclarationBlock($selector['attributes']);
+                if ($this->shouldMapCss2Html) {
+                    $this->mapStyles2HtmlAttrs($newStyleDeclarations, $node);
+                }
                 $node->setAttribute(
                     'style',
                     $this->generateStyleStringFromDeclarationsArrays($oldStyleDeclarations, $newStyleDeclarations)
@@ -320,6 +331,77 @@ class Emogrifier
         }
 
         $this->copyCssWithMediaToStyleNode($xmlDocument, $xpath, $cssParts['media']);
+    }
+
+    /**
+     * Applies $newStyles to $node.
+     *
+     * This method maps CSS styles to HTML attributes and adds those to the
+     * node.
+     *
+     * @param string[] $newStyles the new CSS styles taken from the global styles
+     *                            to be applied to this node.
+     * @param \DOMNode $node      node to apply styles to.
+     *
+     * @return void
+     */
+    private function mapStyles2HtmlAttrs(array $newStyles, \DOMNode $node)
+    {
+        foreach ($newStyles as $styleName => $styleValue) {
+            // Strip !important indicator
+            $styleValue = trim(str_replace('!important', '', $styleValue));
+            if ($styleName === 'background-color') {
+                // Always override HTML attributes. CSS takes precedence.
+                $node->setAttribute('bgcolor', $styleValue);
+            } elseif ($styleName === 'background') {
+                // Parse out the color, if any
+                $styles = explode(' ', $styleValue);
+                $first = $styles[0];
+                if (!is_numeric(substr($first, 0, 1)) && substr($first, 0, 3) !== 'url') {
+                    // This is not a position or image, assume it's a color
+                    $node->setAttribute('bgcolor', $first);
+                }
+            } elseif ($styleName === 'width') {
+                $width = preg_replace('/[^0-9.%]/', '', $styleValue);
+                $node->setAttribute('width', $width);
+            } elseif ($styleName === 'height') {
+                $height = preg_replace('/[^0-9.%]/', '', $styleValue);
+                $node->setAttribute('height', $height);
+            } elseif ($styleName === 'margin' && ($node->nodeName === 'table' || $node->nodeName === 'img')) {
+                $margins = preg_split('/\s+/', $styleValue);
+                if (count($margins) === 1) {
+                    if ($margins[0] == 'auto') {
+                        $node->setAttribute('align', 'center');
+                    }
+                } elseif (count($margins) === 2 || count($margins) === 3) {
+                    if ($margins[1] == 'auto') {
+                        $node->setAttribute('align', 'center');
+                    }
+                } else {
+                    if ($margins[1] === 'auto' && $margins[3] === 'auto') {
+                        $node->setAttribute('align', 'center');
+                    }
+                }
+            } elseif ($styleName === 'text-align' &&
+                        ($node->nodeName === 'p' || $node->nodeName === 'td' || $node->nodeName === 'div')) {
+                if ($styleValue === 'left' || $styleValue === 'right' ||
+                        $styleValue === 'center' || $styleValue === 'justify') {
+                    // Only set values that are valid for the align attribute
+                    $node->setAttribute('align', $styleValue);
+                }
+            } elseif ($styleName === 'float' && ($node->nodeName === 'table' || $node->nodeName === 'img')) {
+                if ($styleValue === 'left' || $styleValue === 'right') {
+                    // Only set values that are valid for the align attribute
+                    $node->setAttribute('align', $styleValue);
+                }
+            } elseif ($styleName === 'border') {
+                if ($styleValue === 'none' || $styleValue === '0') {
+                    $node->setAttribute('border', '0');
+                }
+            } elseif ($styleName === 'border-spacing' && $node->nodeName === 'table') {
+                $node->setAttribute('cellspacing', $styleValue);
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


oliverklee:
Hi @sanderkruger, thanks for the PR!

Could you please squash the commits and also add unit tests that test the new functionality (including the default behavior and the switch)? Thanks!

Classes/Emogrifier.php
```diff
@@ -323,6 +334,81 @@ protected function process(\DOMDocument $xmlDocument)
...
+     * This method maps CSS styles to HTML attributes and adds those to the
+     * node.
+     *
+     * @param string[] $newStyles the new CSS styles taken from the global styles
+     *                            to be applied to this node.
+     * @param \DOMNode $node      node to apply styles to.
+     *
+     * @return void
+     */
+    private function mapStyles2HtmlAttrs($newStyles, \DOMNode $node)
```
oliverklee: Please add the "array" type hinting for the first parameter (as part of the declaration).

Classes/Emogrifier.php
```diff
@@ -323,6 +334,81 @@ protected function process(\DOMDocument $xmlDocument)
...
+    {
+        foreach ($newStyles as $styleName => $styleValue) {
+            // Strip !important indicator
+            $styleValue = str_replace('!important', '', $styleValue);
+            if ($styleName == 'background-color') {
```
oliverklee: Generally, please use the strict comparisons (=== and !==).

Classes/Emogrifier.php
```diff
@@ -323,6 +334,81 @@ protected function process(\DOMDocument $xmlDocument)
...
+                // Always override HTML attributes. CSS takes precedence.
+                $node->setAttribute('bgcolor', $styleValue);
+            } elseif ($styleName == 'background') {
+                // Parse out the color, if any
+                $styles = explode(' ', $styleValue);
```
oliverklee: What about cases with multiple spaces inbetween?

Classes/Emogrifier.php
```diff
@@ -323,6 +334,81 @@ protected function process(\DOMDocument $xmlDocument)
...
+                $first = $styles[0];
+                if (!is_numeric(substr($first, 0, 1)) && substr($first, 0, 3) != 'url') {
```
oliverklee: is_numeric produces false positives: "3e14" or "3.14159" or "-1" would be numeric, too. I propose a regular expression instead.

Classes/Emogrifier.php
```diff
@@ -323,6 +334,81 @@ protected function process(\DOMDocument $xmlDocument)
...
+                    // Only set values that are valid for the align attribute
+                    $node->setAttribute('align', $styleValue);
+                }
+            } elseif ($styleName == 'border') {
+                if ($styleValue == 'none' || $styleValue == '0') {
+                    $node->setAttribute('border', '0');
+                }
+            } elseif ($styleName == 'border-spacing' && $node->nodeName == 'table') {
+                $node->setAttribute('cellspacing', $styleValue);
+            }
```
oliverklee: Generally, I find this method very long. I propose using either a switch or a look-up table instead (or a combination).

Classes/Emogrifier.php
sanderkruger: is_numeric tests a single digit, so I'm not worried about other formats. Basically, it needs to test whether the first background style is a color. If the property starts with a digit, the color is omitted and it is a position. If the property starts with url, the color is omitted and it's an image.

I should add checks for repeat and named positions.

Classes/Emogrifier.php
sanderkruger: I agree in principle, but I found that conditions are rather diverse to squeeze them in a lookup table. I'm not sure if you'd like a switch with lots of little functions to replace the case bodies.

sanderkruger:
I added a set of tests and improved the code some more. I'm afraid I don't have the complete travis toolchain locally, so that results in a few extra commits to fix coding rules.

sanderkruger:
Hi I managed to find my way to squash my commits and decided to put the bugfix on multi-line media queries in this branch too (because of a few merges on the other branch, I can't get that one straightened out).

I've processed all the comments except for the length of the method. Not sure what to do about it. A switch will actually make things longer unless offloading functionality in other methods. A table would probably become rather complicated because the conditions include test on the style, tagname, value and there's also replacements and parsing of styles.

Please let me know your thoughts on this.

Classes/Emogrifier.php
sanderkruger: Hello Oliver,

Do you plan to pull this feature on the main branch soon? If not, I'd 
like to reorganize my branches a little bit.

Unfortunately, after forking I just started developing on my master 
branch and everything I do on master will go into this pull request. So 
I've been holding off on doing more changes until it was merged. If you 
need more time, I'll close the current pull request and create it again 
on a feature branch of my fork so I can continue on my master branch.

Regards,
Sander Kruger.

------------------------------------------------------------------------

*From:* Oliver Klee
*Sent:* Monday, September 28, 2015 9:45AM
*To:* Jjriv/emogrifier
*Cc:* Sander Kruger
*Subject:* Re: [emogrifier] [FEATURE] Add CSS to HTML attribute mapping 
(#232)

In Classes/Emogrifier.php 
<https://github.com/jjriv/emogrifier/pull/232#discussion_r40524553>:

> +                        $styleValue == 'center' || $styleValue == 'justify') {
> +                    // Only set values that are valid for the align attribute
> +                    $node->setAttribute('align', $styleValue);
> +                }
> +            } elseif ($styleName == 'float' && ($node->nodeName == 'table' || $node->nodeName == 'img')) {
> +                if ($styleValue == 'left' || $styleValue == 'right') {
> +                    // Only set values that are valid for the align attribute
> +                    $node->setAttribute('align', $styleValue);
> +                }
> +            } elseif ($styleName == 'border') {
> +                if ($styleValue == 'none' || $styleValue == '0') {
> +                    $node->setAttribute('border', '0');
> +                }
> +            } elseif ($styleName == 'border-spacing' && $node->nodeName == 'table') {
> +                $node->setAttribute('cellspacing', $styleValue);
> +            }

Generally, I find this method very long. I propose using either a switch 
or a look-up table instead (or a combination).

—
Reply to this email directly or view it on GitHub 
<https://github.com/jjriv/emogrifier/pull/232/files#r40524553>.

oliverklee:
Hi @sanderkruger, I want this PR to be perfect before I merge it. So I think it still needs some work (I'll add my review comments in a minute).

Generally, could you please update your master to the upstream and rebase this PR? And could you please update the commit message to have "[FEATURE]" at the beginning of the subject line? This would save me the hassle of having to check out your branch and create a separate PR when I merge this PR. Thanks!

As for having accidentally done some work on the master: You can still move the new commits to a separate branch. This Stackoverflow post describes how to do this: http://stackoverflow.com/questions/1628563/move-the-most-recent-commits-to-a-new-branch-with-git

Keep up the good work!

Classes/Emogrifier.php
```diff
@@ -148,6 +148,14 @@ class Emogrifier
     private $shouldKeepInvisibleNodes = true;
 
     /**
+     * Determines whether CSS styles that have an equivalent HTML attribute
+     * should be mapped and attached to those elements.
+     *
+     * @var bool
+     */
+    private $shouldMapCss2Html = false;
```
oliverklee: Please don't use abbreviations: shouldMapCss2Html -> shouldMapCssToHtml

In addition, the mapping is to HTML _attributes_. So the naming should be mapCssToHtmlAttributes.

Classes/Emogrifier.php
```diff
@@ -304,6 +312,9 @@ protected function process(\DOMDocument $xmlDocument)
                     $oldStyleDeclarations = [];
                 }
                 $newStyleDeclarations = $this->parseCssDeclarationBlock($selector['attributes']);
+                if ($this->shouldMapCss2Html) {
+                    $this->mapStyles2HtmlAttrs($newStyleDeclarations, $node);
```
oliverklee: Please let's keep the naming of the property and the method consistent: Either both should use the term "Styles" or "Css".

Classes/Emogrifier.php
```diff
@@ -323,6 +334,77 @@ protected function process(\DOMDocument $xmlDocument)
...
+     * This method maps CSS styles to HTML attributes and adds those to the
+     * node.
+     *
+     * @param string[] $newStyles the new CSS styles taken from the global styles
+     *                            to be applied to this node.
+     * @param \DOMNode $node      node to apply styles to.
+     *
+     * @return void
+     */
+    private function mapStyles2HtmlAttrs(array $newStyles, \DOMNode $node)
```
oliverklee: No abbreviations, please: 2 -> To

Tests/Unit/EmogrifierTest.php
```diff
@@ -1763,4 +1763,245 @@ public function emptyMediaQueriesAreRemoved()
...
+        self::assertContains(
+            '<body style="background: url(bg.png) top;">',
+            $html
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function mapBackground2BgColor()
```
oliverklee: For all these tests, I propose you use a data provider that provides the CSS style rule and the expected HTML attribute(s): https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers

The return annotation for the data provider then should probably be string[][] (an array of arrays of strings), and the test method then will need to respective param annotations.

Classes/Emogrifier.php
```diff
@@ -323,6 +334,77 @@ protected function process(\DOMDocument $xmlDocument)
...
+    {
+        foreach ($newStyles as $styleName => $styleValue) {
```
oliverklee: As a first step to simplify this method, I propose moving the coding within the loop to a separate, well-name method.

Classes/Emogrifier.php
oliverklee: I think that for this method, we pass in only one set of styles anyway. So there is no need to name them $newStyles. $style should be sufficient.

Classes/Emogrifier.php
oliverklee: Yes, I'm favor of having separate methods that are called from the switch cases.

oliverklee: Maybe we can also have a lookup table for the easy cases (one style attribute directly matches to an attribute).

sanderkruger:
I'm closing this pull request in favour of #250.

PR #233, opened 2015-09-27
Fix for multi-line media queries

sanderkruger:
Add s modifier to regex so . matches newline. And break so a query is only added once.

I found that multi-line media queries are removed by emogrifier. This is because the regex to find them doesn't handle multi-line declarations.

This fix adds the s modifier to the regex so that the newline character can be part of the CSS.

And to prevent emogrifier to add the same media query for every selector matching an element, a break is added to the loop finding matching selectors.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -671,6 +671,7 @@ class Emogrifier
             foreach ($this->parseSelectors($mediaQuery['css']) as $selector) {
                 if ($this->existsMatchForCssSelector($xpath, $selector['selector'])) {
                     $mediaQueriesRelevantForDocument[] = $mediaQuery['query'];
+                    break;
                 }
             }
         }
@@ -687,8 +688,7 @@ class Emogrifier
      */
     private function extractMediaQueriesFromCss($css)
     {
-        preg_match_all('#(?<query>@media[^{]*\\{(?<css>(.*?)\\})(\\s*)\\})#', $css, $mediaQueries);
-
+        preg_match_all('#(?<query>@media[^{]*\\{(?<css>(.*?)\\})(\\s*)\\})#s', $css, $mediaQueries);
         $result = [];
         foreach (array_keys($mediaQueries['css']) as $key) {
             $result[] = [
--- a/Tests/Unit/EmogrifierTest.php
+++ b/Tests/Unit/EmogrifierTest.php
@@ -1763,4 +1763,27 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
             $result
         );
     }
+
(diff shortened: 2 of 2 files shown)


oliverklee:
Hi @sanderkruger, thanks for the PR!

Could you please add a unit test that fails without the fix and that is green with the fix? This way, we can be sure that the fix actually solves the problem - and we can also be sure that we won't break this functionality during later changes or refactorings.

Optimally, you can do a commit --amend with the new changes and then push -f so that there will still be only one commit in this PR.

Thanks! :heart:

sanderkruger:
I'm sorry about the double commit. I use SourceTree with emogrifier in a submodule and I'm afraid that SourceTree doesn't let you force push and I couldn't get it to work from the command line.

I could try to open a new branch and merge this one into one commit. Would you prefer that? I suppose that if you accept this pull request it would also only make a single commit on master.

oliverklee:
@sanderkruger A few words about pull requests and our workflow which will hopefully be helpful to you. :-)

First of all, there is no need to apologize for not knowing something. One of the great things about contributing to open source projects (in addition to the warm, fuzzy feeling of contributing to something greater and of leaving a footprint) is the opportunity to learn something new. So not knowing something which one might learn is perfectly fine. :-)

Then about commits and pull requests: We (in the Emogrifier project) try to keep the commit history as clean as possible. This means that there should only be exactly one commit per PR, resulting in two commits per merged PR (the merge commit and the original commit) that end up in the master history. If there are multiple commits in a PR, we (i.e., usually I) need to check out that branch, squash the commits and merge manually on the command line instead of being able to do a (far quicker) merge using the GitHub UI. So having clean PRs with only one commit saves us work and gets your PRs merged more quickly.

If you create a pull request and push to that branch again, the PR will be automatically updated (as you probably already have noticed). And if you changed the original commit, you'll need to do a force-push (push -f).

There are two ways to end up with exactly one (new) commit per branch (or per PR):

1. After the first commit, you add to the commit by doing a commit --amend. (This is the method which I usually prefer.)

2. If you already have more than one commit, you squash them into one. This page explains how to this: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html


About Travis and continuous integration: You can actually run the style checks and the unit tests locally on your machine - even before you commit your changes. Our contributing guide explains how to run the corresponding commands: https://github.com/jjriv/emogrifier/blob/master/CONTRIBUTING.md


So far about the explanations. Maybe I should add them to the CONTRIBUTING file as well ...

Thanks for your contributions! I will review your two PRs during the next few days. (And if you could squash your changes in the meantime, more power to you!)

PR #234, opened 2015-09-28
Bugfix multiline media queries2

sanderkruger:
Add s modifier to regex so . matches newline. And break so a query is only added once.

I found that multi-line media queries are removed by emogrifier. This is because the regex to find them doesn't handle multi-line declarations.

This fix adds the s modifier to the regex so that the newline character can be part of the CSS.

And to prevent emogrifier to add the same media query for every selector matching an element, a break is added to the loop finding matching selectors.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -671,6 +671,7 @@ class Emogrifier
             foreach ($this->parseSelectors($mediaQuery['css']) as $selector) {
                 if ($this->existsMatchForCssSelector($xpath, $selector['selector'])) {
                     $mediaQueriesRelevantForDocument[] = $mediaQuery['query'];
+                    break;
                 }
             }
         }
@@ -687,8 +688,7 @@ class Emogrifier
      */
     private function extractMediaQueriesFromCss($css)
     {
-        preg_match_all('#(?<query>@media[^{]*\\{(?<css>(.*?)\\})(\\s*)\\})#', $css, $mediaQueries);
-
+        preg_match_all('#(?<query>@media[^{]*\\{(?<css>(.*?)\\})(\\s*)\\})#s', $css, $mediaQueries);
         $result = [];
         foreach (array_keys($mediaQueries['css']) as $key) {
             $result[] = [
--- a/Tests/Unit/EmogrifierTest.php
+++ b/Tests/Unit/EmogrifierTest.php
@@ -1763,4 +1763,77 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
             $result
         );
     }
+
+    /**
+     * @test
+     */
+    public function multiLineMediaQueryIsAppliedOnce()
+    {
+        $css = "@media all {\r\n" .
+                ".medium {font-size:18px;}\r\n" .
+                ".small {font-size:14px;}\r\n" .
+                '}';
+        $this->subject->setCss($css);
+
+        $this->subject->setHtml($this->html5DocumentType . '<html><body>' .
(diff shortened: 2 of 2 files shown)


Classes/Emogrifier.php
```diff
@@ -671,6 +671,7 @@ private function copyCssWithMediaToStyleNode(\DOMDocument $xmlDocument, \DOMXPat
             foreach ($this->parseSelectors($mediaQuery['css']) as $selector) {
                 if ($this->existsMatchForCssSelector($xpath, $selector['selector'])) {
                     $mediaQueriesRelevantForDocument[] = $mediaQuery['query'];
+                    break;
```
oliverklee: Why this change? Is it necessary for fixing the bug?

(As far as I can see, it only limits the media queries to exactly one query.)

Tests/Unit/EmogrifierTest.php
```diff
@@ -1763,4 +1763,27 @@ public function emptyMediaQueriesAreRemoved()
             $result
         );
     }
+
+    /**
+     * @test
+     */
+    public function multiLineMediaQuery()
```
oliverklee: Please let's make all test names full sentences that describe some behavior. I propose the name "multiLineMediaQueryIsApplied".

Tests/Unit/EmogrifierTest.php
```diff
@@ -1763,4 +1763,27 @@ public function emptyMediaQueriesAreRemoved()
...
+    {
+        $mediaQuery = "@media all and (max-width: 500px) {\r\n" .
+                ".medium {font-size:18px;}\r\n" .
+                ".small {font-size:14px;}\r\n" .
```
oliverklee: Please let's keep this as simple as possible (no media query, just two rules).

Tests/Unit/EmogrifierTest.php
```diff
@@ -1763,4 +1763,27 @@ public function emptyMediaQueriesAreRemoved()
             $result
         );
     }
+
+    /**
+     * @test
+     */
+    public function multiLineMediaQuery()
+    {
+        $mediaQuery = "@media all and (max-width: 500px) {\r\n" .
```
oliverklee: Please let's have two separate test cases: one with Windows line endings (\r\n) and one with Unix line endings (\n) so we know both cases work properly.

oliverklee: The variable then also should be names just $css.

Classes/Emogrifier.php
sanderkruger: Yes this is necessary to prevent adding the same mediaquery multiple times. If you look closely, you see that the loop walks through the selectors in the mediaquery. Any match in the document will add the whole mediaquery. If multiple selectors match, the old code (without the break) would add the complete mediaquery multiple times.

The break prevents this. At the first matching selector, the whole media query is added and the loop is aborted.

oliverklee: Ah, okay. Could you please move this to a separate PR (including two unit tests: one that tests that the bug is fixed, and one that tests that multiple queries still work)?

sanderkruger: The 2 issues are related and I think it's not a good idea to complicate things with separate PRs. I have modified the unit tests so that they fail if a selector is added more than once.

So I strongly suggest to keep them in a single PR, to keep things simple, but you're the boss.

oliverklee:
Hi @sanderkruger, I've polished your PR a bit and merged it as #241. Thanks a bunch for contributing!

Issue #235, opened 2015-09-29
data uris not supported


sanderkruger:
Style declarations with images referenced as a [Data URI](https://en.wikipedia.org/wiki/Data_URI_scheme) will break.

The regex in the method parseCssDeclarationBlock probably trips over the semicolon in the data URI syntax and mistakes it for a style declaration delimiter.

I'm not enough of a crack in regular expressions to correct this.

sanderkruger:
I found that the problem is not so much with the regular expression as with the code that splits a css block into declarations. It uses explode(';' for this, chopping up the data uri.

I tried out a workaround that works (but isn't bullet-proof) for the most common data uris:
first replace ';base64' with '@base64', then do all the parsing then reverse the replacement.

If this is good enough for all the users of this project, I can make a PR for it.

oliverklee:
Sounds good!

oliverklee:
Fixed in #238 / #243.

PR #238, opened 2015-09-30
Fix parsing of CSS for data-uri's

sanderkruger:
CSS declarations with data uris contain semicolons. Semicolon is also the separator between CSS declarations. Unfortunately, this makes parsing CSS blocks a bit more complicated than just splitting the string around semicolons.

This fix covers the most widely used data uris for embedding images (base64 encoded, no charset).

By replacing the string ';base64' with '@base64' before parsing and doing the reverse after parsing, the blocks can be conveniently split around semicolons, provided there are no other semicolons lurking in the CSS.

Fixes bug #235

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -1213,14 +1213,20 @@ class Emogrifier
         }
 
         $properties = [];
-        $declarations = explode(';', $cssDeclarationBlock);
+        /* Replace the semicolon in base64-encoded data uris with a @ to prevent breaking
+         * up these declarations
+         */
+        $mangledDeclarations = str_replace(';base64', '@base64', $cssDeclarationBlock);
+        $declarations = explode(';', $mangledDeclarations);
         foreach ($declarations as $declaration) {
             $matches = [];
             if (!preg_match('/ *([A-Za-z\\-]+) *: *([^;]+) */', $declaration, $matches)) {
                 continue;
             }
             $propertyName = strtolower($matches[1]);
-            $propertyValue = $matches[2];
+            /* Put the ; back in front of the base64 declaration
+             */
+            $propertyValue = str_replace('@base64', ';base64', $matches[2]);
             $properties[$propertyName] = $propertyValue;
         }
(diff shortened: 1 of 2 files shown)


bramley:
To split on ';' except when it is followed by 'base64' you can use this regex

    $declarations = preg_split('/;(?!base64)/', $cssDeclarationBlock);

I think this will behave similarly to explode when ';' is at the beginning or end of the string by creating an empty declarations entry, but that can be avoided by adding PREG_SPLIT_NO_EMPTY as a third parameter.

oliverklee:
Hi @bramley, I tried your approach, but this caused the test to fail i.e., it didn't work properly.

oliverklee:
Hi @sanderkruger, I've polished and merged your PR as #243. Thanks for contributing to Emogrifier! :heart:

bramley:
The preg_match within the loop would also need to change as that is testing for ';'.
But looking at it a bit more closely, the test for is unnecessary because after the explode,

    explode(';', $cssDeclarationBlock);
 $declaration won't ever contain ';'. This part of the regex `([^;]+)` will consume the rest of the string including any trailing spaces.
So that regex could be simplified. Changing it to 

    if (!preg_match('/ *([A-Za-z\\-]+) *: *(.+)/', $declaration, $matches)) {
then the test passes for both the new code and also using my original suggestion.
Also, the tests for leading and surrounding spaces ` *` should probably be for any whitespace to allow tabs and line breaks (unless the text has already been normalised).

sanderkruger:
I agree that the [^;]+ pattern doesn't make sense. There can't be any semicolon.

Replacing ``` *``` with ```\s*``` is a good idea too.

And the preg_split solution is more elegant than my initial solution and easier to extend. data uris can also contain ';charset' so the explode can be replaced with

```
preg_split('/;(?!base64|charset)/', $cssDeclarationBlock);
```

I'd propose that you put it on another PR since this one is closed.

oliverklee:
@bramley @sanderkruger Would you be willing to create a PR for this (including more unit tests to cover so-far uncovered edge cases if necessary)?

bramley:
I can look at creating a new PR but won't be until Monday.

PR #250, opened 2015-10-11
[FEATURE] Add CSS to HTML attribute mapper

sanderkruger:
I moved pull-request #232 to a new branch so I can do other stuff on my master branch.

This addition enables you to put more styles in the CSS and dismisses the need to put for example width and height attributes inline. Outlook 2007, for example will ignore width and height CSS properties on images so you really need the width and height as HTML attributes.

mapStyles2HTMLAttrs takes CSS properties that have HTML equivalents and adds these HTML attributes to the DOM nodes.

The biggest advantage I see is that you can have cleaner HTML markup and more easy building a system with email messages that can be branded dynamically with different stylesheets.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -154,6 +154,41 @@ class Emogrifier
     private $shouldKeepInvisibleNodes = true;
 
     /**
+     * Determines whether CSS styles that have an equivalent HTML attribute
+     * should be mapped and attached to those elements.
+     *
+     * @var bool
+     */
+    private $shouldMapCssToHtml = false;
+
+    /**
+     * This multi-level array contains simple mappings of CSS properties to
+     * HTML attributes. If a property maps simply to an attribute, the mapping
+     * can be a string, the name of the attribute. If a mapping only applies to
+     * certain HTML nodes or only for certain values, the mapping is an object
+     * with a whitelist of nodes and values.
+     *
+     * @var mixed[]
+     */
+    private $cssToHtmlMap = [
+        'background-color' => 'bgcolor',
+        'text-align' => [
+            'attribute' => 'align',
+            'nodes' => ['p', 'td', 'div'],
+            'values' => ['left', 'right', 'center', 'justify']
+        ],
+        'float' => [
+            'attribute' => 'align',
+            'nodes' => ['table', 'img'],
+            'values' => ['left', 'right']
+        ],
+        'border-spacing' => [
+            'attribute' => 'cellspacing',
+            'nodes' => ['table']
+        ]
+    ];
+
+    /**
      * The constructor.
      *
      * @param string $html the HTML to emogrify, must be UTF-8-encoded
@@ -311,6 +346,9 @@ class Emogrifier
                     $oldStyleDeclarations = [];
                 }
                 $newStyleDeclarations = $this->parseCssDeclarationsBlock($cssRule['declarationsBlock']);
+                if ($this->shouldMapCssToHtml) {
+                    $this->mapCssToHtmlAttributes($newStyleDeclarations, $node);
+                }
                 $node->setAttribute(
                     'style',
                     $this->generateStyleStringFromDeclarationsArrays($oldStyleDeclarations, $newStyleDeclarations)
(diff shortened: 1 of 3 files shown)


sanderkruger:
@oliverklee this version includes changes in relation to your code review.

Tests/Unit/EmogrifierTest.php
```diff
@@ -1861,4 +1861,57 @@ public function dataUrisAreConserved()
...
+
+    /**
+     * Data provider for css to html mapping.
+     *
+     * @return array[]
+     */
+    public function cssToHtmlMappingDataProvider()
+    {
+        return [
+            'Map bakground-color to bgcolor'
```
oliverklee: baCkground

Tests/Unit/EmogrifierTest.php
```diff
@@ -1861,4 +1861,57 @@ public function dataUrisAreConserved()
...
+                => ['', 'body {background-color: red;}', 'body', 'bgcolor="red"'],
+            'Do not map background url'
```
oliverklee: url -> URL

Tests/Unit/EmogrifierTest.php
```diff
@@ -1861,4 +1861,57 @@ public function dataUrisAreConserved()
...
+                => ['', 'body {background: red top;}', 'body', 'bgcolor="red"'],
+            'Map width and height'
+                => ['', 'body {width: 100%; height: 200px;}', 'body', 'width="100%" height="200"'],
+            'Map text-align:center to align=center'
+                => ['<p>mid</p>', 'p {text-align: center;}', 'p', 'align="center"'],
+            'Do not map text-align: inherit'
+                => ['<p>plain</p>', 'p {text-align: interit;}', 'p', 'style'],
+            'Map margin: 0 auto to align=center'
+                => ['<img>', 'img {margin: 0 auto;}', 'img', 'align="center"'],
+            'Map float: right to aling=right'
```
oliverklee: aling -> align

sanderkruger:
Updated to review

Classes/Emogrifier.php
```diff
@@ -323,6 +334,94 @@ protected function process(\DOMDocument $xmlDocument)
...
+            }
+        } elseif ($styleName === 'text-align' &&
+                    ($node->nodeName === 'p' || $node->nodeName === 'td' || $node->nodeName === 'div')) {
+            if ($styleValue === 'left' || $styleValue === 'right' ||
+                    $styleValue === 'center' || $styleValue === 'justify') {
+                // Only set values that are valid for the align attribute
+                $node->setAttribute('align', $styleValue);
+            }
+        } elseif ($styleName === 'float' && ($node->nodeName === 'table' || $node->nodeName === 'img')) {
+            if ($styleValue === 'left' || $styleValue === 'right') {
```
oliverklee: I see a pattern emerging here. I propose we use it the following way to further simplify this method:

1. Some properties map only to certain HTML elements. I propose adding a string[][] array (as a class property to the Emogrifier class) that uses the property names as keys and allowed HTML element names (= nodeName) as values in the inner arrays. If the property is a key (isset) in this array and the HTML element name is not in_array, bail out (in a guard clause). Only properties that have limits on the HTML elements should be in this array.

2. Some properties only are usable with certain values. Create a similar array with a value whitelist.

3. We then can change this if/elseif/if construct to a switch construct.

What do you think?

Classes/Emogrifier.php
```diff
@@ -323,6 +334,94 @@ protected function process(\DOMDocument $xmlDocument)
...
+    }
+
+    /**
+     * Applies $styles to $node.
+     *
+     * This method maps a CSS rule to HTML attributes and adds those to the
+     * node.
+     *
+     * @param string $styleName   the name of the style rule to map.
+     * @param string $styleValue  the value of the style rule to map.
```
oliverklee: According to the CSS nomenclature, those are called "property" and "value".

https://developer.mozilla.org/en-US/docs/Web/CSS/Syntax

Classes/Emogrifier.php
```diff
@@ -323,6 +334,94 @@ protected function process(\DOMDocument $xmlDocument)
...
+    {
+        foreach ($styles as $styleName => $styleValue) {
+            // Strip !important indicator
+            $styleValue = trim(str_replace('!important', '', $styleValue));
+            $this->mapCssToHtmlAttribute($styleName, $styleValue, $node);
+        }
+    }
+
+    /**
+     * Applies $styles to $node.
```
oliverklee: There is no parameter $styles anymore. I think this complete line is a copy'n'paste error.

Tests/Unit/EmogrifierTest.php
```diff
@@ -1861,4 +1861,57 @@ public function dataUrisAreConserved()
...
+            'Do not map text-align: inherit'
+                => ['<p>plain</p>', 'p {text-align: interit;}', 'p', 'style'],
+            'Map margin: 0 auto to align=center'
+                => ['<img>', 'img {margin: 0 auto;}', 'img', 'align="center"'],
+            'Map float: right to align=right'
+                => ['<img>', 'img {float: right;}', 'img', 'align="right"'],
+            'Map border: none to border=0'
+                => ['<img>', 'img {border: none;}', 'img', 'border="0"'],
+            'Map border-spacing: 0 to cellspacing=0'
+                => ['<table><tr><td></td></tr></table>', 'table {border-spacing: 0;}', 'table', 'cellspacing="0"']
```
oliverklee: For CSS properties that are limited to certain HTML elements or certain values, we also need one test (or data provider line) for each allowed element and value, and also one each for a case where neither applies, i.e., where the style is not applied.

Tests/Unit/EmogrifierTest.php
```diff
@@ -1861,4 +1861,57 @@ public function dataUrisAreConserved()
...
+                => ['', 'body {width: 100%; height: 200px;}', 'body', 'width="100%" height="200"'],
+            'Map text-align:center to align=center'
+                => ['<p>mid</p>', 'p {text-align: center;}', 'p', 'align="center"'],
+            'Do not map text-align: inherit'
+                => ['<p>plain</p>', 'p {text-align: interit;}', 'p', 'style'],
+            'Map margin: 0 auto to align=center'
+                => ['<img>', 'img {margin: 0 auto;}', 'img', 'align="center"'],
+            'Map float: right to align=right'
+                => ['<img>', 'img {float: right;}', 'img', 'align="right"'],
+            'Map border: none to border=0'
```
oliverklee: If I see this correctly, some of the mappings are still missing from the test.

oliverklee:
This PR is going to be really neat! :heart:

Classes/Emogrifier.php
sanderkruger: I think with a lookup table you don't need a switch or if/elseif. You just go through the levels of the lookup table. But I'm concerned that this function doesn't like to be in a lookup table.

For example, for most properties, you can apply the value to the attribute, but not for all. For margin and background, you need to parse the value of the property. For width and height too, because the 'px' must be stripped. I don't think there's any sane lookup mechanism to map the parse rules (or perhaps a regexp?).

For those properties whose value can be mapped, I think a lookup table is fine. As you suggested, a 2-dim array can do the trick, with one level for the property and a second for a whitelist of tag names.

I think this could work if the leafs of the lookup table can be both strings (the attribute name) and functions (a function to parse the property value and return an attribute-value pair).

Then the simple ones are captured with a simple string and the others with a function.

I'm not convinced this solution will be 'simpler' per se. But perhaps it will be easier to extend without getting lengthy.

oliverklee: I agree that it will be hard (or very complicated) to but the complete method in a look-up table. This is why I proposed having the two arrays which would allow us get rid of lots of additional &&s and convert the method to a switch.

After those refactorings, we can see whether an additional look-up table for the simple cases (without conversion) will make sense.

sanderkruger:
@oliverklee I modified the function to use a lookup table for the simple cases. Please have a look and tell me if this is what you had in mind.

sanderkruger:
@oliverklee I think something's wrong with travis. Have a look at the switch statement. I think it's fine. Travis is counting spaces as tabs for the switch coding rules, I think (look at the requirement of 13 tabs instead of 16, that's the number of spaces).

oliverklee:
I'll have a look at it. It could also be that the sniff just is the wrong one for PSR-2.

oliverklee:
The first switch statement of this project indeed has uncovered that our code sniffer required some formatting that contradicted the formatting required by PSR-2. ;-)

I've fixed this in #270. Could you please rebase?

Classes/Emogrifier.php
```diff
@@ -153,6 +153,41 @@ class Emogrifier
...
     /**
+     * Determines whether CSS styles that have an equivalent HTML attribute
+     * should be mapped and attached to those elements.
+     *
+     * @var bool
+     */
+    private $shouldMapCssToHtml = false;
+
+    /**
+     * This multi-level array contains simple mappings of css properties to
```
oliverklee: css -> CSS

Classes/Emogrifier.php
```diff
@@ -153,6 +153,41 @@ class Emogrifier
...
+     * HTML attributes. If a property maps simply to an attribute, the mapping
+     * can be a string, the name of the attribute. If a mapping only applies to
+     * certain HTML nodes or only for certain values, the mapping is an object
+     * with a whitelist of nodes and values.
+     *
+     * @var array
```
oliverklee: Please let's make this more specific: mixed[]

Classes/Emogrifier.php
```diff
@@ -329,6 +367,107 @@ protected function process(\DOMDocument $xmlDocument)
...
+     *
+     * @param string $property    the name of the CSS property to map.
+     * @param string $value       the value of the style rule to map.
+     * @param \DOMNode $node      node to apply styles to.
+     *
+     * @return void
+     */
+    private function mapCssToHtmlAttribute($property, $value, \DOMNode $node)
+    {
+        if (isset($this->cssToHtmlMap[$property])) {
```
oliverklee: Please let's move the first block into a private method canMapCssPropertyToHtml (or something similar).

Classes/Emogrifier.php
```diff
@@ -329,6 +367,107 @@ protected function process(\DOMDocument $xmlDocument)
...
+            case 'background':
+                // Parse out the color, if any
+                $styles = explode(' ', $value);
+                $first = $styles[0];
+                if (!is_numeric(substr($first, 0, 1)) && substr($first, 0, 3) !== 'url') {
+                    // This is not a position or image, assume it's a color
+                    $node->setAttribute('bgcolor', $first);
+                }
+                break;
+            case 'width':
```
oliverklee: Please add a comment (directly between the two cases) like this so it's clear to the reader that you haven't just forgotten the break:

// intentional fall-through

Classes/Emogrifier.php
```diff
@@ -329,6 +367,107 @@ protected function process(\DOMDocument $xmlDocument)
...
+                    }
+                }
+                break;
+            case 'border':
+                if ($value === 'none' || $value === '0') {
+                    $node->setAttribute('border', '0');
+                }
+                break;
+            default:
+                break;
```
oliverklee: The default case does not need a break statement.

Classes/Emogrifier.php
```diff
@@ -329,6 +367,107 @@ protected function process(\DOMDocument $xmlDocument)
...
+            case 'width':
+            case 'height':
+                // Remove 'px'. This regex only conserves numbers and %
+                $number = preg_replace('/[^0-9.%]/', '', $value);
+                $node->setAttribute($property, $number);
+                break;
+            case 'margin':
+                if ($node->nodeName === 'table' || $node->nodeName === 'img') {
+                    $margins = preg_split('/\s+/', $value);
```
oliverklee: Could you please use a double backslash here so it's clear to the reader that you're not escaping for the PHP string, but for the regular expression?

Classes/Emogrifier.php
```diff
@@ -329,6 +367,107 @@ protected function process(\DOMDocument $xmlDocument)
...
+                    return;
+                }
+                $attribute = $mapping['attribute'];
+            }
+            $node->setAttribute($attribute, $value);
+            return;
+        }
+        /** The property is not found in the 'simple-mapping' table. Check the
+         * more exotic cases
+         */
```
oliverklee: Could you please change this to a //-comment? /**-comments are for PHPDoc only (var attributes etc.).

sanderkruger:
@oliverklee I added a few more negative tests on elements that should not be mapped for complete coverage. And I improved the complex mapping method to use another method to parse out shorthand CSS values to their respective values top, bottom, left and right.

oliverklee:
@sanderkruger Great, thanks! I'll have a look at it.

(And thanks for notifying me. Unfortunately, GitHub does not send any notification if someone pushes a new version to a PR.)

oliverklee:
Hi @sanderkruger, I've just polished and merged your PR as #288. Thanks for contributing! :christmas_tree:

PR #258, opened 2015-10-12
[BUGFIX] Parsing CSS declaration blocks

bramley:
Revert to splitting declaration blocks using a regex.
Split on ; except when it is followed by charset or base64.
Make the validation of property/value pairs and surrounding white space more
robust.

Prior to the changes three tests failed
* nl before colon and after value, trailing semicolon
* one declaration with missing dash in property name
* one declaration with invalid character in property name

There seems to be an odd, to me, convention of using double backslashes when only one is needed (such as \s*), and also using backslash when it is not needed (such as for - in [A-Za-z-]. These changes haven't followed that convention.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -1209,19 +1209,17 @@ class Emogrifier
         }
 
         $properties = [];
+        $declarations = preg_split('/;(?!base64|charset)/', $cssDeclarationsBlock);
 
-        // Replace the semicolon in base64-encoded data URIs with a @ to prevent breaking
-        // up these declarations.
-        $mangledDeclarationsBlock = str_replace(';base64', '@base64', $cssDeclarationsBlock);
-        $declarations = explode(';', $mangledDeclarationsBlock);
         foreach ($declarations as $declaration) {
+            $declaration = trim($declaration);
             $matches = [];
-            if (!preg_match('/ *([A-Za-z\\-]+) *: *([^;]+) */', $declaration, $matches)) {
+
+            if (!preg_match('/^([A-Za-z\\-]+)\\s*:\\s*(.+)$/', $declaration, $matches)) {
                 continue;
             }
             $propertyName = strtolower($matches[1]);
-            // Put the semicolon back in front of the base64 declaration
-            $propertyValue = str_replace('@base64', ';base64', $matches[2]);
+            $propertyValue = $matches[2];
             $properties[$propertyName] = $propertyValue;
         }
         $this->caches[self::CACHE_KEY_CSS_DECLARATIONS_BLOCK][$cssDeclarationsBlock] = $properties;
--- a/Tests/Unit/EmogrifierTest.php
+++ b/Tests/Unit/EmogrifierTest.php
@@ -679,7 +679,7 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
     /**
      * Data provider for emogrifyDropsWhitespaceFromCssDeclarations.
      *
-     * @return array[]
+     * @return string[][]
      */
     public function cssDeclarationWhitespaceDroppingDataProvider()
     {
@@ -692,6 +692,9 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
             'space before trailing semicolon' => [' color:#000 ;', 'color: #000;'],
             'space after trailing semicolon' => [' color:#000; ', 'color: #000;'],
             'space after property value, no trailing semicolon' => [' color:#000; ', 'color: #000;'],
+            'newline before property value, trailing semicolon' => ["\ncolor:#222; ", 'color: #222;'],
+            'newline before colon and after value, trailing semicolon' => ["color\n:#333\n; ", 'color: #333;'],
+            'newline after colon, trailing semicolon' => ["color:\n#333; ", 'color: #333;'],
         ];
     }
 
@@ -722,7 +725,7 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
     /**
      * Data provider for emogrifyFormatsCssDeclarations.
      *
-     * @return array[]
+     * @return string[][]
      */
     public function formattedCssDeclarationDataProvider()
     {
@@ -735,6 +738,9 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
                 => ['color: #000; width: 3px;', 'color: #000; width: 3px;'],
             'two declaration separated by semicolon and Linefeed' => [
                 'color: #000;' . PHP_EOL . 'width: 3px;', 'color: #000; width: 3px;'
+            ],
+            'one declaration with leading dash in property name' => [
+                '-webkit-text-size-adjust:none;', '-webkit-text-size-adjust: none;'
             ],
         ];
     }
@@ -757,6 +763,42 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
 
         self::assertContains(
             'html style="' . $expectedStyleAttributeContent . '">',
+            $this->subject->emogrify()
+        );
+    }
+
+    /**
+     * Data provider for emogrifyInvalidDeclaration.
+     *
+     * @return string[][]
+     */
+    public function invalidDeclarationDataProvider()
+    {
+        return [
+            'missing dash in property name' => ['font weight: bold;'],
+            'invalid character in property name' => ['-9webkit-text-size-adjust:none;'],
+            'missing :' => ['-webkit-text-size-adjust none'],
+            'missing value' => ['-webkit-text-size-adjust :'],
+        ];
+    }
+
+    /**
+     * @test
+     *
+     * @param string $cssDeclarationBlock the CSS declaration block (without the curly braces)
+     *
+     * @dataProvider invalidDeclarationDataProvider
+     */
+    public function emogrifyDropsInvalidDeclaration($cssDeclarationBlock)
+    {
+        $html = $this->html5DocumentType . '<html></html>';
+        $css = 'html {' . $cssDeclarationBlock . '}';
+
+        $this->subject->setHtml($html);
+        $this->subject->setCss($css);
+
+        self::assertContains(
+            '<html style="">',
             $this->subject->emogrify()
         );
     }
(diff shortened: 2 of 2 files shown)


oliverklee:
Hi @bramley, thanks a bunch for the pull request! :heart:

We've just merged two other PRs that conflict with yours, though. Could you please update your master to the upstream master and then rebase?

And you could also please prefix the commit message subject with [BUGFIX] and uppercase the CSS? I'd like to have a nice and clean commit history, and this would allow me to merge the PR directly here on GitHub instead of having to check out the branch locally, amend the commit and then create a new PR for that (as I cannot edit your PR directly). Thanks!

oliverklee:
About the double backslashes: For regular expressions, we use double backslashes for two reasons:

a) This makes it clear to the reader that the backslash is intended to be an escape character for the regular expression, not for creating the PHP string. So this makes the intent of the code more clear.
b) In some cases, having only one backslash breaks things (e.g., \n in double-quoted strings or when you need an escaped backslash in a regular expression (which would require four backslashes)). _Always_ escaping the backslash in strings unless it is needed as an escape character for PHP string creation removes this problem.

You are right that the (double) backslash before the trailing minus in the character class (i.e., in the square brackets) is not technically required. The backslash is there merely to indicate to the reader that the minus is there not by error, but by intention. (So it's about communicating the intent again.)

Tests/Unit/EmogrifierTest.php
```diff
@@ -1848,7 +1891,8 @@ public function dataUrisAreConserved()
     {
         $html = $this->html5DocumentType . '<html></html>';
         $this->subject->setHtml($html);
-        $styleRule = 'background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAUCAIAAAAC64paAAABUk' .
+        $styleRule = 'background-image: url(data:image/png;charset=utf-8;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAUC' .
```
oliverklee: If I understand https://developer.mozilla.org/en-US/docs/Web/HTTP/data_URIs correctly, the following cases should be valid:

data:image/png,iVBORw0KG...
data:image/png;charset=utf-8,iVBORw0KG...
data:image/png;base64,iVBORw0KG...
data:image/png;charset=utf-8;base64,iVBORw0KG...

It'd like to have all these covered by tests. I propose having a data provider for the part between "data:image/png" and the base64-encoded data.

Tests/Unit/EmogrifierTest.php
```diff
@@ -750,6 +756,43 @@ public function emogrifyFormatsCssDeclarations($cssDeclarationBlock, $expectedSt
     }
 
     /**
+     * Data provider for emogrifyInvalidDeclaration.
+     *
+     * @return array[]
```
oliverklee: Please let's make this more specific: string[][] (even if some other return annotations still are using the not-so-specific type)

Tests/Unit/EmogrifierTest.php
```diff
@@ -750,6 +756,43 @@ public function emogrifyFormatsCssDeclarations($cssDeclarationBlock, $expectedSt
...
+    {
+        $html = $this->html5DocumentType . '<html></html>';
+        $css = 'html {' . $cssDeclarationBlock . '}';
+
+        $this->subject->setHtml($html);
+        $this->subject->setCss($css);
+
+        self::assertNotContains(
+            $notExpectedStyleAttributeContent,
+            $this->subject->emogrify()
```
oliverklee: Actually, there should be no style attribute at all if the parsing is correct. So I propose to have an assertContains with '<html>' instead.

Tests/Unit/EmogrifierTest.php
```diff
@@ -750,6 +756,43 @@ public function emogrifyFormatsCssDeclarations($cssDeclarationBlock, $expectedSt
...
+
+    /**
+     * @test
+     *
+     * @param string $cssDeclarationBlock the CSS declaration block (without the curly braces)
+     * @param string $notExpectedStyleAttributeContent the value that should not be withint the style attribute
+     *
+     * @dataProvider invalidDeclarationDataProvider
+     */
+    public function emogrifyInvalidDeclaration($cssDeclarationBlock, $notExpectedStyleAttributeContent)
```
oliverklee: Please let's make the test names sentences about the behavior. My suggestion: emogrifyDropsInvalidCssDeclarations

Tests/Unit/EmogrifierTest.php
```diff
@@ -750,6 +756,43 @@ public function emogrifyFormatsCssDeclarations($cssDeclarationBlock, $expectedSt
...
+     *
+     * @return array[]
+     */
+    public function invalidDeclarationDataProvider()
+    {
+        return [
+            'one declaration with missing dash in property name' => ['font weight: bold;', 'weight:'],
+            'one declaration with invalid character in property name' => ['-9webkit-text-size-adjust:none;', 'webkit'],
+            'one declaration with missing :' => ['-webkit-text-size-adjust none', 'webkit'],
+            'one declaration with missing value' => ['-webkit-text-size-adjust :', 'webkit'],
```
oliverklee: I think the prefix "one declaration with" is not really necessary here in the descriptions.

Tests/Unit/EmogrifierTest.php
```diff
@@ -680,6 +680,9 @@ public function cssDeclarationWhitespaceDroppingDataProvider()
             'space before trailing semicolon' => [' color:#000 ;', 'color: #000;'],
             'space after trailing semicolon' => [' color:#000; ', 'color: #000;'],
             'space after property value, no trailing semicolon' => [' color:#000; ', 'color: #000;'],
+            'nl before property value, trailing semicolon' => ["\ncolor:#222; ", 'color: #222;'],
+            'nl before colon and after value, trailing semicolon' => ["color\n:#333\n; ", 'color: #333;'],
+            'nl after colon, trailing semicolon' => ["color:\n#333; ", 'color: #333;'],
```
oliverklee: Could you please change the "nl" to "NL" or (even better) "newline" in the (new) descriptions?

Tests/Unit/EmogrifierTest.php
bramley: Like the other data providers it is an array of associative arrays whose values are string. Elsewhere string[][] is used.

oliverklee: "an array of (associative) arrays whose values are string" is what string[][] means (mind the two pairs of square brackets). :-)

(We'll also try to get rid of all param/return array and make them more specific: #195)

oliverklee: Oh, looks like I missed a pair of square brackets in my first comment. Sorry for the confusion I've caused. :confused:

oliverklee: Oh, the second pair of brackets in my first comment was eaten by GitHub's MarkDown renderer. :-(

Tests/Unit/EmogrifierTest.php
bramley: It actually generates an empty style attribute
`<html style="">`

Tests/Unit/EmogrifierTest.php
bramley: This looks to be focusing on data uris, whereas the code is just providing some exceptions when splitting on ';'. I think that the current test is sufficient for now. If in the future other special cases arise then they would need to be added.

Tests/Unit/EmogrifierTest.php
bramley: Changed

Tests/Unit/EmogrifierTest.php
bramley: Changed

Tests/Unit/EmogrifierTest.php
bramley: Changed

bramley:
I think the current version is now ok.

Tests/Unit/EmogrifierTest.php
oliverklee: Okay, the style node then should not be there in the first place. Let's tackle this separately from this PR. I've entered an issue for this: #259

Tests/Unit/EmogrifierTest.php
oliverklee: Sounds fair. I've entered issue #260 for this.

Classes/Emogrifier.php
```diff
@@ -1209,19 +1209,17 @@ private function parseCssDeclarationsBlock($cssDeclarationsBlock)
...
 
         $properties = [];
+        $declarations = preg_split('/;(?!base64|charset)/', $cssDeclarationsBlock);
 
-        // Replace the semicolon in base64-encoded data URIs with a @ to prevent breaking
-        // up these declarations.
-        $mangledDeclarationsBlock = str_replace(';base64', '@base64', $cssDeclarationsBlock);
-        $declarations = explode(';', $mangledDeclarationsBlock);
         foreach ($declarations as $declaration) {
+            $declaration = trim($declaration);
```
oliverklee: Please don't overwrite variables because then the same variable can have two different meanings depending on how far down in the method it is used. This makes understanding the method harder and also makes moving code around harder.

Proposed solution 1: use a new variable $trimmedDeclaration
Proposed solution 2: inline $declaration in the preg_match (the only place where it is used AFAICS)
Proposed solution 3: drop the trim altogether and modify the preg_match to also cope with leading and trailing whitespace

Tests/Unit/EmogrifierTest.php
```diff
@@ -692,6 +692,9 @@ public function cssDeclarationWhitespaceDroppingDataProvider()
             'space before trailing semicolon' => [' color:#000 ;', 'color: #000;'],
             'space after trailing semicolon' => [' color:#000; ', 'color: #000;'],
             'space after property value, no trailing semicolon' => [' color:#000; ', 'color: #000;'],
+            'newline before property value, trailing semicolon' => ["\ncolor:#222; ", 'color: #222;'],
+            'newline before colon and after value, trailing semicolon' => ["color\n:#333\n; ", 'color: #333;'],
```
oliverklee: Please let's split this up in two separate lines:
1. 'newline before colon, trailing semicolon'
2. 'newline after value, trailing semicolon'

We also should have a test for the case of a newline after the semicolon.

Tests/Unit/EmogrifierTest.php
```diff
@@ -762,6 +768,42 @@ public function emogrifyFormatsCssDeclarations($cssDeclarationBlock, $expectedSt
...
+            'invalid character in property name' => ['-9webkit-text-size-adjust:none;'],
+            'missing :' => ['-webkit-text-size-adjust none'],
+            'missing value' => ['-webkit-text-size-adjust :'],
+        ];
+    }
+
+    /**
+     * @test
+     *
+     * @param string $cssDeclarationBlock the CSS declaration block (without the curly braces)
```
oliverklee: For the parameter, either it should be plural (declarationsBlock) or just "declaration" (as we always have only one declaration for this test case). (In MDN (which I use as reference for this), they use the plural term: https://developer.mozilla.org/de/docs/Web/CSS/Syntax)

oliverklee:
Almost there!

Tests/Unit/EmogrifierTest.php
oliverklee: The point is to have the tests (or data provider items) as fine-grained as possible so that you know exactly what part of the behavior is broken when something fails.

bramley:
Sorry but I have now run out of enthusiasm for this. Please feel free to fix whatever you think is wrong.

oliverklee:
@bramley Please don't be frustrated by the reviews. I really appreciate your PR and want it the best it can be before I merge it.

I'll take care of the two open to-dos and then merge your PR in a minute.

Classes/Emogrifier.php
oliverklee: done

Tests/Unit/EmogrifierTest.php
oliverklee: done

Tests/Unit/EmogrifierTest.php
oliverklee: done

oliverklee:
@bramley I've fixed the remaining issues and merged your PR as #261. Thanks for contributing to Emogrifier! :heart:

PR #279, opened 2015-10-27
Proposal: better formatting of preg_replace rules in translates ToXpath

OzzyCzech:
Check way of formatting preg_replace rules that we use in our projects

This PR is just proposal - if you don't like that, never mind :-D close it without merge...

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -1022,28 +1022,26 @@ class Emogrifier
         $trimmedLowercaseSelector = trim($lowercasePaddedSelector);
         $xPathKey = md5($trimmedLowercaseSelector);
         if (!isset($this->caches[self::CACHE_KEY_XPATH][$xPathKey])) {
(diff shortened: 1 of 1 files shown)


oliverklee:
Hi @OzzyCzech, thanks for the PR! I like the new formatting, and I would also like to keep the descriptions (which currently are the array keys). Any ideas how to combine both?

OzzyCzech:
I am add extra space to be easily read, what do you thing about that?

oliverklee:
I don't think we need the extra spacing or the blank lines as they spread the code too far to be readble. However, I really like the comment lines. :-)

So I propose getting rid of the extra spaces while keeping the comment lines (on separate lines so we don't have any end-of-line comments).

OzzyCzech:
done...

oliverklee:
Hi @OzzyCzech, I've polished your PR a bit and merged it as #281. Thanks a bunch for contributing! :heart:

PR #283, opened 2015-12-11
"!important" style tag fix

white43:
I've found something that looks like a bug.

```html
<html>
<head>
    <style>
    p { margin: 1px !important; }
    p { margin: 2px !important; }
    </style>
</head>
<body>
    <p>Text here</p>
</body>
</html>
```
After precessing with emogrifier it turns into:

```html
<!DOCTYPE html>
<html>
<head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head>
<body>
    <p style="margin: 1px !important;">Text here</p>
</body>
</html>
```

So i've made quick fix and added one more test. Hope it was helpful to you.


oliverklee:
Hi @white43, I've polished and merged your PR as #292. Thanks for contributing! :christmas_tree:

PR #293, opened 2015-12-27
[BUGFIX] First-child and last-child selectors are broken

### LINKED ISSUE #286: first-child and last-child selectors are broken
There was a basic misunderstanding of what the selector first-child means in the fix for issue  https://github.com/jjriv/emogrifier/pull/257

From https://developer.mozilla.org/en-US/docs/Web/CSS/%3Afirst-child
The :first-child CSS pseudo-class represents any element that is the first child element of its parent.

The code, and the unit tests, think that it means the first child of the named element.

        // :first-child
        '/([^\\/]+):first-child/i'                 => '\\1/*[1]',

            'BODY:first-child matches first child'
                => ['body:first-child {' . $styleRule . '} ', '#<p class="p-1" style="' . $styleRule . '">#'],

Similarly for last-child, and possibly other similar selectors.

oliverklee:
Fixes #286

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 
 ### Fixed
+- First-child and last-child selectors are broken
+  ([#293](https://github.com/jjriv/emogrifier/pull/293))
 - Second !important rule needs to overwrite the first one
   ([#292](https://github.com/jjriv/emogrifier/pull/292))
 
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -165,9 +165,9 @@ class Emogrifier
         // descendant
         '/\\s+/'                                   => '//',
         // :first-child
-        '/([^\\/]+):first-child/i'                 => '\\1/*[1]',
+        '/([^\\/]+):first-child/i'                 => '*[1]/self::\\1',
         // :last-child
-        '/([^\\/]+):last-child/i'                  => '\\1/*[last()]',
+        '/([^\\/]+):last-child/i'                  => '*[last()]/self::\\1',
         // attribute only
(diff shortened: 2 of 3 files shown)


oliverklee:
I've used the XPath equivalents as listed on https://en.wikibooks.org/wiki/XPath/CSS_Equivalents.

@bramley: Could you please test this PR and also verify it by reading?

bramley:
The XPaths are not correct. The first-child one is selecting *all* elements that are the first child of their parent. Similarly the last-child is selecting all elements that are the last child of their parent.

The Symfony css selector package generates this XPath for p:first-child

    descendant-or-self::*/*[name() = 'p' and (position() = 1)]

While that could be generated by the preg_replace, the XPath expression can be more complex so I don't think it is easy (or even possible) to process a general expression with a preg_replace, e.g. p.red:first-child generates

    descendant-or-self::div/*[@class and contains(concat(' ', normalize-space(@class), ' '), ' red ') and (name() = 'p') and (position() = 1)]

As there can be subtle differences between similar XPath expressions, I'm not sure it is a good idea to try to implement css to XPath conversion without having a good knowledge of both. Particularly as there is a proven package that already does that. Wouldn't it be better to treat this as "missing functionality" or as a bug and fix it by using the Symfony component?

oliverklee:
Hi @bramley, thanks for the remarks. I've fixed the code and added some tests that fail with the old version of the patch. Could you please re-review? Thanks!

bramley:
Yes, this looks better. 
You might want to change these two test descriptions to refer to the div element, not p

            'P:first-child matches not first child with mismatching tag'
                => ['div:first-child {' . $styleRule . '} ', '#<p class="p-1">#'],

            'P:last-child not matches last child with mismatching tag'
                => ['div:last-child {' . $styleRule . '} ', '#<p class="p-3">#'],

oliverklee:
@bramley Good catch! Like this (re-pushed)?

PR #295, opened 2016-01-24
Removed references to mbstring dependency

demis-palma:
Updated composer, Readme and changelog to reflect the fact that mbstring is no longer a dependency.


oliverklee:
Hi @demis-palma, I've just polished and merged your PR as #297. Thanks for contributing! :heart:

demis-palma:
You are welcome.

PR #298, opened 2016-02-07
Add normalize-space to compare class values correctly and test

mikeswain:
Hi I noticed that in a lot of PHP generate markup  it inserts line feeds in HTML class attributes because people start a newline in the middle of strings to add a <?php ... ?> section. In your test against the class attributes using normalize-space to remove these should be the quickest way. Unit test included ...


oliverklee:
Hi @mikeswain, thanks for your PR! I'll review it now and add my comments. Could you then please re-push the updated branch (push -f) and add a comment that you've pushed an update? Thanks!

oliverklee:
Could you please also add a line to the changelog (newest entries first)?

Tests/Unit/EmogrifierTest.php
```diff
@@ -2067,4 +2067,20 @@ public function emogrifierNotMapsCssToHtmlIfFeatureIsNotEnabled()
             $html
         );
     }
+ 
+    /**
+     * @test
+     */
+    public function emogrifierMapsClassWithLineFeedInAttrib()
```
oliverklee: Attrib -> Attribute

oliverklee: We should test all cases (preferably with a data provider):

- leading space
- trailing space
- leading NL
- trailing NL
- leading tab
- trailing tab
- leading CRLF
- trailing CRLF

Tests/Unit/EmogrifierTest.php
```diff
@@ -2067,4 +2067,20 @@ public function emogrifierNotMapsCssToHtmlIfFeatureIsNotEnabled()
...
+    {
+        $css = '.test {color: red;}';
+        $this->subject->setHtml($this->html5DocumentType . "<html><body><div class='test\n'></div></body></html>");
+        $this->subject->setCss($css);
+
+        $html = $this->subject->emogrify();
+
+	self::assertContains('style="color: red;', $html );
```
oliverklee: Could you please add the closing quote, too?

Tests/Unit/EmogrifierTest.php
```diff
@@ -2067,4 +2067,20 @@ public function emogrifierNotMapsCssToHtmlIfFeatureIsNotEnabled()
...
         );
     }
+ 
+    /**
+     * @test
+     */
+    public function emogrifierMapsClassWithLineFeedInAttrib()
+    {
+        $css = '.test {color: red;}';
+        $this->subject->setHtml($this->html5DocumentType . "<html><body><div class='test\n'></div></body></html>");
```
oliverklee: Please use double (escaped) quotes for the attribute value so this matches the way HTML attributes are usually written.

Classes/Emogrifier.php
```diff
@@ -1303,9 +1303,9 @@ private function matchIdAttributes(array $match)
      */
     private function matchClassAttributes(array $match)
     {
-        return ($match[1] !== '' ? $match[1] : '*') . '[contains(concat(" ",@class," "),concat(" ","' .
+        return ($match[1] !== '' ? $match[1] : '*') . '[contains(concat(" ",normalize-space(@class)," "),concat(" ","' .
             implode(
-                '"," "))][contains(concat(" ",@class," "),concat(" ","',
+                '"," "))][contains(concat(" ",normalize-space(@class)," "),concat(" ","',
                 explode('.', substr($match[2], 1))
             ) . '"," "))]';
```
oliverklee: Are currently both changes covered by the unit test, i.e., will the test still fail if I remove either of the changes?

Classes/Emogrifier.php
```diff
@@ -1303,9 +1303,9 @@ private function matchIdAttributes(array $match)
      */
     private function matchClassAttributes(array $match)
     {
-        return ($match[1] !== '' ? $match[1] : '*') . '[contains(concat(" ",@class," "),concat(" ","' .
+        return ($match[1] !== '' ? $match[1] : '*') . '[contains(concat(" ",normalize-space(@class)," "),concat(" ","' .
```
oliverklee: We have this change in two places now. Could you please move it to a local variable to avoid the duplication (as a small refactoring)?

Classes/Emogrifier.php
mikeswain: I could take it out the second one coz it doesn't break the test - I haven't managed to get it to break, but experience tells me it probably needs it

Classes/Emogrifier.php
mikeswain: It can't go in a local variable as its part of an xpath- its actually returning an expression to be evaluated in another context. The xpath evaluator should however optimize it away anyway.

Tests/Unit/EmogrifierTest.php
mikeswain: will do

Classes/Emogrifier.php
oliverklee: Then we'll need a test that breaks without that change. (Otherwise, the change is not needed, or our tests don't test enough.)

oliverklee: @mikeswain Do we now have at least one test that would break without this change?

Tests/Unit/EmogrifierTest.php
```diff
@@ -1759,6 +1759,7 @@ public function addExcludedSelectorKeepsNonMatchingElementsInEmogrification()
     }
 
     /**
+    /**
```
oliverklee: Could you please undo this indentation change?

oliverklee:
@mikeswain Any news on this PR?

oliverklee:
Closing this PR as there haven't been any changes. @mikeswain Could you please reopen this PR once you've done the necessary changes and re-pushed? Thanks!

Issue #237, opened 2015-09-30
Noisy error on malformed CSS


Synchro:
Much like #179, I've been getting this from something malformed:
```
2015-09-30 14:43:49 DOMXPath::query(): Invalid predicate
#0 System_ErrorHandler->ERROR_HANDLER
#1 DOMXPath->query line 287
#2 Pelago\Emogrifier->process line 211
#3 Pelago\Emogrifier->emogrify line 44
```

That line is:

    $nodesMatchingCssSelectors = $xpath->query($this->translateCssToXpath($selector['selector']));

So it may well be fixed by a fix for #179 by eliminating the bogus selector, but it shouldn't throw a noisy error like this anyway. Better to just ignore the content. When I've tracked down the content responsible for this error, I'll add it in here so we can build a test for it.

oliverklee:
What CSS is causing this problem?

Renkas:
I got the same problem.

 in /var/www/vendor/pelago/emogrifier/Classes/Emogrifier.php – DOMXPath::query('//a[href^="tel"]') at line 723

oliverklee:
Okay, so it looks like Emogrifier currently does not support negative attribute selectors (i.e., this is missing feature). Anyone willing to implement this?

Synchro:
That just sounds like XPath and CSS selectors do not share the same syntax for that kind of lookup. I suspect in my case it was simply malformed CSS rather than a missing feature - e.g. if someone puts HTML tags in the CSS, which I've seen. From the source of my CSS, I strongly suspect it's just random junk rather than anything as sophisticated as this missing selector, so the rename may not be quite appropriate.

oliverklee:
Okay, I've reverted the changes to this ticket. Thanks for the input! (We already have #227 for the missing feature anyway.)

oliverklee:
@Renkas: So it looks like you're experiencing #227.

Synchro:
I've managed to track down some CSS that's causing this error - it's an empty media query.

In the Emogrifier class, the `extractMediaQueriesFromCss` method finds the media queries, but having an empty one makes it go out of sync and it starts treating a selector as properties.

I've set up the content and the regex it uses [here](https://regex101.com/r/fQ8kA9/1), and you can see it going wrong.

Following on from that, when it gets to `existsMatchForCssSelector`, the first CSS selector it passes is:

```
}

table.body
```

Which is invalid, but still gets translated into the invalid XPath:

```xpath
[...]

PR #306, opened 2016-04-28
Fix for #237

Synchro:
This revised method doesn't have problems with empty media queries and has no problems with my strange stylesheets! I also simplified the loop syntax a bit. Ideally this would have a test case that contains an empty media query, but I can't figure out your test system.


Synchro:
Ah, dang, I've still got some old commits in here...

PR #307, opened 2016-04-28
Proposed fix for #237

Synchro:
Resubmission of a clean version of #306

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -924,6 +924,8 @@ class Emogrifier
 
     /**
      * Extracts the media queries from $css.
+     * Skips empty media queries.
+     * @link http://stackoverflow.com/questions/36910664/recursive-subroutine-regex-to-match-css-media-queries/36911319#36911319
      *
      * @param string $css
      *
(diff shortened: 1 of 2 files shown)


Synchro:
The build failure is an unrelated composer problem.

oliverklee:
I'll have a look at the composer problem.

oliverklee:
Hi @Synchro, could you also please add regression tests for this (i.e., unit tests that fail without the fix and that pass with it)? Thanks!

Synchro:
This test works for me - but another started failing. It turned out to be because it contained invalid markup so I fixed that.

oliverklee:
@Synchro I've just fixed the Composer-related breakage on master. Could you please rebase, force-push and then comment here? Thanks!

Synchro:
I've rebased, squashed and pushed.

Classes/Emogrifier.php
```diff
@@ -924,21 +924,26 @@ private function copyCssWithMediaToStyleNode(\DOMDocument $xmlDocument, \DOMXPat
...
-        preg_match_all('#(?<query>@media[^{]*\\{(?<css>(.*?)\\})(\\s*)\\})#s', $css, $mediaQueries);
+        preg_match_all('/@media\b[^{]*({((?:[^{}]+|(?1))*)})/', $css, $mediaQueries, PREG_SET_ORDER);
         $result = [];
-        foreach (array_keys($mediaQueries['css']) as $key) {
-            $result[] = [
-                'css' => $mediaQueries['css'][$key],
-                'query' => $mediaQueries['query'][$key],
-            ];
+        foreach ($mediaQueries as $query) {
+            if (!empty(trim($query[2]))) {
```
oliverklee: You'll need to put trim(...) into a local variable so the empty call also works in PHP 5.4:
https://travis-ci.org/jjriv/emogrifier/jobs/128679215#L181

Synchro:
Should work now!

oliverklee:
Hi @Synchro, I've polished your PR a bit and pushed it as #319. Could you please check the polished version and comment there? Thanks!

PR #316, opened 2016-06-09
Adds support to PHP 5.3 converting all `[]` declarations to `array()`.

robertotremonti:


oliverklee:
Hi @robertotremonti, thanks for the PR!

However, as stated in the composer.json, Emogrifier requires at least PHP 5.4. This was a conscious decision as PHP 5.3 had its "end of life" (i.e., not security fixes anymore) in August 2014, and 5.4 in September 2015. If your site is reachable over the internet, please upgrade to a supported version of PHP as soon as possible to keep your site from getting hacked.

(And if you really need to use PHP 5.3., you can use Emogrifier v0.1, which is the last version to support PHP 5.3.)

robertotremonti:
Thanks for tour reply, @oliverklee.
I understand your point of view, but I think that just changing `[]` to `array()` is not a big deal.
No need to change other parts of code in order to support PHP 5.3.

robertotremonti:
`emogrifyBodyContent()` method is not available.
I didn't find a way to get body content.

oliverklee:
If you need the latest features, I recommend using a version of PHP that has not reached its end of life. (The more PHP versions we support in Emogrifier, the more work maintaining the code is, and the less new PHP features we can use. So I usually remove support for a PHP version when (or shortly after) it has reached its end of life.)

robertotremonti:
Thanks for your clarification.

PR #308, opened 2016-05-04
[BUG] Ignore pseudo elements when combined with pseudo selector

jorrit:
Ignores CSS statements of the form `el:pseudoselector:pseudoelement` where pseudoselector is a supported selector such as `:last-child` and pseudoelement is an unsupported selector such as `:after`.

PR #318, opened 2016-06-13
[BUGFIX] Ignore pseudo-class when combined with pseudo-element

oliverklee:
Ignores CSS statements of the form "el:pseudoclass::pseudoelement"
where "pseudoclass" is a supported selector such as ":last-child"
and pseudoelement is an unsupported selector such as "::after".

Fixes #308

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


jorrit:
My use case was in the test. If the test still succeeds and the other tests are OK too, than it is fine with me!

oliverklee:
Great, thanks for the feedback - and thanks for contributing! ❤️

PR #320, opened 2016-07-01
[BUGFIX] Fix media regex parsing

markcarver:
The current regex doesn't take into account possible minification.

Also, the `print` media query was never added to the `$media` variable and stripped from `$cssForAllowedMediaTypes`, thus causing print attributes to be applied to the DOM elements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Silence purposefully ignored PHP Warnings
   ([#400](https://github.com/MyIntervals/emogrifier/pull/400))
+- Fix media regex parsing
+  ([#320](https://github.com/MyIntervals/emogrifier/pull/320))
 
 
 ### Security
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -1100,7 +1100,7 @@ class Emogrifier
 
         $media = '';
         $cssForAllowedMediaTypes = preg_replace_callback(
-            '#@media\\s+(?:only\\s)?(?:[\\s{\\(]' . $mediaTypesExpression . ')\\s?[^{]+{.*}\\s*}\\s*#misU',
+            '#@media\\s+(?:only\\s)?(?:[\\s{\\(]\\s*' . $mediaTypesExpression . ')\\s*[^{]*{.*}\\s*}\\s*#misU',
             function ($matches) use (&$media) {
                 $media .= $matches[0];
             },
@@ -1110,7 +1110,7 @@ class Emogrifier
         // filter the CSS
         $search = [
             'import directives' => '/^\\s*@import\\s[^;]+;/misU',
-            'remaining media enclosures' => '/^\\s*@media\\s[^{]+{(.*)}\\s*}\\s/misU',
+            'remaining media enclosures' => '/\\s*@media\\s+[^{]+{(.*)}\\s*}\\s*/misU',
         ];
 
         $cleanedCss = preg_replace($search, '', $cssForAllowedMediaTypes);
--- a/Tests/Unit/EmogrifierTest.php
+++ b/Tests/Unit/EmogrifierTest.php
@@ -1147,7 +1147,15 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
             'style in "only screen" media type rule' => ['@media only screen { h1 { color:red; } }'],
             'style in "only all" media type rule' => ['@media only all { h1 { color:red; } }'],
             'style in "screen" media type rule' => ['@media screen { h1 { color:red; } }'],
+            'style in "print" media type rule' => ['@media print { * { color:#000 !important; } }'],
             'style in media type rule without specification' => ['@media { h1 { color:red; } }'],
+            'style with multiple media type rules' => [
+                '@media all { p { color: #000; } }' .
+                '@media only screen { h1 { color:red; } }' .
+                '@media only all { h1 { color:red; } }' .
+                '@media print { * { color:#000 !important; } }' .
+                '@media { h1 { color:red; } }'
+            ],
         ];
     }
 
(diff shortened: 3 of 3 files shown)


oliverklee:
Hi @markcarver, thanks for your PR!

Could you please add these things so I can merge the PR:

1. a changelog line (newest on top)
2. unit tests (at least one test for each fixed bug) that fail without the fix and that pass with the fixes

When doing these changes, I recommend doing a git commit --amend so there will still be only one commit per PR (which makes the merging less work for me).

Thanks again for contributing! Looking forward to the additional changes! ❤️

markcarver:
I'm not good with tests

oliverklee:
Or could you post some examples of code and the output with and without the fixes?

oliverklee:
And testing also is not very hard to learn. You can start by reading and copy'n'pasting the existing tests.

markcarver:
```php
$html = '<div class="navbar"></div>';
$css = '@media print{*{color:#000!important}.navbar{display:none}}';
$emogrifier = new Emogrifier($html, $css);
$output = $emogrifier->emogrify();
```

Actual output:
```html
<!DOCTYPE html>
<html>
<head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head>
<body></body>
</html>
```

Expected output:
```html
<!DOCTYPE html>
<html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
<style type="text/css">@media print{*{color:#000!important}.navbar{display:none}}</style>
</head>
<body><div class="navbar"></div></body>
</html>
```

oliverklee:
We've moved the repository to https://github.com/MyIntervals/emogrifier . If you have an "upstream" branch in your fork, please make sure to update the URL:

`git remote set-url upstream git@github.com:MyIntervals/emogrifier.git`

markcarver:
I have never understood people who "refuse" PRs because there's no tests. I have clearly explained the problem and offered a solution over a year ago.

If you want the test so bad, do it yourself. I'm not going to learn yet another test tooling used just to satisfy this stupid "requirement".

oliverklee:
@markcarver, when we accept a PR to an open source project, we agree to maintain the new/changed code practically forever. This includes that we want to make sure that the changed code works correctly in the first place, and that it also keeps working (i.e., it does not get broken by other changes). Having the changes covered by automated tests is the only way to ensure that.

(And as far as writing the tests is concerned: Yes, of course, I could do that (and probably will sometime), but I do this in my spare time, and this bug fix is not very important to me personally. So if it's important to you, providing unit tests will definitely help get it merged sooner.)

I hope that this explanation helps you understand this policy better. ☕️

markcarver:
I've updated this PR with changes to the changelog + tests.

---

> when we accept a PR to an open source project, we agree to maintain the new/changed code practically forever.

I'm well aware of how open source works. I contribute and maintain several open source projects.

> Having the changes covered by automated tests is the only way to ensure that.

I'm not disputing this fact at all. What I'm disputing is whose responsibility it is to ensure that.

> I do this in my spare time, and this bug fix is not very important to me personally.

I think you're failing to see my point of view.

I do this in my spare time as well.

What probably takes you no time at all (because you're familiar with this project) takes me 10x longer (or more).

I am, primarily, a front-end developer. Backend/PHP isn't my strong suit, especially when it comes to tests. One could say I'm a unicorn though and can muddle my way through these kinds of things (which is how I found and fixed this bug).

Case in point:

I finally caved and decided to just make the test so you'll just commit the PR (for a fix I have already solved and grow tired of patching).

However, I just spent 2 hours trying to figure out why the tests (from HEAD) were spitting out PHP Warnings, only to find out that this is technically expected? So I created #400 for you to fix that (will likely have to rebase one or the other because of the changelog commits, but meh, whatever).

Second, the "tests" are already over 2k+ LOC. You really expect me, someone, who doesn't really know the intricacies of PHP Unit testing and the nuances of this specific project's helper methods and whatnot, to just "jump in" and just add something in the "right way"?

I also just spent several hours just trying to figure out how some of this testing stuff works (`@dataProvider` magic???). I originally created two separate tests only to figure out that I could just add stuff to `EmogrifierTest::validMediaPreserveDataProvider` and then add a separate test to mock minification of the same thing.

As far as the changelog... that's really the responsibility of a project maintainer and there are easy ways to automate that.

These kinds of responsibilities shouldn't be put on someone who's simply trying to fix some code.

Developers who help contribute, in any small way, should not be subjugated to perform tasks meant for the maintenance of a project (i.e. outsourcing maintainership).

---

So yes, you mentioned "time" and I understand that your time is valuable, but so is mine.

It's about mutual respect and understanding how to, efficiently, allocate time for all parties involved.

You could have added these tests in a fraction of a second, but because it's "personally not important to you", you forced me to waste more of my time/day in "learning" this project's tests.

I had other plans today, but thanks for explaining that only your time is important and not mine.

markcarver:
Rebased now that #400 is merged

oliverklee:
Thanks! I'll have a look at it.

oliverklee:
Polished an merged in #402. Thanks for contributing, @markcarver! :heart:

oliverklee:
@markcarver A few words regarding your comments.

So we both maintain open source projects, and we agree that automated tests are very important.

We disagree on who's responsibility it is to write the automated tests for a change: You think that it's the project maintainer's responsibility, and I think it's the responsibility of the person doing the changes (i.e., usually the person creating the PR). You do this because it lowers the barrier for creating PRs, and I do this because I think the author of a change knows best how to test the changes and what the expected behavior should be. This is why we have different policies in our projects on this, and that's okay - not every project needs to have the same policies.

As you have seen, for a project where the maintainers usually don't take the time to write the tests for a PR from a contributor, providing tests with your PR greatly speeds things up.

I take your frustration about the tests as a nudge that we should make our tests easier to read and modify. I've created ticket #403 for this.

There are two things I'd like to point out, though:

1. You say that it takes you 10x the time to write the tests compared to than it would take me, and you also write that it would take me "a fraction of a second". I doubt both numbers (as you know better than I do what the new behavior after the change should be), and also the factor will be very different the second time you work with PHPUnit. You might also indeed be a unicorn, as usually we get PRs from people who regularly code PHP and who often already have experience with PHPUnit.

2. For me, this is not about whose time is (more) important, and I also don't see this as an issue of respect. For me, it's about who has the most interest of "scratching their own itch".

You also mentioned that there are easy ways to automate creating and maintaining a nice changelog. Which tools do you recommend for this?

PR #326, opened 2016-07-18
[BUGFIX] Ignore value with words more than one in the attribute selector

taiyonoryoshy:
Fixes #326

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -163,7 +163,7 @@ class Emogrifier
         // adjacent sibling
         '/\\s+\\+\\s+/'                            => '/following-sibling::*[1]/self::',
         // descendant
-        '/\\s+/'                                   => '//',
+        '/\\s+(?=.*[^\]]{1}$)/'                    => '//',
         // :first-child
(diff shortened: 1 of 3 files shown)


Classes/Emogrifier.php
```diff
@@ -163,7 +163,7 @@ class Emogrifier
         // adjacent sibling
         '/\\s+\\+\\s+/'                            => '/following-sibling::*[1]/self::',
         // descendant
-        '/\\s+/'                                   => '//',
+        '/\\s+(?=.*[^\]]{1}$)/'                    => '//',
```
oliverklee: Can you please make the backslash a double backslash to make it clear to the reader that this is no PHP escaping? (And, thinking of that,  I think we need four backslashes here: one set of escaping for PHP, and one set of escaping for regexp.)

oliverklee:
If I understand the fix correctly, the problem is that for attribute value selectors, we were only checking for words. With this fix, two words are also allowed (but not three or four). However, I think we need to think for "any string that does not contain double quotes or closing square braces" instead. Or am I misreading this completely?

What do you think?

oliverklee:
Marking this as a duplicate pf #327.

PR #328, opened 2016-07-29
Add Install to documentation

joubertredrat:
HI guy, I liked this lib a lot, but my developer had difficulties because he don't know composer a lot. Then I think that is good idea to add how to install by composer. Thanks


oliverklee:
@joubertredrat Great, thanks! Merged. Thanks for contributing! ❤️

PR #334, opened 2016-09-01
Save @import tag

asfwebmaster:
It removes @import tag from css.
Example:
>>>>-----------
```
<style type="text/css">
 @import url('https://fonts.googleapis.com/css?family=Open+Sans:400,400italic,700,700italic');
 @media{ body, table, td, p, a, li, blockquote {-webkit-text-size-adjust:none !important;}  }
</style>
```
------------->>>>
```
<style type="text/css">
 @media{ body, table, td, p, a, li, blockquote {-webkit-text-size-adjust:none !important;}  }
</style>
```
<<<<------------


Classes/Emogrifier.php
```diff
@@ -911,13 +911,14 @@ private function attributeValueIsImportant($attributeValue)
      *
      * @return void
      */
-    private function copyCssWithMediaToStyleNode(\DOMDocument $xmlDocument, \DOMXPath $xPath, $css)
+    private function copyCssWithMediaToStyleNode(\DOMDocument $xmlDocument, \DOMXPath $xPath, $css, $imports = '')
```
oliverklee: This pull request does not change the calls of this method, i.e, $imports will always be an empty string. So there will be no change in behavior. Or am I missing something?

oliverklee:
Hi @asfwebmaster, thanks for the PR!

It still needs some more love, though. Could you please reformat the code to be PSR-2 compliant? And could you please add unit tests that fail without your changes and pass with them? Thanks!

Classes/Emogrifier.php
oliverklee: And why does the new parameter have a default value? Do you expects (at least) two different calls (one with the argument and one without it)?

Classes/Emogrifier.php
```diff
@@ -1077,6 +1078,14 @@ function ($matches) use (&$media) {
             $cssWithoutComments
         );
 
+		//Keep @imports
+		$imports = '';
+		preg_match_all('/^\\s*@import\\s[^;]+;/misU', $css, $importsMatches, PREG_PATTERN_ORDER);
+		if( !empty($importsMatches[0]) ){
+			$importsArray = reset($importsMatches);
```
oliverklee: You don't need a reset before doing an implode.

Classes/Emogrifier.php
asfwebmaster: Hi, I did removed the default value for imports and added another missing line so the imports are passed now to that function.

Classes/Emogrifier.php
asfwebmaster: reset was giving me the first element of the array, same as $importsMatches[0] so I did remove the reset and replaced with $importsMatches[0]. Can you give me some more info what  is not PSR-2. Thanks in advanced man.

Classes/Emogrifier.php
```diff
@@ -911,13 +911,14 @@ private function attributeValueIsImportant($attributeValue)
      *
      * @return void
      */
-    private function copyCssWithMediaToStyleNode(\DOMDocument $xmlDocument, \DOMXPath $xPath, $css)
+    private function copyCssWithMediaToStyleNode(\DOMDocument $xmlDocument, \DOMXPath $xPath, $css, $imports)
```
oliverklee: Please add the missing param annotation.

Classes/Emogrifier.php
```diff
@@ -911,13 +911,14 @@ private function attributeValueIsImportant($attributeValue)
...
     {
-        if ($css === '') {
+        if ($css === '' && $imports === '') {
             return;
         }
 
         $mediaQueriesRelevantForDocument = [];
+        !$imports ?: $mediaQueriesRelevantForDocument[] = "\n".$imports."\n";
```
oliverklee: Please don't use the Elvis operator for commands. (It should only be used for return values or assignments). Please use an IF instead.

oliverklee: Please put spaces around the concatenation dots.

Classes/Emogrifier.php
```diff
@@ -1077,6 +1078,14 @@ function ($matches) use (&$media) {
             $cssWithoutComments
         );
 
+	//Keep @imports
```
oliverklee: This comment is a smell that the additional code should go into a separate method.

Classes/Emogrifier.php
```diff
@@ -1077,6 +1078,14 @@ function ($matches) use (&$media) {
...
+	$imports = '';
+	preg_match_all('/^\\s*@import\\s[^;]+;/misU', $css, $importsMatches, PREG_PATTERN_ORDER);
+	if( !empty($importsMatches[0]) ){
```
oliverklee: No spaces after the opening parenthesis or before the closing one, please. But there should be a space between the closing parenthesis and the opening curly brace.

oliverklee:
Now that I think of it, I'm not sure that loading externals font files from an within e-mail is a good idea in the first place. On desktop email clients like Thunderbird, the loading of external resources is disabled by default (i.e., the font will not be loaded), and on smartphones, you wouldn't want to eat up the bandwidth for relatively cosmetic things like web fonts in emails. I opt for continuing to remove the imports. @jjriv What do you think?

andrispraulitis:
@oliverklee Thunderbird is just one of the email clients that do not support <link> tag for custom fonts but does support @import. I don't see why this should not be allowed. It should really be my choice if I want to include custom fonts and eat bandwidth.

jjriv:
Anyone sending emails for marketing is likely going to want to use web fonts at some point. Campaign Monitor supports this functionality. I would argue that we should support this if email sending companies also support it, since Emogrifier is an alternative to CM and should remain competitive. And Emogrifier could be used to build such an email sending service. 

Here is the article from Campaign Monitor:
https://www.campaignmonitor.com/resources/guides/web-fonts-in-email/

It is also embraced by Litmus:
https://litmus.com/blog/typography-tips-for-email-designers

And MailChimp:
http://kb.mailchimp.com/campaigns/design/limitations-of-html-email

oliverklee:
@jjriv Thanks for the info! Then let's have this.

@asfwebmaster Can you please update your PR to fix the PSR-2 issues and also add tests that fail without your change and pass with it (and also adapt the existing tests as needed)?

Also, can you please add a changelog entry (newest on top) and squash your commits (as explained in the CONTRIBUTING file)? Thanks!

Classes/Emogrifier.php
```diff
@@ -350,7 +350,8 @@ protected function process(\DOMDocument $xmlDocument)
         if ($this->isStyleBlocksParsingEnabled) {
             $allCss .= $this->getCssFromAllStyleNodes($xPath);
         }
-
+ 		
+		$cssImports = $this->extractImportsFromCss($allCss);
```
oliverklee: Can you please read over all changes and fix the indentation? Thanks!

Classes/Emogrifier.php
```diff
@@ -394,6 +395,7 @@ protected function process(\DOMDocument $xmlDocument)
             $this->removeInvisibleNodes($xPath);
         }
 
+	$this->copyImportsToStyleNode($xmlDocument, $xPath, $cssImports);
```
oliverklee: Please use 4 spaces for indentation, not tabs or 8 spaces. You can find the PSR-2 style guide here:
https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md

I highly recommend that you read it. It's not very long, and it's used by a lot of projects.

Classes/Emogrifier.php
asfwebmaster: believe me I m trying to, the git editor not saving it correct

oliverklee:
@asfwebmaster I recommend you read up on "git commit --amend" and on squashing commits. Otherwise (if you wait and add more commits to this PR), squashing commits gets harder the more commits you have.

asfwebmaster:
Hi Man, Can you please help me with this issue. I m not very good with git. I did all this changes on github.com the problem is that when I change the tabs to 4 in their editor nothing is changed, or is changed but not as should be. If you can edit it somehow would be great if not can you please give me a suggestion how to do it? Thank you in advance.

oliverklee:
I'm glad to help.

First of all, are you using an IDE (like PhpStorm) for doing the Git stuff, or are you doing it on the command line? I highly recommend doing this on the command line as you then see what you are doing (whereas an IDE will hide the stuff "under the hood" and might also do things differently).

On the command line, you can do ```git status``` to see which files are changed. If you change the indentation, the changed file should be listed. Is it listed?

And if you are really new to Git on the command line, I highly recommend the free course "Try Git" on CodeSchool: https://www.codeschool.com/courses/try-git

oliverklee:
Oh, I just read that you did all changes in the web interface of GitHub. This does not work well at all. You absolutely need to check out your fork and work on a local copy to get a clean PR. Please have a look at the CONTRIBUTING document - I've linked some helpful guides from it.

oliverklee:
This guide explains the thing with forks and PRs: https://gist.github.com/Chaser324/ce0505fbed06b947d962

oliverklee:
Oh, and you absolutely need to have a local copy to work on so you can run the tests locally.

PR #359, opened 2017-01-07
Add style blocks to body rather than head

Gabe-SpotMe:
Foundation for Email recommends inlining css into the body rather than head so I added an option to allow for that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## x.y.z (unreleased)
 
 ### Added
-
+- Added setting for putting the media query style block to <body> vs <head>
 
 ### Changed
 
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -155,6 +155,17 @@ class Emogrifier
     private $shouldKeepInvisibleNodes = true;
 
     /**
+     * Determines whether media query style blocks should be added to the body versus the head of the HTML.
+     * 
+     * If set to true, the <style> blocks will be added to the <body>
+     * 
+     * If set to false, the <style> blocks will be added to the <head>
+     *
+     * @var bool
+     */
+    private $shouldAddStyleToBody = false;
+
+    /**
      * @var string[]
      */
     private $xPathRules = [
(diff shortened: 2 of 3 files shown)


Classes/Emogrifier.php
```diff
@@ -1046,6 +1079,20 @@ private function getOrCreateHeadElement(\DOMDocument $document)
...
     /**
+     * Returns the existing body element in $document.
+     *
+     * @param \DOMDocument $document
+     *
+     * @return \DOMNode the body element
+     */
+    private function getBodyElement(\DOMDocument $document)
+    {
+        $body = $document->getElementsByTagName('body')->item(0);
```
oliverklee: You can also inline $body.

Classes/Emogrifier.php
```diff
@@ -155,6 +155,17 @@ class Emogrifier
...
     /**
+     * Determines whether media query style blocks should be added to the body versus the head of the HTML.
+     * 
+     * If set to true, the <style> blocks will be added to the <body>
+     * 
+     * If set to false, the <style> blocks will be added to the <head>
+     *
+     * @var bool
+     */
+    private $shouldAddStyleToBody = false;
```
oliverklee: I think we can make this new behavior the default instead of making it configurable. @Gabe-SpotMe, @jjriv What do you think?

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## x.y.z (unreleased)
 
 ### Added
-
+- Added setting for putting the media query style block to <body> vs <head>
```
oliverklee: Could you please add the issue/PR number so this line is the same format as the other changelog entries?

oliverklee:
@Gabe-SpotMe Thanks a bunch for the PR! However, the new behavior must be covered with unit tests that fail without the changes and that pass with them.

I'd like to help you add these unit tests. For starters, have you been successful running the existing unit tests locally on your machine? (Please have a look in the CONTRIBUTING.md file for details.)

oliverklee:
@jjriv I propose we just make this the new behavior without an option to switch back to the old behavior. (I currently don't see a use case where someone would need to have the old behavior back.) What do you think?

jjriv:
I agree with @oliverklee . The less configuration options the better. This is a clearly established default. Let's make this the new default behavior.

Classes/Emogrifier.php
Gabe-SpotMe: I'm not sure what you mean...can you clarify?

oliverklee: With "inlining" I mean "replace $body (where it's used) directly with its definition. :-)

oliverklee:
Closing this in favor of #396.

Issue #360, opened 2017-01-16
How to get more information about wrong xPath selector


OzzyCzech:
We are using emogrifier a lot but sometimes is hard to get right information what's wrong. Mostly we get just `PHP Warning: DOMXPath::query(): Invalid expression in /vendor/pelago/emogrifier/Classes/Emogrifier.php:366`

there is my proposal how to get more information about that type of warnings:

```
class Emogrifier extends \Pelago\Emogrifier {

  protected function process(\DOMDocument $xmlDocument) {

    set_error_handler(
      function ($errno, $errstr, $errfile, $errline, $errcontext) {
        if (isset($errcontext['cssRule']['selector'])) {
          throw new \Exception(
              $errstr . ' in selector >> ' . $errcontext['cssRule']['selector'] . ' <<'
          );
        }
      }, E_WARNING
    );

    parent::process($xmlDocument);

    restore_error_handler();
  }

}
```

It's throw an error like : `DOMXPath::query(): Invalid expression in selector >> html input[disabled] << ` now you can easily found what's wrong.

oliverklee:
Hi @OzzyCzech, great proposal! Would you be willing to create a pull request for this (including unit tests)?

OzzyCzech:
sure

OzzyCzech:
done, check latest version

PR #361, opened 2017-01-17
Getting more information about failing xPath selector

OzzyCzech:
BC break:
> Newly throws InvalidArgumentException during processing invalid
> xPath selectors instead of silent PHP warning.

It’s related to #360 issue


Classes/Emogrifier.php
```diff
@@ -1501,4 +1505,34 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
+     * @param string $errStr
+     * @param string $errFile
+     * @param int $errLine
+     * @param array $errContext
+     *
+     * @return bool
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function errorHandler($errNo, $errStr, $errFile, $errLine, $errContext)
```
oliverklee: Please let's make this more specific, e.g., handleXpathError.

Classes/Emogrifier.php
```diff
@@ -1501,4 +1505,34 @@ private function getNodesToExclude(\DOMXPath $xPath)
 
         return $excludedNodes;
     }
+
+    /**
+     * Handle warnings in process() method
```
oliverklee: The first line of a method comment should always be in the third person: "[This method] Handles …"

Classes/Emogrifier.php
```diff
@@ -1501,4 +1505,34 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
+     *
+     * @param int $errNo
+     * @param string $errStr
+     * @param string $errFile
+     * @param int $errLine
+     * @param array $errContext
```
oliverklee: Do we know more about the array, e.g., string[][] or int[][]?

Classes/Emogrifier.php
oliverklee: Please add an "array" type hint to the last parameter.

OzzyCzech:
Looks that HHVM do not fire Exception at all :(

Classes/Emogrifier.php
```diff
@@ -1501,4 +1505,34 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
+     *
+     * @return bool
```
oliverklee: We don't need a return value as this method throws an exception if there is a problem and does not throw it if everything is fine. So this can be return void, and the return false can go away.

Classes/Emogrifier.php
oliverklee: No abrreviations ("err"), please. Actually, I think that we don't need the prefix anyway as the context already is given by the method.

OzzyCzech:
https://github.com/facebook/hhvm/issues/5790

Classes/Emogrifier.php
oliverklee: Please also add some description on what the method actually does.

Classes/Emogrifier.php
```diff
@@ -1501,4 +1505,34 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function errorHandler($errNo, $errStr, $errFile, $errLine, $errContext)
+    {
+        if ($errNo === E_WARNING && isset($errContext['cssRule']['selector'])) {
```
oliverklee: Do we need this check, or will the handler also be called in other cases?

oliverklee:
On the HHVM problem: Can we work around this somehow? Or document that this feature only is available on PHP, not on HHVM, and skip the test in that case?

oliverklee:
Could you also please add a line to the change log file (newest on top) and add a "Fixes #360" to the body of the commit message so the other issue will be auto-closed on merging? Thanks!

Classes/Emogrifier.php
OzzyCzech: context it's just an array of object contains context variables from current scope

Classes/Emogrifier.php
OzzyCzech: Return false have specific reason, we don't want to catch all errors and throw them away
If the function returns FALSE then the normal error handler continues.

oliverklee: Ah, okay, I didn't know that. Thanks for the explanation. Could you then please add something like "bool always false" plus an explanation to the return annotation? Thanks!

Classes/Emogrifier.php
OzzyCzech: not really :) but
1. code sniffer trigger error about not used variable
2. errorHandler will be called even for E_ERRORS

Classes/Emogrifier.php
oliverklee: Do we know anything about the types of the contained context variables? Or will they be mixed int, string, bool etc?

Classes/Emogrifier.php
OzzyCzech: Sure

Classes/Emogrifier.php
OzzyCzech: No, they will be arrays, objects and strings ... mixed

OzzyCzech:
Only way is skip this test in HHVM

Classes/Emogrifier.php
oliverklee: Okay, then "array" will suffice. Thanks for the info!

oliverklee:
Okay, then we need to skip it depending on the runtime. Here is an example: https://github.com/sebastianbergmann/phpunit/issues/1356

OzzyCzech:
done, check latest version

oliverklee:
@OzzyCzech Thanks! I'll have a look at it!

Classes/Emogrifier.php
```diff
@@ -308,7 +310,9 @@ public function emogrifyBodyContent()
         }
 
         $xmlDocument = $this->createXmlDocument();
+        set_error_handler([$this, 'handleXpathError'], E_WARNING);
         $this->process($xmlDocument);
+        restore_error_handler();
```
oliverklee: We have exactly the same code here and in emogrify (starting from the check for empty $this->html). Please let's move this to a separate private method to avoid the duplication.

Classes/Emogrifier.php
```diff
@@ -1501,4 +1505,37 @@ private function getNodesToExclude(\DOMXPath $xPath)
 
         return $excludedNodes;
     }
+
+    /**
+     * Handles invalid xPath expression warnings, generated by process() method,
+     * during querying \DOMDocumentand and trigger \InvalidArgumentException
```
oliverklee: There's a space missing before the "and".

Classes/Emogrifier.php
```diff
@@ -1501,4 +1505,37 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
+     * with invalid selector
```
oliverklee: Please add the missing period at the end.

Tests/Unit/EmogrifierTest.php
```diff
@@ -967,27 +967,22 @@ public function emogrifyRemovesStyleNodes()
...
+     *
+     * @expectedException \InvalidArgumentException
      */
-    public function emogrifyIgnoresInvalidCssSelector()
+    public function emogrifyFailsOnInvalidCssSelector()
     {
+        // HHVM ignore custom error handler settings for libxml
+        // @see https://github.com/facebook/hhvm/issues/5790
+        if (defined('HHVM_VERSION')) {
+            $this->markTestIncomplete('HHVM ignore custom error handler');
```
oliverklee: This should be markTestSkipped, not markTestIncomplete. (markTestIncomplete serves as a kind of "to do" marker.)

Classes/Emogrifier.php
OzzyCzech: Maybe we can move handler code to `process()` or add `processWithErrorHandling()` method

Classes/Emogrifier.php
```diff
@@ -1501,4 +1518,37 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
+     * @param string $message
+     * @param string $file
+     * @param int $line
+     * @param array $context
+     *
+     * @return bool always false
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function handleXpathError($type, $message, $file, $line, array $context)
```
oliverklee: Does this method need to be public, or could it be private?

Classes/Emogrifier.php
```diff
@@ -404,6 +404,23 @@ protected function process(\DOMDocument $xmlDocument)
...
+    /**
+     * Applies $this->css to $xmlDocument and handle process errors.
+     *
+     * This method places the CSS inline.
+     *
+     * @param \DOMDocument $xmlDocument
+     *
+     * @return void
+     */
+    private function processWithErrorHandling(\DOMDocument $xmlDocument)
```
oliverklee: Thanks for moving this to a separate method.

process now only is called from this method. So I think the error handling could go directly into process. What do you think?

Classes/Emogrifier.php
```diff
@@ -1501,4 +1518,37 @@ private function getNodesToExclude(\DOMXPath $xPath)
 
         return $excludedNodes;
     }
+
+    /**
+     * Handles invalid xPath expression warnings, generated by process() method,
+     * during querying \DOMDocument and trigger \InvalidArgumentException
+     * with invalid selector
```
oliverklee: Please add the missing period at the end of the sentence.

Classes/Emogrifier.php
OzzyCzech: Sure, it's can be part of process()

OzzyCzech: Sure, that can be part of process method

Classes/Emogrifier.php
OzzyCzech: It's have to be public, it's callback. But there is one option. Callback can be hidden to directly to the process like `set_error_handler(function() { }, ...);` but that's avoid options to overwrite handler behavior.

OzzyCzech: For me is clear api better then have this function in public space

oliverklee: I agree. Then let's keep it this way. Thanks for the explanation!

OzzyCzech: Shall I leave it as is or hide callback to process?

oliverklee: Please leave it as a public method. Thanks!

OzzyCzech: :) ok, check latest version

CHANGELOG.md
```diff
@@ -25,7 +25,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Security
 
+## 1.2.0
```
oliverklee: This line should not be here. (We'll add it later for the release.)

CHANGELOG.md
```diff
@@ -25,7 +25,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
 
+### Added
```
oliverklee: There's also already an "Added" section (line 9) where you can add the feature.

Classes/Emogrifier.php
```diff
@@ -329,6 +329,7 @@ public function emogrifyBodyContent()
      */
     protected function process(\DOMDocument $xmlDocument)
```
oliverklee: This method now needs the "throws" annotation (and an explanation on the same line explaining that the Exception is thrown for invalid XPath).

Classes/Emogrifier.php
```diff
@@ -402,6 +403,7 @@ protected function process(\DOMDocument $xmlDocument)
         }
 
         $this->copyCssWithMediaToStyleNode($xmlDocument, $xPath, $cssParts['media']);
+        restore_error_handler();
```
oliverklee: Do the two lines need enclose this part of the code, or can we narrow it down so it encloses only the lines that actually do some XPath processing?

CHANGELOG.md
OzzyCzech: I am blind 🤦‍♂️

oliverklee: ;-)

Classes/Emogrifier.php
OzzyCzech: I am just move both lines closer to the queries, that's better

OzzyCzech:
You are welcome!

PR #382, opened 2017-05-15
Added exceptions when finding a unsupported css tag.

winus:
I wanted to merge bootstrap.css file into my html. But is failed on a @-ms-something hack, which it did not recognize. With this simple fix is does proceed and works just fine. But you have to get them errors by manual though...


winus:
Never mind.. It didn't check out the latest version.

PR #392, opened 2017-06-22
Fix InvalidArgumentException

jriboux:
Fix InvalidArgumentException when xpath does not find the css selector (eg: css3 keyframes).

You can reproduce the InvalidArgumentException using the following css :
```css
@-moz-keyframes animate-stripes{
    0% {background-position: 0 0;} 100% {background-position: 60px 0;}
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - Debug mode. Throw debug exceptions only if debug is active.
+  ([#392](https://github.com/jjriv/emogrifier/pull/392))
+
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## x.y.z (unreleased)
 
 ### Added
-
+- Debug mode. Throw debug exceptions only if debug is active.
 
 ### Changed
 
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -1533,22 +1533,18 @@ class Emogrifier
      *
      * @throws \InvalidArgumentException
      */
-    public function handleXpathError($type, $message, $file, $line, array $context)
+     public function handleXpathError($type, $message, $file, $line, array $context)
     {
-        if ($type === E_WARNING && isset($context['cssRule']['selector'])) {
-            if ($this->debug) {
-                throw new \InvalidArgumentException(
-                    sprintf(
-                        '%s in selector >> %s << in %s on line %s',
-                        $message,
-                        $context['cssRule']['selector'],
-                        $file,
-                        $line
-                    )
-                );
-            } else {
-                return false;
-            }
+        if ($this->debug && $type === E_WARNING && isset($context['cssRule']['selector'])) {
+            throw new \InvalidArgumentException(
+                sprintf(
+                    '%s in selector >> %s << in %s on line %s',
+                    $message,
+                    $context['cssRule']['selector'],
+                    $file,
+                    $line
+                )
+            );
         }
 
         // the normal error handling continues when handler return false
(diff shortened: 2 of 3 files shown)


oliverklee:
HI @jriboux, thanks a lot for the PR!

Could you please add a unit test that fails without the change and that passes with it? And could you also please add a line to the changelog file?

Classes/Emogrifier.php
```diff
@@ -367,8 +367,11 @@ protected function process(\DOMDocument $xmlDocument)
         $cssRules = $this->parseCssRules($cssParts['css']);
         foreach ($cssRules as $cssRule) {
             // query the body for the xpath selector
-            $nodesMatchingCssSelectors = $xPath->query($this->translateCssToXpath($cssRule['selector']));
-            // ignore invalid selectors
+            try {
+                $nodesMatchingCssSelectors = $xPath->query($this->translateCssToXpath($cssRule['selector']));
+            } catch (\InvalidArgumentException $e) {
+                $nodesMatchingCssSelectors = false;
+            }            // ignore invalid selectors
```
oliverklee: Please move this comment to above the catch (in its own line).

jriboux:
Sorry, my changes are in conflict with pull request #361 .
I need emogrify to ignore bad css selectors but developers should be able to debug.

Could we implement a debug switch, false by default that would : 
 - Enable throws of InvalidArgumentException if true
 - Silence xpath errors if false

oliverklee:
Yes, a debug switch would be good. It should be off by default, and we should have a method public setDebug(bool $debug): void.

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## x.y.z (unreleased)
 
 ### Added
-
+- Debug mode. Throw debug exceptions only if debug is active.
```
oliverklee: Could you also please link the PR (i.e., use the same format as for the other changelog entries)?

Classes/Emogrifier.php
```diff
@@ -226,6 +226,13 @@ class Emogrifier
     ];
 
     /**
+     * If true, allow debug Exception to be thrown.
```
oliverklee: Let's make this more specific: … Emogrifier will throw Exceptions when it encounters an error instead of silently ignoring them.

Classes/Emogrifier.php
```diff
@@ -1529,18 +1536,32 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
-            throw new \InvalidArgumentException(
-                sprintf(
-                    '%s in selector >> %s << in %s on line %s',
-                    $message,
-                    $context['cssRule']['selector'],
-                    $file,
-                    $line
-                )
-            );
+            if ($this->debug) {
```
oliverklee: We don't need the else here as throwing an exception will exit the method anyway.

Classes/Emogrifier.php
```diff
@@ -1529,18 +1536,32 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
+                return false;
+            }
         }
 
         // the normal error handling continues when handler return false
         return false;
     }
+
+    /**
+     * Set debug mode.
```
oliverklee: Please always use the third person for the first sentence of a method description: [This Method] "Sets the debug mode."

Classes/Emogrifier.php
```diff
@@ -1529,18 +1536,32 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
+     *
+     * @param bool $debug set to true to enable debug mode
+     */
```
oliverklee: Please also add the return void annotation.

Tests/Unit/EmogrifierTest.php
```diff
@@ -956,10 +956,27 @@ public function emogrifyForInvalidCssSelectorThrowsException()
...
         );
 
+        $this->subject->setDebug(true);
         $this->subject->emogrify();
     }
 
     /**
+     * @test
+     */
+    public function emogrifyIgnoreInvalidCssSelectors()
```
oliverklee: Please also add another test that tests for the behavior without debug mode.

jriboux: I kept the test from pull request #361 "Handling invalid xPath expression warnings" and added the following line to test debug mode :
`$this->subject->setDebug(true);`

See lines 948 to 961

>

oliverklee:
Merged. Thanks for contributing to Emogrifier, @jriboux! ❤️

Issue #356, opened 2016-12-16
CSS Selector :not() is not parsed


wimulkeman:
The following CSS is not parsed by Emogrifier.

`td.columns table:not(.button),
td.column table:not(.button),
th.columns table:not(.button),
th.column table:not(.button) {
  width: 100%; }`

The part which seems to produce a problem is :not(.button). There is no error produced from within Emogrifier.

In the old versions, the CSS would still be available in the stylesheet in the header of the mail, but not converted into inline CSS. In the newer versions of Emogrifier (1.1.0), this part of the CSS is also removed from the stylesheet in the header by Emogrifier, which produces a more default behavior, but disabled the usage of the CSS functionality completely.

The Symfony CSS Selector, suggested in https://github.com/jjriv/emogrifier/issues/287 seems to fix this problem with this https://github.com/symfony/css-selector/blob/master/Parser/Parser.php#L238.

oliverklee:
Fixed by #431.

Issue #390, opened 2017-06-09
Support :not(.class)


joeworkman:
When using a CSS selector like `table:not(.button)`, it cause the following fatal error: 

```
Uncaught InvalidArgumentException: DOMXPath::query(): Invalid expression in selector
```

oliverklee:
Fixed by #431.

PR #395, opened 2017-07-28
Add support for :not selector

Renkas:
Fixes issue: #356, #390


Renkas:
Tests failing are not because of this PR. They were failing before.

Same problem as in issue: https://github.com/jjriv/emogrifier/issues/350

oliverklee:
Ah, okay. I'll take care of fixing the tests and the code sniffer warnings first. I'll the post here once that is done so you can rebase.

Thanks for contributing! :heart:

oliverklee:
Hi @Renkas, I've just checked. The build on the Emogrifier master is green:
https://travis-ci.org/jjriv/emogrifier/builds

The build failure for your PR is due to some code sniffer warnings (lines that are too long).

In addition, there are some PHP warnings when running the tests (also on master), but those don't break the build. I'll take care of those, but this shouldn't block your PR.

Could you please fix the style warnings and repush?

You can also run PHP_CodeSniffer locally before pushing:
vendor/bin/phpcs --standard=Configuration/PhpCodeSniffer/Standards/Emogrifier/ Classes/ Tests/

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Debug mode. Throw debug exceptions only if debug is active.
   ([#392](https://github.com/jjriv/emogrifier/pull/392))
-
+- Support for CSS :not() selector.
+  ([#395](https://github.com/jjriv/emogrifier/pull/395))
```
oliverklee: The entries should be ordered in reverse chronological order, i.e., newest on top. Could you please move this entry to the top of the "Added" section?

Tests/Unit/EmogrifierTest.php
```diff
@@ -670,6 +670,8 @@ public function selectorDataProvider()
                 => ['p:last-child {' . $styleRule . '} ', '#<p class="p-6" style="' . $styleRule . '">#'],
             'DIV:last-child does not match last child with mismatching tag'
                 => ['div:last-child {' . $styleRule . '} ', '#<p class="p-6">#'],
+            'P:not(.p-6) matches all but P.p-6'
+                => ['p:not(.p-6) {' . $styleRule . '}', '#<p class="p-6">#'],
```
oliverklee: These should be two separate tests (i.e., cases in the data provider): One that it matches the other elements, and one that it does not match the corresponding element.

Classes/Emogrifier.php
```diff
@@ -1289,40 +1292,63 @@ function (array $matches) {
...
+                $notSelector = trim($notSelector, '()');
+                $finalXpath = $finalXpath . '[not(' . $this->translateCssToXpathPass($notSelector, true) . ')]';
+            }
 
             $this->caches[self::CACHE_KEY_SELECTOR][$xPathKey] = $finalXpath;
         }
         return $this->caches[self::CACHE_KEY_SELECTOR][$xPathKey];
     }
 
+    private function translateCssToXpathPass($trimmedLowercaseSelector, $inline = false)
```
oliverklee: This method needs some PHPDoc. Also, I'd prefer there to be two methods instead of a switch that changes the behavior.

Classes/Emogrifier.php
```diff
@@ -1289,40 +1292,63 @@ function (array $matches) {
...
+        $xPathWithIdAttributeAndClassMatchers = preg_replace_callback(
+            '/([^\\/]+):nth-child\\(\\s*(odd|even|[+\\-]?\\d|[+\\-]?\\d?n(\\s*[+\\-]\\s*\\d)?)\\s*\\)/i',
+            [$this, 'translateNthChild'],
+            $xPathWithIdAttributeAndClassMatchers
+        );
+        $finalXpath = preg_replace_callback(
+            '/([^\\/]+):nth-of-type\\(\s*(odd|even|[+\\-]?\\d|[+\\-]?\\d?n(\\s*[+\\-]\\s*\\d)?)\\s*\\)/i',
+            [$this, 'translateNthOfType'],
+            $xPathWithIdAttributeAndClassMatchers
+        );
```
oliverklee: If I read this correctly, you're also just moving code around. To have things separate and to make this easier to review, could you please split this into two PRs: One refactoring PR that just moves code around without changing the behavior, and one PR after that adds the new behavior, making use of the then-already-added refactorings.

oliverklee:
@Renkas Any news on this PR? (Please don't feel discouraged by the review comments. I aim to get your changes into the code base in their best form.)

Classes/Emogrifier.php
Renkas: Not just moving code around. The same exact code is supposed to be run 2 times.

First for regular xPath query and second time to create :not query. There is no point in writing same code twice. Only thing changed in the code is I added variable $matchClassAttributesCallback that changes if the result should be full xPath selector or only the inline part needed for :not query.

Classes/Emogrifier.php
```diff
@@ -1289,41 +1292,71 @@ function (array $matches) {
...
-                [$this, 'matchIdAttributes'],
-                $roughXpath
-            );
-            $xPathWithIdAttributeAndClassMatchers = preg_replace_callback(
-                self::CLASS_ATTRIBUTE_MATCHER,
-                [$this, 'matchClassAttributes'],
-                $xPathWithIdAttributeMatchers
-            );
+            $notSelector = null;
+            if (strstr($trimmedLowercaseSelector, ':not')) {
```
oliverklee: According to http://php.net/strstr , using strpos !== false is faster for finding out whether a string has a certain substring.

Classes/Emogrifier.php
```diff
@@ -1289,41 +1292,71 @@ function (array $matches) {
...
-                $xPathWithIdAttributeAndClassMatchers
-            );
-            $finalXpath = preg_replace_callback(
-                '/([^\\/]+):nth-of-type\\(\s*(odd|even|[+\\-]?\\d|[+\\-]?\\d?n(\\s*[+\\-]\\s*\\d)?)\\s*\\)/i',
-                [$this, 'translateNthOfType'],
-                $xPathWithIdAttributeAndClassMatchers
-            );
+            $finalXpath = '//' . $this->translateCssToXpathPass($trimmedLowercaseSelector);
+
+            if ($notSelector) {
```
oliverklee: $notSelector currently is either a string or null, not a boolean. I propose to always make it a string and then check for !== '' here.

Classes/Emogrifier.php
```diff
@@ -1289,41 +1292,71 @@ function (array $matches) {
...
+                $notSelector = trim($notSelector, '()');
```
oliverklee: Please let's use two different local variables for this.

Classes/Emogrifier.php
```diff
@@ -1289,41 +1292,71 @@ function (array $matches) {
...
-                self::ID_ATTRIBUTE_MATCHER,
-                [$this, 'matchIdAttributes'],
-                $roughXpath
-            );
-            $xPathWithIdAttributeAndClassMatchers = preg_replace_callback(
-                self::CLASS_ATTRIBUTE_MATCHER,
-                [$this, 'matchClassAttributes'],
-                $xPathWithIdAttributeMatchers
-            );
+            $notSelector = null;
```
oliverklee: Can we put this whole section in a well-name private method, e.g. "extractNotSelector" (or something similar)?

Classes/Emogrifier.php
```diff
@@ -1289,41 +1292,71 @@ function (array $matches) {
...
 
             $this->caches[self::CACHE_KEY_SELECTOR][$xPathKey] = $finalXpath;
         }
         return $this->caches[self::CACHE_KEY_SELECTOR][$xPathKey];
     }
 
     /**
+     * @param string $trimmedLowercaseSelector
+     *
+     * @param bool $inline If set to true it only outputs inline part of xPath class selector. Needed for :not selector
```
oliverklee: No behavior toggles as parameter, please. Instead, please let's have two separate methods that each do exactly one thing (while one method may call the other).

Classes/Emogrifier.php
oliverklee: Ah, okay. Thanks for the clarification!

Classes/Emogrifier.php
```diff
@@ -1340,11 +1373,21 @@ private function matchIdAttributes(array $match)
...
-        return ($match[1] !== '' ? $match[1] : '*') . '[contains(concat(" ",@class," "),concat(" ","' .
+        return ($match[1] !== '' ? $match[1] : '*') . '[' . $this->matchClassAttributesInline($match) . ']';
+    }
+
+    /**
+     * @param string[] $match
+     *
+     * @return string
+     */
+    private function matchClassAttributesInline(array $match)
```
oliverklee: Please also add a documentation comment that explains what this method does.

Classes/Emogrifier.php
Renkas: You are welcome to change that. I'm not going to duplicate code - that's much worse solution.

oliverklee: I think you misunderstood me. I'll create a PR for this refactoring.

oliverklee:
We've moved the repository to https://github.com/MyIntervals/emogrifier . If you have an "upstream" branch in your fork, please make sure to update the URL:

`git remote set-url upstream git@github.com:MyIntervals/emogrifier.git`

Renkas:
What is the status?

oliverklee:
I seems that you have re-pushed. Unfortunately. GitHub does not send any notifications of this. :-(

Thanks for the reminder! :cookie: I'll have another look at the PR.

oliverklee:
I've moved the refactoring part to #413 and will also do some refactoring after that, making this PR here a lot smaller. Hold on.

oliverklee:
@Renkas Could you please rebase, repush and then post here? Then I'll do the next (and final) part of the preparatory refactoring. Thanks!

Renkas:
Should be all ok now

oliverklee:
@Renkas Thanks!

Classes/Emogrifier.php
```diff
@@ -1344,33 +1344,65 @@ function (array $matches) {
...
+
+    /**
      * @param string $trimmedLowercaseSelector
      *
+     * @param bool $inline If set to true it only outputs inline part of xPath class selector. Needed for :not selector
+     *
      * @return string
      */
-    private function translateCssToXpathPass($trimmedLowercaseSelector)
+    private function translateCssToXpathPass($trimmedLowercaseSelector, $inline = false)
```
oliverklee: I've just created the PR #419 to make this more elegant. Comments welcome!

README.md
```diff
@@ -166,6 +166,7 @@ Emogrifier currently support the following
  * attribute only
  * first-child
  * last-child
+ * not
```
oliverklee: In the latest README, I've linked the corresponding MDN entry and also sorted the supported selectors the same way they are sorted in MDN. When you rebase, could you also do this for the `not` selector (which is a pseudo-class)? Thanks!

Tests/Unit/EmogrifierTest.php
```diff
@@ -587,6 +587,10 @@ public function selectorDataProvider()
             => ['p:last-child {' . $styleRule . '} ', '#<p class="p-6" style="' . $styleRule . '">#'],
             'DIV:last-child does not match last child with mismatching tag'
             => ['div:last-child {' . $styleRule . '} ', '#<p class="p-6">#'],
+            'P:not(.p-6) matches all but P.p-6'
+            => ['p:not(.p-6) {' . $styleRule . '}', '#<p class="p-6">#'],
+            'P:not(.p-6) matches all but P.p-6 check if other paragraphs got style applied'
+            => ['p:not(.p-6) {' . $styleRule . '}', '#<p class="p-5" ' . $styleAttribute . '>#'],
```
oliverklee: After #417 has been merged, you can also add this to matchedCssDataProvider and nonMatchedCssDataProvider, if you like. (You can also keep this as is, and then I'll convert this later.)

Classes/Emogrifier.php
```diff
@@ -1344,33 +1344,65 @@ function (array $matches) {
...
     {
-        $roughXpath = '//' .
```
oliverklee: Why are you removing the `//`?

Classes/Emogrifier.php
```diff
@@ -1344,33 +1344,65 @@ function (array $matches) {
...
+
         return $this->caches[self::CACHE_KEY_SELECTOR][$xPathKey];
     }
 
     /**
+     * @param string $trimmedLowercaseSelector Whole CSS selector. :not selector part will be removed
+     *
+     * @return string :not selector content or empty string
+     */
+    private function extractNotSelector(&$trimmedLowercaseSelector)
```
oliverklee: Please don't use a reference here. We don't want the caller to have their variable modified.

oliverklee:
@Renkas The PRs are merged. Could you please rebase? Thanks!

oliverklee:
I've just merged PR #420. @Renkas, could you please rebase and then post a short message? Then I'll re-review the PR. Thanks!

oliverklee:
@Renkas Any news? (I'd like to get this in the next release.)

Tests/Unit/EmogrifierTest.php
Renkas: I'll leave it to you then

Renkas:
Should be all good from my side

oliverklee:
Hi @Renkas, I've polished you PR a bit and merged it as #431. Thanks for contributing to Emogrifier! :heart:

PR #398, opened 2017-09-07
[FEATURE] PHP 7.2 support

finwe:


finwe:
@oliverklee Followed contributing guidelines, but I am unable to request a review.

There are a few warnings in 7.2 build - deprecation notice about `each()` which is caused by old version of PHPUnit and DOM warnings which are present in previous builds, so I let them be.

oliverklee:
Thanks for the PR, @finwe! :heart:  Looks generally good!

Could you also please remove the obsolete "Requirements" section from the README (as it's all covered by the composer.json now)?

composer.json
```diff
@@ -29,7 +29,7 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0, <=7.1.99"
+        "php": "^5.4 || ^7.0.0 || ^7.1.0 || ^7.2.0"
```
oliverklee: To have this consistent, this should be 5.4.0, not 5.4.

Also, for the 7 versions, we need to use ~, not ^, as the ^ would also allow 7.3.0 (if I understand the operator correctly): https://getcomposer.org/doc/articles/versions.md#tilde-version-range-

finwe: You are correct with the caret operator. Changed accordingly.

oliverklee:
Merged. Thanks for contributing to Emogrifier, @finwe! :cookie: :heart:

finwe:
And thanks to you for such a swift response!

PR #400, opened 2017-09-17
[BUGFIX] Silence purposefully ignored PHP warnings

markcarver:
NOTE: I know this doesn't have tests because this kind of logic isn't really "testable" in the first place other than what already exists.

Instead, I have verbosely commented why the error handler must always throw an exception and not be tied to the debug option.

This fixes #361 and #392 which introduced this new error handler and started spitting out PHP Warnings when it's ultimately just ignored anyway.


oliverklee:
Hi @markcarver, thanks for the PR! I'll have a look at it.

oliverklee:
It looks like we'll also need to migrate the Travis job so it can run for this PR. I'm on it.

Classes/Emogrifier.php
```diff
@@ -373,11 +373,21 @@ protected function process(\DOMDocument $xmlDocument)
...
-            $nodesMatchingCssSelectors = $xPath->query($this->translateCssToXpath($cssRule['selector']));
-            // ignore invalid selectors
-            if ($nodesMatchingCssSelectors === false) {
-                continue;
+            // There's no real way to test "PHP Warning" output generated by the following XPath query unless PHPUnit converts it to an exception.
+            // Unfortunately, this would only apply to tests and not work for production executions, which can still flood logs/output unnecessarily.
+            // Instead, Emogrifier's error handler should always throw an exception and it must be caught here and only rethrown if in debug mode.
+            try {
+              // query the body for the xpath selector
+              $nodesMatchingCssSelectors = $xPath->query($this->translateCssToXpath($cssRule['selector']));
```
oliverklee: I think we should have a comment here explaining that this will never be false as we already handle this with via the error handler.

Classes/Emogrifier.php
```diff
@@ -373,11 +373,21 @@ protected function process(\DOMDocument $xmlDocument)
...
+            }
+            catch (\InvalidArgumentException $e) {
+              // Rethrow exception if in debug mode.
```
oliverklee: This comment (and also the next one) are not necessary as the code already is "speaking" enough, and not having the (redundant) comment will allow us to not have to maintain it. :-)

Classes/Emogrifier.php
```diff
@@ -373,11 +373,21 @@ protected function process(\DOMDocument $xmlDocument)
...
-            // query the body for the xpath selector
-            $nodesMatchingCssSelectors = $xPath->query($this->translateCssToXpath($cssRule['selector']));
-            // ignore invalid selectors
-            if ($nodesMatchingCssSelectors === false) {
-                continue;
+            // There's no real way to test "PHP Warning" output generated by the following XPath query unless PHPUnit converts it to an exception.
+            // Unfortunately, this would only apply to tests and not work for production executions, which can still flood logs/output unnecessarily.
+            // Instead, Emogrifier's error handler should always throw an exception and it must be caught here and only rethrown if in debug mode.
+            try {
+              // query the body for the xpath selector
```
oliverklee: This comment also is not needed as the code already is "speaking" enough.

Classes/Emogrifier.php
```diff
@@ -373,11 +373,21 @@ protected function process(\DOMDocument $xmlDocument)
...
         $cssRules = $this->parseCssRules($cssParts['css']);
         foreach ($cssRules as $cssRule) {
-            // query the body for the xpath selector
-            $nodesMatchingCssSelectors = $xPath->query($this->translateCssToXpath($cssRule['selector']));
-            // ignore invalid selectors
-            if ($nodesMatchingCssSelectors === false) {
-                continue;
+            // There's no real way to test "PHP Warning" output generated by the following XPath query unless PHPUnit converts it to an exception.
+            // Unfortunately, this would only apply to tests and not work for production executions, which can still flood logs/output unnecessarily.
+            // Instead, Emogrifier's error handler should always throw an exception and it must be caught here and only rethrown if in debug mode.
```
oliverklee: Great comment!

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 
 ### Fixed
+- Silence purposefully ignored PHP Warnings
+  ([#400](https://github.com/MyIntervals/emogrifier/pull/400))
```
oliverklee: Thanks!

Classes/Emogrifier.php
```diff
@@ -1535,7 +1545,8 @@ private function getNodesToExclude(\DOMXPath $xPath)
      */
     public function handleXpathError($type, $message, $file, $line, array $context)
     {
-        if ($this->debug && $type === E_WARNING && isset($context['cssRule']['selector'])) {
+        // Do not tie this conditional to debug mode as it causes unnecessary output. See further explanation in ::process() where this exception is caught.
```
oliverklee: :+1:

markcarver:
Done

oliverklee:
Merged. Thanks for contributing, @markcarver! :heart:

PR #404, opened 2017-09-30
Make inline style priority higher than css block priority

zoliszabo:
IMPORTANT! This fix may be a breaking change (depending on the input)!

Right now, original inline style rules are left at the beginning of the
final inline style string, which may be wrong in some cases
Consider
`<style>p {padding:10px}</style>`
`<p style="padding-left:20px"></p>`
This results in
`<p style="padding-left:20px;padding:10px"></p>`
which is wrong.
This fix helps to reverse priorities and it outputs
`<p style="padding:10px;padding-left:20px"></p>`

(Some existing tests had to be modified too...)

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -355,9 +355,12 @@ class Emogrifier
             foreach ($nodesWithStyleAttributes as $node) {
                 if ($this->isInlineStyleAttributesParsingEnabled) {
                     $this->normalizeStyleAttributes($node);
-                } else {
-                    $node->removeAttribute('style');
                 }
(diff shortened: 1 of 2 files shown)


oliverklee:
Hi @zoliszabo, thanks for the PR! It looks like PHP_CodeSniffer produces two warnings. Could you please fix them, re-push and then post a short comment here? Then I'll review your PR.

zoliszabo:
Hi @oliverklee, I fixed the code to comply with the coding style. Please review the fix and get back with your thoughts. Thanks!

oliverklee:
Merged. @zoliszabo, thanks for contributing! :heart:

PR #405, opened 2017-09-30
[BUGFIX] Fixed mapped html attribute values

zoliszabo:
Here is another PR with 2 small fixes regarding mapped HTML attributes:
 - Original inline styles were not considered before
 - Final "auto" values were not considered before

Please note that PR may introduce breaking changes.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -406,9 +406,6 @@ class Emogrifier
                     $oldStyleDeclarations = [];
                 }
                 $newStyleDeclarations = $this->parseCssDeclarationsBlock($cssRule['declarationsBlock']);
-                if ($this->shouldMapCssToHtml) {
-                    $this->mapCssToHtmlAttributes($newStyleDeclarations, $node);
-                }
                 $node->setAttribute(
                     'style',
                     $this->generateStyleStringFromDeclarationsArrays($oldStyleDeclarations, $newStyleDeclarations)
@@ -422,11 +419,45 @@ class Emogrifier
             $this->fillStyleAttributesWithMergedStyles();
         }
 
+        if ($this->shouldMapCssToHtml) {
+            $this->mapAllInlineStylesToHtmlAttributes($xPath);
+        }
+
         if ($this->shouldKeepInvisibleNodes) {
             $this->removeInvisibleNodes($xPath);
         }
 
         $this->copyCssWithMediaToStyleNode($xmlDocument, $xPath, $cssParts['media']);
+    }
+
+    /**
+     * Searches for all nodes with a style attribute, transforms the CSS found
+     * to HTML attributes and adds those attributes to each node.
+     *
+     * @param \DOMXPath $xPath
+     *
+     * @return void
+     *
+     * @throws \RuntimeException
+     */
+    private function mapAllInlineStylesToHtmlAttributes(\DOMXPath $xPath)
+    {
+        $nodesWithStyleAttributes = $xPath->query('//*[@style]');
+        if ($nodesWithStyleAttributes === false) {
+            if ($this->debug) {
+                throw new \RuntimeException(
+                    'Possibly malformed DOMXPath query expression or invalid DOMXPath object.',
+                    1508437884
+                );
+            }
+            return;
+        }
+
+        /** @var \DOMElement $node */
+        foreach ($nodesWithStyleAttributes as $node) {
+            $inlineStyleDeclarations = $this->parseCssDeclarationsBlock($node->getAttribute('style'));
+            $this->mapCssToHtmlAttributes($inlineStyleDeclarations, $node);
+        }
     }
 
     /**
--- a/Tests/Unit/EmogrifierTest.php
+++ b/Tests/Unit/EmogrifierTest.php
@@ -1997,10 +1997,7 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
         $this->subject->enableCssToHtmlMapping();
         $html = $this->subject->emogrify();
 
-        self::assertContains(
-            '<' . $tagName . ' ' . $attributes,
-            $html
-        );
+        self::assertRegExp('/<' . preg_quote($tagName, '/') . '[^>]+' . preg_quote($attributes, '/') . '/', $html);
     }
 
     /**
@@ -2120,5 +2117,50 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
             '<body>' . $style . '</body>',
             $result
         );
+    }
+
+    /**
+     * Asserts that $html contains a $tagName tag with the $attribute attribute.
+     *
+     * @param string $html the HTML string we are searching in
+     * @param string $tagName the HTML tag we are looking for
+     * @param string $attribute the attribute we are looking for (with or even without a value)
+     */
+    private function assertHtmlStringContainsTagWithAttribute($html, $tagName, $attribute)
+    {
+        self::assertTrue(
+            preg_match('/<' . preg_quote($tagName, '/') . '[^>]+' . preg_quote($attribute, '/') . '/', $html) > 0
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyPrefersInlineStyleOverCssBlockStyleForHtmlAttributesMapping()
+    {
+        $this->subject->setHtml(
+            '<html><head><style>p {width:1px}</style></head><body><p style="width:2px"></p></body></html>'
+        );
+        $this->subject->enableCssToHtmlMapping();
+
+        $result = $this->subject->emogrify();
+
+        $this->assertHtmlStringContainsTagWithAttribute($result, 'p', 'width="2"');
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyCorrectsHtmlAttributesMappingWhenMultipleMatchingRulesAndLastRuleIsAuto()
+    {
+        $this->subject->setHtml(
+            '<html><head><style>p {width:1px}</style></head><body><p class="autoWidth"></p></body></html>'
+        );
+        $this->subject->setCss('p.autoWidth {width:auto}');
+        $this->subject->enableCssToHtmlMapping();
+
+        $result = $this->subject->emogrify();
+
+        self::assertContains('<p class="autoWidth" style="width: auto;">', $result);
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 
 ### Fixed
+- Fixed mapped html attribute values
+  ([#405](https://github.com/MyIntervals/emogrifier/pull/405))
 - Make sure the HTML always has a BODY element
   ([#410](https://github.com/MyIntervals/emogrifier/pull/410))
 - Make inline style priority higher than css block priority


oliverklee:
Hi @zoliszabo, thanks for the PR! ❤️

I'd like to have one fix per PR only so the history is more clear, and it's easier to revert a commit if a fix breaks something. Could you please split up this PR into two separate PRs? Thanks!

zoliszabo:
Hey @oliverklee,

Actually there is one single small change that fixes both bugs. There isn't really a way to make two separate commits from it.
Maybe my PR description wasn't precise enough.

Thanks!

oliverklee:
@zoliszabo, ah, okay. Thanks for the clarification!

oliverklee:
@zoliszabo Could you please rebase and solve the merge conflict (now that #404 is merged)? Thanks!

zoliszabo:
@oliverklee Rebase done. Thanks!

Tests/Unit/EmogrifierTest.php
```diff
@@ -1997,10 +1997,7 @@ public function emogrifierMapsSuitableCssToHtmlIfFeatureIsEnabled($body, $css, $
         $this->subject->enableCssToHtmlMapping();
         $html = $this->subject->emogrify();
 
-        self::assertContains(
-            '<' . $tagName . ' ' . $attributes,
-            $html
-        );
+        self::assertRegExp('/<' . $tagName . '[^>]+' . $attributes . '/', $html);
```
oliverklee: You'll need to use preg_quote on $tagName (just to be consistent) and $attributes as you are now converting this matcher to a regular expression.

Tests/Unit/EmogrifierTest.php
```diff
@@ -2104,4 +2101,35 @@ public function emogrifyKeepsInlineStylePriorityVersusStyleBlockRules()
 
         self::assertContains('<p style="padding: 10px; padding-left: 20px;">', $result);
     }
+    
+    /**
+     * @test
+     */
+    public function emogrifyPrefersInlineStyleOverCSSBlockStyleForHtmlAttributesMapping()
```
oliverklee: CamelCase, please: CSS -> Css

Tests/Unit/EmogrifierTest.php
```diff
@@ -2104,4 +2101,35 @@ public function emogrifyKeepsInlineStylePriorityVersusStyleBlockRules()
...
+    {
+        $this->subject->setHtml(
+            '<html><head><style>p {width:1px}</style></head><body><p style="width:2px"></p></body></html>'
+        );
+        $this->subject->enableCssToHtmlMapping();
+
+        $result = $this->subject->emogrify();
+
+        self::assertContains('width="2"', $result);
```
oliverklee: This should be more specific, i.e., it should test that this is an attribute of the P node.

Tests/Unit/EmogrifierTest.php
```diff
@@ -2104,4 +2101,35 @@ public function emogrifyKeepsInlineStylePriorityVersusStyleBlockRules()
...
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyDoesCorrectHtmlAttributesMappingWhenMultipleMatchingRulesAndLastRuleIsAuto()
```
oliverklee: DoesCorrect -> Corrects

Tests/Unit/EmogrifierTest.php
```diff
@@ -2104,4 +2101,35 @@ public function emogrifyKeepsInlineStylePriorityVersusStyleBlockRules()
...
+    {
+        $this->subject->setHtml(
+            '<html><head><style>p {width:1px}</style></head><body><p class="autoWidth"></p></body></html>'
+        );
+        $this->subject->setCss('p.autoWidth {width:auto}');
+        $this->subject->enableCssToHtmlMapping();
+
+        $result = $this->subject->emogrify();
+
+        self::assertNotContains('width="1"', $result);
```
oliverklee: This also should be more specific (see above).

Classes/Emogrifier.php
```diff
@@ -422,6 +419,17 @@ protected function process(\DOMDocument $xmlDocument)
             $this->fillStyleAttributesWithMergedStyles();
         }
 
+        if ($this->shouldMapCssToHtml) {
```
oliverklee: Please let's move this to a new, well-named private method so this method doesn't grow any longer, and the good name will make understanding the method easier.

zoliszabo:
Implemented requested changes. @oliverklee please take another look. Thanks!

oliverklee:
@zoliszabo Thanks for the update! I'm currently reviewing the new version.

In the meantime, could you please add a changelog entry (newest on top)? Thanks!

Classes/Emogrifier.php
```diff
@@ -422,6 +419,10 @@ protected function process(\DOMDocument $xmlDocument)
...
+            $this->mapAllInlineStylesToHtmlAttributes($xPath);
+        }
```
oliverklee: Nice!

Classes/Emogrifier.php
```diff
@@ -430,6 +431,26 @@ protected function process(\DOMDocument $xmlDocument)
     }
 
     /**
+     * This method searches for all nodes with a style attribute, transforms the CSS found
+     * to HTML attributes and adds those attributes to each node
```
oliverklee: Just to make this perfect, could you please add the period at the end of the sentence and remove "This method" from the description (so that it starts with "Searches for …"?

Classes/Emogrifier.php
```diff
@@ -430,6 +431,26 @@ protected function process(\DOMDocument $xmlDocument)
...
+     *
+     * @param \DOMXPath $xPath
+     *
+     * @return void
+     */
+    private function mapAllInlineStylesToHtmlAttributes(\DOMXPath $xPath)
+    {
+        $nodesWithStyleAttributes = $xPath->query('//*[@style]');
+        if ($nodesWithStyleAttributes !== false) {
```
oliverklee: You can convert this to a guard clause to get rid of one level of indentation and make this easier to read:
``` php
if ($nodesWithStyleAttributes ==) false) {
    return;
}

/** @var … */
```

(Guard clauses are the only exception to the "don't have early returns" rule.)

Classes/Emogrifier.php
```diff
@@ -430,6 +431,26 @@ protected function process(\DOMDocument $xmlDocument)
...
+            /** @var \DOMElement $node */
+            foreach ($nodesWithStyleAttributes as $node) {
+                $inlineStyleDeclarations = $this->parseCssDeclarationsBlock($node->getAttribute('style'));
+                $this->mapCssToHtmlAttributes($inlineStyleDeclarations, $node);
+            }
```
oliverklee: Beautiful!

Tests/Unit/EmogrifierTest.php
```diff
@@ -1997,10 +1997,7 @@ public function emogrifierMapsSuitableCssToHtmlIfFeatureIsEnabled($body, $css, $
         $this->subject->enableCssToHtmlMapping();
         $html = $this->subject->emogrify();
 
-        self::assertContains(
-            '<' . $tagName . ' ' . $attributes,
-            $html
-        );
+        self::assertRegExp('/<' . preg_quote($tagName) . '[^>]+' . preg_quote($attributes) . '/', $html);
```
oliverklee: You'll also need to provide `'/'` as the second parameter to `preg_quote` so that the delimiter also will get escaped.

Tests/Unit/EmogrifierTest.php
```diff
@@ -2121,4 +2118,41 @@ public function emogrifyMovesStyleElementFromHeadToBody()
             $result
         );
     }
+
+    private function htmlStringContainsTagWithAttribute($html, $tagName, $attribute)
+    {
+        return preg_match('/<' . preg_quote($tagName) . '[^>]+' . preg_quote($attribute) . '/', $html) > 0;
```
oliverklee: You'll also need to provide `'/'` as the second parameter to `preg_quote` so that the delimiter also will get escaped.

Tests/Unit/EmogrifierTest.php
```diff
@@ -2121,4 +2118,41 @@ public function emogrifyMovesStyleElementFromHeadToBody()
             $result
         );
     }
+
+    private function htmlStringContainsTagWithAttribute($html, $tagName, $attribute)
```
oliverklee: Please add a PHPDoc string for with the `@param` and `@return` annotations.

Tests/Unit/EmogrifierTest.php
```diff
@@ -2121,4 +2118,41 @@ public function emogrifyMovesStyleElementFromHeadToBody()
...
+    {
+        return preg_match('/<' . preg_quote($tagName) . '[^>]+' . preg_quote($attribute) . '/', $html) > 0;
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyPrefersInlineStyleOverCssBlockStyleForHtmlAttributesMapping()
```
oliverklee: Great test method name! :+1:

Tests/Unit/EmogrifierTest.php
```diff
@@ -2121,4 +2118,41 @@ public function emogrifyMovesStyleElementFromHeadToBody()
...
+        $this->subject->setHtml(
+            '<html><head><style>p {width:1px}</style></head><body><p class="autoWidth"></p></body></html>'
+        );
+        $this->subject->setCss('p.autoWidth {width:auto}');
+        $this->subject->enableCssToHtmlMapping();
+
+        $result = $this->subject->emogrify();
+
+        self::assertNotContains('width="1"', $result);
+        self::assertFalse($this->htmlStringContainsTagWithAttribute($result, 'p', 'width="1"'));
```
oliverklee: This assertion is not needed anymore as it never can be true if the assertNotContains above is true.

Classes/Emogrifier.php
zoliszabo: Should we treat `false` (`$xPath->query(...)` error) and empty array (empty `DOMNodeList`) differently? Or a simple `== false` is enough?

Tests/Unit/EmogrifierTest.php
zoliszabo: Of course, this happens when you are in a hurry...

On the other hand, I think checking just for the attribute name without a value (`width=`) would be better here, because we should make sure that the `width` attribute is missing, not just that it does not have the value `1`. What do you think?

Classes/Emogrifier.php
```diff
@@ -430,6 +431,28 @@ protected function process(\DOMDocument $xmlDocument)
...
+     * to HTML attributes and adds those attributes to each node.
+     *
+     * @param \DOMXPath $xPath
+     *
+     * @return void
+     */
+    private function mapAllInlineStylesToHtmlAttributes(\DOMXPath $xPath)
+    {
+        $nodesWithStyleAttributes = $xPath->query('//*[@style]');
+        if ($nodesWithStyleAttributes == false) {
```
oliverklee: Generally, please always use strict comparisons: `=== false`

oliverklee: The check for `false` (the error case) should indeed be enough. If the node list is empty (which is a perfectly valid case), the `foreach` will just be a no-oop.

Actually, I think we should throw a `\RuntimeException` if the result is false because then either the expression is invalid (which would be a bug in Emogrifier) or the `$xPath` would be invalid (we which should catch earlier). Please also add the corresponding `@throws` annotation.

oliverklee: I love how refactoring brings buried subtle bugs and not-so-clean programming to the surface! :heart:

Tests/Unit/EmogrifierTest.php
```diff
@@ -2121,4 +2118,48 @@ public function emogrifyMovesStyleElementFromHeadToBody()
...
+    {
+        $this->subject->setHtml(
+            '<html><head><style>p {width:1px}</style></head><body><p class="autoWidth"></p></body></html>'
+        );
+        $this->subject->setCss('p.autoWidth {width:auto}');
+        $this->subject->enableCssToHtmlMapping();
+
+        $result = $this->subject->emogrify();
+
+        self::assertFalse($this->htmlStringContainsTagWithAttribute($result, 'p', 'width='));
```
oliverklee: > On the other hand, I think checking just for the attribute name without a value (width=) would be better here, because we should make sure that the width attribute is missing, not just that it does not have the value 1. What do you think?

Yes, I agree.

This would also allow us to change the private method to `assertHtmlStringContainsTagWithAttribute` (so that it already contains the `assert*` call in it).

Classes/Emogrifier.php
zoliszabo: Throw the exception in debug mode only?

Tests/Unit/EmogrifierTest.php
zoliszabo: That was the first version of the private method, however here we need to make sure that there is no `$attribute` attribute for `$tagName`... and then `assertHtmlStringContainsTagWithAttribute` is not OK anymore. This is why I modified it to a `bool` function and used `assertTrue`/`assertFalse`. Am I missing something? (it's late here)

oliverklee: Can't we then use a basic assertContains here instead?

Classes/Emogrifier.php
oliverklee: Yes, that makes sense.

Tests/Unit/EmogrifierTest.php
zoliszabo: Yes, `assert*Not*Contains` should be enough.

oliverklee: I was thinking of using an `assertContains` on the expected tag instead (by which we also test that the generated tag looks a certain way and not any other way).

zoliszabo:
Hi @oliverklee, latest feedback was implemented. Please check it out. Thanks!

oliverklee:
Merged. Thanks for contributing, @zoliszabo! :heart:

PR #406, opened 2017-10-12
[CLEANUP] Simplify and clean up the test case (part 1)

oliverklee:
- remove superfluous words from test names
- make the tests for adding HEAD and HTML by default clearer
- add a test case about not adding the BODY element


oliverklee:
Hold on, I'd like to put the autoformat in a separate PR …

PR #407, opened 2017-10-12
[CLEANUP] Autoformat the Emogrifier class

oliverklee:


oliverklee:
It looks like the autoformat broke something. I'm looking into it.

oliverklee:
This should be green now.

PR #419, opened 2017-10-21
[CLEANUP] Refactor translateCssToXpathPass for callbacks

oliverklee:
Now a callback can be used for matching the class attributes, making
inline matching possible.

This is the final refactoring before #395.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -1385,18 +1385,34 @@ class Emogrifier
     }
 
     /**
+     * Flexibly translates the CSS selector $trimmedLowercaseSelector to an xPath selector.
+     *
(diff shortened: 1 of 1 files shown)


jjriv:
I agree this should be the default. I can only think of one issue that would conflict, and that is when the contents of the <style></style> tags are left intact. @oliverklee you would know more about this than I because you have taken emogrifier to the next level. I'm not as familiar with the code. 

That said, as long as the inline styles won't get overriden by other CSS, i do think this should be the default behavior and will be nice in keeping the inline CSS clean.

oliverklee:
@jjriv Were you referring to the change in #420 with your comment?

jjriv:
Doh! Yes I was. Thx for catching that.

Issue #378, opened 2017-03-21
Remove the !important flag from CSS rules


maoxiong:
Outlook doesn't support the !important tag in any inline styles, so this should be removed. (If the !important flag is used, it actually ignores the whole rule!)

There is no need for it anyway, as the stylesheets are stripped out anyway.

See: https://www.campaignmonitor.com/blog/email-marketing/2010/05/outlook-2007-and-the-inline-important-declaration/

oliverklee:
Hi @maoxiong, are you referring to the style tags that are kept in the HEAD section of the document or to the non-CSS styles in the HTML nodes that Emogrifier creates out of the CSS rules?

oliverklee:
And could you please post/attach the HTML and CSS which you put in, the actual resulting HTML, and the HTML that you would expect to get generated? Thanks!

maoxiong:
Hi Oliver,

@oliverklee , Sure this is an easy one. Example attached.

Emogrifier is correctly changing the DIV's inline style, but but also putting the "!important" part in, thus making it ignored in Outlook.

As the styles are all inlined  (and the original head style tags stripped) - there is no need to leave that !important in there anyway :)

Of course the media query css in the style tags should be left as is, but that bit is working fine.
[noimportant.txt](https://github.com/jjriv/emogrifier/files/864064/noimportant.txt)

oliverklee:
Thanks!

oliverklee:
Would you be willing to create a pull request for this (including unit tests and the fix)?

maoxiong:
Hi Oliver - I would love to be able to say yes, but I have no clue how to do that :(

I'm good at finding bugs, but not so good at fixing them! haha.

oliverklee:
This was fixed in #420.

PR #420, opened 2017-10-22
[FEATURE] Remove `!important` from final inline style declarations

zoliszabo:
This solves issue https://github.com/MyIntervals/emogrifier/issues/378,
although I consider this is not a bugfix, but enhancement (which in the
same time happens to fix incorrect behaviour of some email clients).

It was implemented as an option (`$shouldRemoveImportantAnnotation`) and
it is enabled by default, because it renders output simpler, while keeping
it 100% compatible with old output.

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -372,7 +372,6 @@ class Emogrifier
         // grab any existing style blocks from the html and append them to the existing CSS
         // (these blocks should be appended so as to have precedence over conflicting styles in the existing CSS)
         $allCss = $this->css;
-
         if ($this->isStyleBlocksParsingEnabled) {
             $allCss .= $this->getCssFromAllStyleNodes($xPath);
         }
@@ -400,7 +399,6 @@ class Emogrifier
                 if (in_array($node, $excludedNodes, true)) {
                     continue;
                 }
-
                 // if it has a style attribute, get it, process it, and append (overwrite) new stuff
                 if ($node->hasAttribute('style')) {
                     // break it up into an associative array
@@ -416,8 +414,6 @@ class Emogrifier
             }
         }
 
-        restore_error_handler();
-
         if ($this->isInlineStyleAttributesParsingEnabled) {
             $this->fillStyleAttributesWithMergedStyles();
         }
@@ -430,7 +426,11 @@ class Emogrifier
             $this->removeInvisibleNodes($xPath);
         }
 
+        $this->removeImportantAnnotationFromAllInlineStyles($xPath);
+
         $this->copyCssWithMediaToStyleNode($xmlDocument, $xPath, $cssParts['media']);
+
+        restore_error_handler();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


oliverklee:
@zoliszabo Thanks for the PR! Do you see any reason not to remove the option and make this the new default?

@jjriv What do you think?

zoliszabo:
Hi @oliverklee! It would be the new default. I put the option to be able to switch back to the old behavior. Although, I do not see any reason someone would want to, but... who knows. Let me know what you decide and I'll modify the code. Thanks!

oliverklee:
@zoliszabo Yes, then please let's make this the new default.

Could you also please rebase? Thanks!

zoliszabo:
Hi @oliverklee! I've removed the option `$shouldRemoveImportantAnnotation`, now `!important` annotations are stripped out automatically from the final inline CSS declarations. Also did the rebase. Please have a look. Thanks!

Classes/Emogrifier.php
```diff
@@ -430,7 +428,11 @@ protected function process(\DOMDocument $xmlDocument)
             $this->removeInvisibleNodes($xPath);
         }
 
+        $this->removeImportantAnnotationFromAllInlineStyles($xPath);
+
         $this->copyCssWithMediaToStyleNode($xmlDocument, $xPath, $cssParts['media']);
+
+        restore_error_handler();
```
zoliszabo: Moved `restore_error_handler()` here, because both `removeImportantAnnotationFromAllInlineStyles()` and `copyCssWithMediaToStyleNode()` make xPath queries.

Tests/Unit/EmogrifierTest.php
```diff
@@ -2184,4 +2184,69 @@ public function emogrifyCorrectsHtmlAttributesMappingWhenMultipleMatchingRulesAn
...
+            ],
+        ];
+    }
+
+    /**
+     * @test
+     *
+     * @dataProvider cssForImportantRuleRemovalDataProvider
+     */
+    public function emogrifyRemovesImportantRule($originalStyleAttributeContent, $expectedStyleAttributeContent)
```
oliverklee: Could you please add the `@param` annotations?

Tests/Unit/EmogrifierTest.php
```diff
@@ -2184,4 +2184,69 @@ public function emogrifyCorrectsHtmlAttributesMappingWhenMultipleMatchingRulesAn
...
+                'color: red; float: left; clear: both; width: 1px; height: 1px;'
+            ],
+            'multiple declarations, multiple not consecutive !important rules somewhere in the middle' => [
+                'color: red; width: 1px !important; clear: both; height: 1px !important; float: left',
+                'color: red; clear: both; float: left; width: 1px; height: 1px;'
+            ],
+            'multiple declarations, multiple !important rules at the end' => [
+                'color: red; float: left; width: 1px !important; height: 1px !important',
+                'color: red; float: left; width: 1px; height: 1px;'
+            ],
```
oliverklee: Beautiful!

Classes/Emogrifier.php
```diff
@@ -464,6 +466,85 @@ private function mapAllInlineStylesToHtmlAttributes(\DOMXPath $xPath)
...
+            }
+        }
+        $inlineStyleDeclarationsInNewOrder = array_merge(
+            $regularStyleDeclarations,
+            $importantStyleDeclarations
+        );
+        $node->setAttribute(
+            'style',
+            $this->generateStyleStringFromSingleDeclarationsArray($inlineStyleDeclarationsInNewOrder)
+        );
```
oliverklee: Great variable names!

Classes/Emogrifier.php
```diff
@@ -464,6 +466,85 @@ private function mapAllInlineStylesToHtmlAttributes(\DOMXPath $xPath)
     }
 
     /**
+     * Searches for all nodes with a style attribute and removes the "!important" annotations out of
```
oliverklee: The term here is not correct: According to https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#The_!important_exception, `!important` is a rule, not an annotation.

Classes/Emogrifier.php
```diff
@@ -464,6 +466,85 @@ private function mapAllInlineStylesToHtmlAttributes(\DOMXPath $xPath)
...
+     * the inline style declarations, eventually by rearranging declarations.
+     *
+     * @uses \Pelago\Emogrifier::removeImportantAnnotationFromNodeInlineStyle
```
oliverklee: Please remove the `@uses` annotation. I don't see the benefit here.

Classes/Emogrifier.php
```diff
@@ -464,6 +466,85 @@ private function mapAllInlineStylesToHtmlAttributes(\DOMXPath $xPath)
...
+     *
+     * @param \DOMXPath $xPath
+     *
+     * @return void
+     */
+    private function removeImportantAnnotationFromAllInlineStyles(\DOMXPath $xPath)
+    {
+        $nodesWithStyleAttribute = $this->getAllNodesWithStyleAttribute($xPath);
```
oliverklee: As this method is the only method calling getAllNodesWithStyleAttribute, we could always throw the exception there and then catch it here and either do an early return or rethrow the exception in debug mode.

Classes/Emogrifier.php
```diff
@@ -430,7 +428,11 @@ protected function process(\DOMDocument $xmlDocument)
             $this->removeInvisibleNodes($xPath);
         }
 
+        $this->removeImportantAnnotationFromAllInlineStyles($xPath);
+
         $this->copyCssWithMediaToStyleNode($xmlDocument, $xPath, $cssParts['media']);
```
oliverklee: Could you please check whether we could remove some blank lines from this method so the code lines are grouped together (so that related operations stand together)?

Classes/Emogrifier.php
zoliszabo: According to https://www.w3.org/TR/css3-cascade/#importance - which seems to be part of the official specification of CSS in 2017 (https://www.w3.org/TR/CSS/#css) - it is an annotation. Also, "rule" can be confusing, because it is mainly used as the name for a selector + list of declarations unit. But I do not want to force anything, if you confirm that I should change 'annotation' with 'rule', I'll do it :)

Classes/Emogrifier.php
zoliszabo: OK.
As we are talking about the method `getAllNodesWithStyleAttribute()`, I am planning to use it in two other places, where `$xPath->query('//*[@style]')` is now directly used. What do you think? Should I do these modifications within this PR or submit another one after this is merged?

Classes/Emogrifier.php
```diff
@@ -464,6 +466,85 @@ private function mapAllInlineStylesToHtmlAttributes(\DOMXPath $xPath)
...
+            $this->generateStyleStringFromSingleDeclarationsArray($inlineStyleDeclarationsInNewOrder)
+        );
+    }
+
+    /**
+     * Returns a list with all DOM nodes that have a style attribute.
+     *
+     * @param \DOMXPath $xPath
+     *
+     * @return \DOMNodeList
```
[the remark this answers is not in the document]
zoliszabo: Actually, it will never return `false`, only a `\DOMNodeList` (empty or not) or throws the exception.

Classes/Emogrifier.php
```diff
@@ -464,6 +466,85 @@ private function mapAllInlineStylesToHtmlAttributes(\DOMXPath $xPath)
...
+     *
+     * @throws \RuntimeException
+     */
+    private function getAllNodesWithStyleAttribute(\DOMXPath $xPath)
+    {
+        $nodesWithStyleAttribute = $xPath->query('//*[@style]');
+        if ($nodesWithStyleAttribute === false) {
+            if ($this->debug) {
```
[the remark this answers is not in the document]
zoliszabo: The condition `if ($this->debug)` is not used here anymore, but a level higher, just like you recommended above: https://github.com/MyIntervals/emogrifier/pull/420#discussion_r146406532.

zoliszabo:
Hi @oliverklee. I've pushed an update, please take a look. Thanks!

Classes/Emogrifier.php
oliverklee: Thanks for the link! Then let's keep "annotation". (To me, the W3 specs are indeed a lot more explicit on that it's an annotation, not a rule, than the MDN.)

oliverklee:
Merged. Thanks for contributing, @zoliszabo! ❤️ 🍪

PR #425, opened 2017-10-28
[CLEANUP] Use method getAllNodesWithStyleAttribute() everywhere...

zoliszabo:
...where searching for nodes with a `style` attribute.

Had to modify method `handleXpathError()` to also transform eventual PHP warnings generated by
`getAllNodesWithStyleAttribute()` into `\RuntimeException`.

Also had to move style normalization code from the beginning of `process()` into a new private method
called `normalizeStyleAttributesOfAllNodes()`, because phpcs was not letting public method `process`
throw 2 different types of exceptions (and it seemed to be a good idea moving that code block anyway).

### DIFF
--- a/Classes/Emogrifier.php
+++ b/Classes/Emogrifier.php
@@ -346,28 +346,11 @@ class Emogrifier
         $xPath = new \DOMXPath($xmlDocument);
         $this->clearAllCaches();
 
-        // Before be begin processing the CSS file, parse the document and normalize all existing CSS attributes.
-        // This changes 'DISPLAY: none' to 'display: none'.
-        // We wouldn't have to do this if DOMXPath supported XPath 2.0.
-        // Also store a reference of nodes with existing inline styles so we don't overwrite them.
         $this->purgeVisitedNodes();
 
         set_error_handler([$this, 'handleXpathError'], E_WARNING);
 
-        $nodesWithStyleAttributes = $xPath->query('//*[@style]');
-        if ($nodesWithStyleAttributes !== false) {
-            /** @var \DOMElement $node */
-            foreach ($nodesWithStyleAttributes as $node) {
-                if ($this->isInlineStyleAttributesParsingEnabled) {
-                    $this->normalizeStyleAttributes($node);
-                }
-                // Remove style attribute in every case, so we can add them back (if inline style attributes
-                // parsing is enabled) to the end of the style list, thus keeping the right priority of CSS rules;
-                // else original inline style rules may remain at the beginning of the final inline style definition
-                // of a node, which may give not the desired results
-                $node->removeAttribute('style');
-            }
-        }
+        $this->normalizeStyleAttributesOfAllNodes($xPath);
 
         // grab any existing style blocks from the html and append them to the existing CSS
         // (these blocks should be appended so as to have precedence over conflicting styles in the existing CSS)
@@ -445,19 +428,17 @@ class Emogrifier
      */
     private function mapAllInlineStylesToHtmlAttributes(\DOMXPath $xPath)
     {
-        $nodesWithStyleAttributes = $xPath->query('//*[@style]');
-        if ($nodesWithStyleAttributes === false) {
+        try {
+            $nodesWithStyleAttribute = $this->getAllNodesWithStyleAttribute($xPath);
+        } catch (\RuntimeException $e) {
             if ($this->debug) {
-                throw new \RuntimeException(
-                    'Possibly malformed DOMXPath query expression or invalid DOMXPath object.',
-                    1508437884
-                );
+                throw new \RuntimeException($e->getMessage(), 1508437884, $e);
             }
             return;
         }
 
         /** @var \DOMElement $node */
-        foreach ($nodesWithStyleAttributes as $node) {
+        foreach ($nodesWithStyleAttribute as $node) {
             $inlineStyleDeclarations = $this->parseCssDeclarationsBlock($node->getAttribute('style'));
             $this->mapCssToHtmlAttributes($inlineStyleDeclarations, $node);
         }
@@ -479,10 +460,11 @@ class Emogrifier
             $nodesWithStyleAttribute = $this->getAllNodesWithStyleAttribute($xPath);
         } catch (\RuntimeException $e) {
             if ($this->debug) {
-                throw($e);
+                throw new \RuntimeException($e->getMessage(), 1509193363, $e);
             }
             return;
         }
+
         foreach ($nodesWithStyleAttribute as $node) {
             $this->removeImportantAnnotationFromNodeInlineStyle($node);
         }
@@ -525,21 +507,15 @@ class Emogrifier
     /**
      * Returns a list with all DOM nodes that have a style attribute.
      *
+     * Can indirectly generate a \RuntimeException via method handleXpathError().
+     *
      * @param \DOMXPath $xPath
      *
      * @return \DOMNodeList
-     *
-     * @throws \RuntimeException
      */
     private function getAllNodesWithStyleAttribute(\DOMXPath $xPath)
     {
         $nodesWithStyleAttribute = $xPath->query('//*[@style]');
-        if ($nodesWithStyleAttribute === false) {
-            throw new \RuntimeException(
-                'Possibly malformed DOMXPath query expression or invalid DOMXPath object.',
-                1508961431
-            );
-        }
         return $nodesWithStyleAttribute;
     }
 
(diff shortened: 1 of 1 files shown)


Classes/Emogrifier.php
```diff
@@ -445,19 +428,21 @@ protected function process(\DOMDocument $xmlDocument)
...
+        try {
+            $nodesWithStyleAttribute = $this->getAllNodesWithStyleAttribute($xPath);
+        } catch (\RuntimeException $e) {
             if ($this->debug) {
                 throw new \RuntimeException(
-                    'Possibly malformed DOMXPath query expression or invalid DOMXPath object.',
-                    1508437884
+                    $e->getMessage(),
+                    1508437884,
+                    $e
```
oliverklee: Nitpick: I think this will fit in one line again. Could you please reformat this?

Classes/Emogrifier.php
```diff
@@ -479,7 +464,11 @@ private function removeImportantAnnotationFromAllInlineStyles(\DOMXPath $xPath)
             $nodesWithStyleAttribute = $this->getAllNodesWithStyleAttribute($xPath);
         } catch (\RuntimeException $e) {
             if ($this->debug) {
-                throw($e);
+                throw new \RuntimeException(
+                    $e->getMessage(),
+                    1509193363,
+                    $e
```
oliverklee: Ditto (about trying to keep this on one line).

Classes/Emogrifier.php
```diff
@@ -479,7 +464,11 @@ private function removeImportantAnnotationFromAllInlineStyles(\DOMXPath $xPath)
             $nodesWithStyleAttribute = $this->getAllNodesWithStyleAttribute($xPath);
         } catch (\RuntimeException $e) {
             if ($this->debug) {
-                throw($e);
+                throw new \RuntimeException(
+                    $e->getMessage(),
+                    1509193363,
```
oliverklee: We can just rethrow `$e` here, I think.

Classes/Emogrifier.php
```diff
@@ -445,19 +428,21 @@ protected function process(\DOMDocument $xmlDocument)
...
-        if ($nodesWithStyleAttributes === false) {
+        try {
+            $nodesWithStyleAttribute = $this->getAllNodesWithStyleAttribute($xPath);
+        } catch (\RuntimeException $e) {
             if ($this->debug) {
                 throw new \RuntimeException(
-                    'Possibly malformed DOMXPath query expression or invalid DOMXPath object.',
-                    1508437884
+                    $e->getMessage(),
+                    1508437884,
```
oliverklee: We can just rethrow `$e` here, I think.

Oh, now that I think about this, this might result in the same exception code used in several places. How can we solve this in an elegant way?

Classes/Emogrifier.php
```diff
@@ -533,12 +522,10 @@ private function removeImportantAnnotationFromNodeInlineStyle(\DOMElement $node)
...
-        if ($nodesWithStyleAttribute === false) {
-            throw new \RuntimeException(
-                'Possibly malformed DOMXPath query expression or invalid DOMXPath object.',
-                1508961431
-            );
+        try {
+            $nodesWithStyleAttribute = $xPath->query('//*[@style]');
+        } catch (\RuntimeException $e) {
+            throw($e);
         }
```
oliverklee: In this case, we can just not catch it and remove the catch and re-throw construct here (if I'm reading this correctly).

Classes/Emogrifier.php
```diff
@@ -952,6 +939,44 @@ private function removeInvisibleNodes(\DOMXPath $xPath)
     }
 
     /**
+     * Before be begin processing the CSS file, parse the document and normalize all existing CSS attributes.
```
oliverklee: be -> we

Classes/Emogrifier.php
```diff
@@ -952,6 +939,44 @@ private function removeInvisibleNodes(\DOMXPath $xPath)
...
+                }
+                // Remove style attribute in every case, so we can add them back (if inline style attributes
+                // parsing is enabled) to the end of the style list, thus keeping the right priority of CSS rules;
+                // else original inline style rules may remain at the beginning of the final inline style definition
+                // of a node, which may give not the desired results
+                $node->removeAttribute('style');
+            }
+        } catch (\RuntimeException $e) {
+            if ($this->debug) {
+                throw new \RuntimeException(
```
oliverklee: We can just rethrow `$e` here, I think.

Classes/Emogrifier.php
```diff
@@ -952,6 +939,44 @@ private function removeInvisibleNodes(\DOMXPath $xPath)
...
+     * @param \DOMXPath $xPath
+     *
+     * @return void
+     *
+     * @throws \RuntimeException
+     */
+    private function normalizeStyleAttributesOfAllNodes(\DOMXPath $xPath)
+    {
+        try {
+            $nodesWithStyleAttribute = $this->getAllNodesWithStyleAttribute($xPath);
```
oliverklee: This is the only part that could cause the exception if I'm reading this correctly. So we could move the exception handling to directly below/around this line, moving the data wrangling out of the error handling.

Classes/Emogrifier.php
```diff
@@ -1767,6 +1793,18 @@ public function handleXpathError($type, $message, $file, $line, array $context)
...
+        // Catches eventual warnings generated by method getAllNodesWithStyleAttribute()
+        if ($type === E_WARNING && isset($context['xPath'])) {
+            throw new \RuntimeException(
+                sprintf(
+                    '%s in %s on line %s',
+                    $message,
+                    $file,
+                    $line
+                )
+            );
```
oliverklee: While you're at it, could you please add unique exception codes to this exception and the one above?

And: will the `sprintf` parameters fit in one line, and could you please use numbered parameters?

Classes/Emogrifier.php
zoliszabo: Same exception code is why I chose to create a new exception object and specify the original as previous in the exception chain (the third parameter) instead of simply re-throwing $e. This way, a distinct exception code can be defined in every place, where `getAllNodesWithStyleAttribute()` is being used.

Classes/Emogrifier.php
zoliszabo: For the same reason as above, I chose to create a new exception object to be able to specific a distinct exception code.

Classes/Emogrifier.php
zoliszabo: Same reason as above, created new Exception object to be able to define a different exception code just for this method.
Do you have any suggestions instead on how to do it better? Thanks!

Classes/Emogrifier.php
zoliszabo: `sprintf`: Do you mean using `%u` for the exception code or `%1$s` and `%2$s` (as in argument swapping)?

Classes/Emogrifier.php
oliverklee: Ah, makes sense. Thanks for the explanation!

Classes/Emogrifier.php
oliverklee: Actually, both (numbered arguments and typed arguments).

Classes/Emogrifier.php
oliverklee: The point of the exception code is to help find the place where the exception originally has been thrown. So I don't think we need to a different code here. And to find out the place where the exception gets re-thrown, the stack trace of the exception can be used. This will help keep the code simple. What do you think?

Classes/Emogrifier.php
```diff
@@ -533,12 +522,10 @@ private function removeImportantAnnotationFromNodeInlineStyle(\DOMElement $node)
...
     private function getAllNodesWithStyleAttribute(\DOMXPath $xPath)
     {
-        $nodesWithStyleAttribute = $xPath->query('//*[@style]');
-        if ($nodesWithStyleAttribute === false) {
-            throw new \RuntimeException(
-                'Possibly malformed DOMXPath query expression or invalid DOMXPath object.',
-                1508961431
-            );
+        try {
+            $nodesWithStyleAttribute = $xPath->query('//*[@style]');
```
oliverklee: Maybe we can also drop some of the parts where exceptions are caught and then silently ignored for cases where this points to a programming error in Emogrifier, not to not-so-clean (or broken) HTML/CSS. What do you think?

oliverklee: And if we do so, we could do this in a separate clean-up PR before this one so we keep things (more) separate.

oliverklee:
> Do you mean just throwing the RuntimeException in the handleXpathError() method based on the warning (like done in this PR) and then just simply not deal with this exception anymore in the other methods?

Yes (for exceptions that either can occur only due to programming errors, or that we catch somewhere when not in debug mode).

oliverklee:
@zoliszabo Any news? (I'd like to get this in the next release.)

zoliszabo:
Hi @oliverklee, I was busy travelling in the past period and wasn't able to concentrate here, but I will sort this out in the next days.

oliverklee:
Thanks! :heart:

oliverklee:
Marking as invalid as this has been moved to other PRs. Keeping this open until everything is merged somewhere else.

oliverklee:
@zoliszabo Is there anything left in this PR (to be done in separate PRs), or can I close this one?

zoliszabo:
@oliverklee Nothing left, it can be closed.

oliverklee:
Great, thanks!

PR #437, opened 2017-11-26
[BUGFIX] Throw the correct exception for invalid excluded selectors

zoliszabo:
Invalid excluded selectors were not handled before. Accidentally, a
`\RuntimeException` was thrown for these cases by `handleXpathError()`.

Now an `\InvalidArgumentException` is thrown only when in debug mode.
Else, the invalid selector is simply ignored.


Classes/Emogrifier.php
```diff
@@ -1756,6 +1766,20 @@ public function handleXpathError($type, $message, $file, $line, array $context)
             );
         }
 
+        // Catches eventual warnings generated by invalid selectors in method getNodesToExclude()
+        if ($type === E_WARNING && isset($context['selectorToExclude'])) {
```
oliverklee: This can be unified to a `if ($type === E_WARNING)` around both cases.

Tests/Unit/EmogrifierTest.php
```diff
@@ -1766,6 +1766,19 @@ public function removeExcludedSelectorGetsMatchingElementsToBeEmogrifiedAgain()
         self::assertContains('<p class="x" style="margin: 0;"></p>', $result);
     }
 
+    /**
+     * @test
+     *
+     * @expectedException \InvalidArgumentException
+     */
+    public function emogrifyThrowsInvalidArgumentExceptionForInvalidExcludedSelectorInDebugMode()
```
oliverklee: We also need a test for the non-debug case (that tests that no exception is caught, e.g. indirectly by checking that some particular HTML is returned).

Tests/Unit/EmogrifierTest.php
```diff
@@ -1766,6 +1766,19 @@ public function removeExcludedSelectorGetsMatchingElementsToBeEmogrifiedAgain()
         self::assertContains('<p class="x" style="margin: 0;"></p>', $result);
     }
 
+    /**
+     * @test
+     *
+     * @expectedException \InvalidArgumentException
```
oliverklee: Please use 'expectException' instead. It's the recommended way now: https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.exceptions

(I've created issue #438 for converting the other tests.)

Classes/Emogrifier.php
zoliszabo: Actually, since the `handleXpathError()` method is only assigned to the E_WARNING type, I chose to get rid of the `if ($type === E_WARNING)` conditions.

Maybe the `handleXpathError()` method should be renamed to better reflect its scope on warnings only? What do you think?

zoliszabo:
Hi @oliverklee! Feedback implemented. Please also see my comment at https://github.com/MyIntervals/emogrifier/pull/437#discussion_r153071293.
Many thanks!

Classes/Emogrifier.php
zoliszabo: Had to add back condition `if ($type === E_WARNING)` (although it's always true), because phpcs was raising a warning because of parameter `$type` not being used anymore. Is there a better solution here?

oliverklee: Yes, please let's rename it.

oliverklee: As for the unused parameter, you can add an `@codingStandardsIgnoreLine` anotation to the line above it.

Classes/Emogrifier.php
```diff
@@ -1741,27 +1751,43 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
+                        $line
+                    ),
+                    1511683740
+                );
+            }
+
+            // Catches eventual warnings generated by method getAllNodesWithStyleAttribute()
+            if (isset($context['xPath'])) {
+                throw new \RuntimeException(
+                    sprintf('%1$s in %2$s on line %3$u', $message, $file, $line),
```
oliverklee: Here, the some parameters to `sprintf` are missing.

zoliszabo: I am not sure what you mean here.

oliverklee:
@zoliszabo Could you please ping me here in the PR once you've pushed a new version with the unused parameter whitelisted for PHPCS?

zoliszabo:
@oliverklee Renamed `handleXpathError()` to `handleXpathQueryWarnings()`. Added `@codingStandardsIgnoreLine`, like suggested - however, I am not sure whether this line of comment detaches the phpdoc block from its method (e.g. NetBeans loses connection)...

Classes/Emogrifier.php
```diff
@@ -1739,29 +1749,46 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
-                    $file,
-                    $line
-                ),
-                1509279985
-            );
-        }
+    // @codingStandardsIgnoreLine
+    public function handleXpathQueryWarnings($type, $message, $file, $line, array $context)
+    {
+        if ($type === E_WARNING) {
```
oliverklee: Wasn't the idea to remove this if? Otherwise, `$type` would not be unused, and we wouldn't need to silence PHP_CodeSniffer about the unused parameter.

oliverklee:
When `$type` actually is unused, we can also use and end-line comment for the line with the method declaration to avoid the problem with NetBeans:
```
public function handleXpathQueryWarnings($type, $message, $file, $line, array $context) // @codingStandardsIgnoreLine
```

This will cause the line to be too long, but as the codesniffer ignores that line, that will not be a problem. ;-)

oliverklee:
And could you also please rebase (just to be safe …)? Thanks!

Tests/Unit/EmogrifierTest.php
```diff
@@ -1766,6 +1766,46 @@ public function removeExcludedSelectorGetsMatchingElementsToBeEmogrifiedAgain()
...
+        $this->subject->setHtml('<html><p class="x"></p></html>');
+        $this->subject->addExcludedSelector('..p');
+        $result = $this->subject->emogrify();
+        self::assertContains('<p class="x"></p>', $result);
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyIgnoresOnlyInvalidExcludedSelectorWhenNotInDebugMode()
```
oliverklee: Great tests, by the way! :+1:

Classes/Emogrifier.php
zoliszabo: @oliverklee Of course it was... I was just rushing to submit... Removed the extra condition, moved phpcs comment to the end of the line, did the rebase. Please check. And sorry.

oliverklee: Looks good. And there's no need to apologize. Forgetting/missing things just happens, that's normal, and that's what a second pair of eyes is for. :-)

oliverklee:
Merged. Thanks for the PR, @zoliszabo! ❤️

Issue #441, opened 2017-12-27
Invalid / unrecognized CSS selectors in media query blocks are not handled


zoliszabo:
Invalid / unrecognized (not implemented) CSS selectors from within media query blocks are not handled and they are wrongfully causing a `\RuntimeException`.

oliverklee:
This indeed should only throw an exception in debug mode, and be silently ignored in non-debug mode.

I see there's already a branch for this. Looking forward to reviewing the PR! :heart:

zoliszabo:
Fixed with PR https://github.com/MyIntervals/emogrifier/pull/442.

Issue #439, opened 2017-12-15
Fatal error: Uncaught RuntimeException: DOMXPath::query():


nikmauro:
Fatal error: Uncaught RuntimeException: DOMXPath::query(): Invalid expression in /var/www/vhosts/.com/httpdocs/vendor/pelago/emogrifier/Classes/Emogrifier.php on line 1108 in /var/www/vhosts/.com/httpdocs/vendor/pelago/emogrifier/Classes/Emogrifier.php:1785 Stack trace: #0 [internal function]: Pelago\Emogrifier->handleXpathQueryWarnings(2, 'DOMXPath::query...', '/var/www/vhosts...', 1108, Array) #1 /var/www/vhosts/.com/httpdocs/vendor/pelago/emogrifier/Classes/Emogrifier.php(1108): DOMXPath->query('//*[contains(co...') #2 /var/www/vhosts/.com/httpdocs/vendor/pelago/emogrifier/Classes/Emogrifier.php(1063): Pelago\Emogrifier->existsMatchForCssSelector(Object(DOMXPath), '.mobile_link a[...') #3 /var/www/vhosts/.com/httpdocs/vendor/pelago/emogrifier/Classes/Emogrifier.php(412): Pelago\Emogrifier->copyCssWithMediaToStyleNode(Object(DOMDocument), Object(DOMXPath), '@media screen a...') #4 /var/www/vhosts/.com/httpdocs/vendor/pelago/emogrifier/Classes/Emogrifier.php(328): Pelago\Em in /var/www/vhosts/.com/httpdocs/vendor/pelago/emogrifier/Classes/Emogrifier.php on line 1785

```CSS

a[href^="tel"], a[href^="sms"] {
                text-decoration: none;
                color: black; /* or whatever your want */
                pointer-events: none;
                cursor: default;
            }

            .mobile_link a[href^="tel"], .mobile_link a[href^="sms"] {
                text-decoration: default;
                color: orange !important; /* or whatever your want */
                pointer-events: auto;
                cursor: default;
            }

 @media only screen and (min-device-width: 768px) and (max-device-width: 1024px) {
            a[href^="tel"], a[href^="sms"] {
                text-decoration: none;
                color: blue; /* or whatever your want */
                pointer-events: none;
                cursor: default;
            }
```

oliverklee:
Hi @nikmauro, thanks for the bug report!

[...]

PR #442, opened 2017-12-27
[BUGFIX] Handle invalid/unrecognized selectors in media query blocks

zoliszabo:
Solves issue #441:
 - Throw `\InvalidArgumentException` for invalid/unrecognized selectors in
   debug mode;
 - Keep media query blocks containing invalid/non-recognized selectors when
   not in debug mode.


Classes/Emogrifier.php
```diff
@@ -1097,15 +1097,26 @@ private function extractMediaQueriesFromCss($css)
...
     private function existsMatchForCssSelector(\DOMXPath $xPath, $cssSelector)
     {
-        $nodesMatchingSelector = $xPath->query($this->translateCssToXpath($cssSelector));
+        try {
+            $nodesMatchingSelector = $xPath->query($this->translateCssToXpath($cssSelector));
+        } catch (\InvalidArgumentException $e) {
+            if ($this->debug) {
+                throw $e;
+            }
+            return true;
```
oliverklee: This should be false, if I'm not mistaken.

Tests/Unit/EmogrifierTest.php
```diff
@@ -2257,4 +2257,29 @@ public function emogrifyRemovesImportantRule($originalStyleAttributeContent, $ex
...
+        $this->subject->setHtml('<html></html>');
+        $this->subject->setCss('@media screen {p^^ {color: red;}}');
+        $this->subject->setDebug(true);
+        $this->subject->emogrify();
+    }
+
+    /**
+     * @test
+     */
+    public function emogrifyKeepsInvalidOrUnrecognizedSelectorsInMediaQueryBlocksInNonDebugMode()
```
oliverklee: Great test names!

Classes/Emogrifier.php
zoliszabo: It may happen that a CSS selector is valid, but is not recognized (not implemented) by Emogrifier. This is why I thought it would be good to keep all selectors we can't figure out whether they are invalid for real or just not (yet) implemented.

Maybe there should be a switch/option (keep unrecognized selectors or not)?

oliverklee: There is no way for Emogrifier to "keep" a selector if it cannot properly match it. Either Emogrifier can match a selector (then it gets converted to an xPath selector, and the CSS will get applied to the HTML), or it cannot match it (then the CSS will not be applied).

zoliszabo: As far as I can understand the code, this keeping/not keeping procedure is applied (only) for media query blocks: if at least one of the selectors from a media query block has matching nodes (is relevant for the document), the whole media query block is kept in the final output. See method `copyCssWithMediaToStyleNode`.

Of course it would make sense to not keep whole media query blocks, but only declarations with relevant selectors, but that's another PR.

oliverklee: I've read this method again: The return value of the method (technically) indicates this:

```php
return $nodesMatchingSelector !== false && $nodesMatchingSelector->length !== 0;
```

This means that it should return true if the selector has found at least one node and has not created any errors. So I think the return value should be false for errors in non-debug mode.

Does this make sense, or am I still missing anything?

zoliszabo: Please consider the following example (as reported in #439, slightly modified):

```
@media only screen and (min-device-width: 768px) and (max-device-width: 1024px) {
    .mobile_link a[href^="tel"] {
        color: blue;
    }
}
```
This would throw a `\RuntimeException` with current master, an `\InvalidArgumentException` in debug mode with this PR. But the CSS selector is actually a correct one, it is just not recognized be Emogrifier yet (see issue #443). So returning `false` would eventually eliminate the above media query block from the end result, which may not be the good decision.
Unfortunately, in case of unrecognized selectors, we do not know whether they are actually invalid or just not implemented, so just to be sure, I think we should trust the input and keep unrecognized CSS selectors in media query blocks, not dump them. This is why a put `return true` in non-debug mode.

oliverklee: Ah, thanks for the explanation. Makes sense.

PR #447, opened 2017-12-28
[CLEANUP] Simplify handleXpathQueryWarnings

oliverklee:


oliverklee:
@zoliszabo Would you like to take a look?

Classes/Emogrifier.php
```diff
@@ -1762,46 +1762,22 @@ private function getNodesToExclude(\DOMXPath $xPath)
...
-                sprintf(
-                    '%1$s in selector >> %2$s << in %3$s on line %4$u',
-                    $message,
-                    $context['selectorToExclude'],
-                    $file,
-                    $line
-                ),
-                1511683740
-            );
+            // warnings generated by invalid/unrecognized selectors in getNodesToExclude
```
zoliszabo: Great simplification!

The only thing to mention is that the comment on this line is not correct. The source of the warning is the `process` method (https://github.com/MyIntervals/emogrifier/blob/e07adc4662359bd977dad6c18dfb314f74365aa9/Classes/Emogrifier.php#L370).

oliverklee: Thanks! Fixed.

PR #461, opened 2018-01-04
[TASK] Prepare the 2.0.0 release

### LINKED ISSUE #421: Release version 2.0.0
I'd like to make the next release version 2.0.0 instead of 1.3.0 as the merged changes (and the pending PRs) could be considered to be breaking.

I'll then also drop compatibility with PHP 5.4 (as announced in the changelog):

@jjriv What do you think about this? Would you be okay with this?

oliverklee:
Closes #421

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 # Emogrifier Change Log
 
 All notable changes to this project will be documented in this file.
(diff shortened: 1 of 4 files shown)


oliverklee:
@jjriv, @zoliszabo, could you please review and merge? Thanks!

PR #463, opened 2018-01-05
[TASK] Test with latest and oldest dependencies on Travis

### LINKED ISSUE #394: Update the composer stuff in Travis
We should make sure that the application works both with the lowest allowed stable version of the dependencies as well as the latest ones. According to Nils Adermann, we also should commit the composer.lock. I'll need to read up on that before doing anything in that direction.

http://naderman.de/slippy/slides/2017-07-13-T3DD17-Composer-Best-Practices.pdf

oliverklee:
This ensures that our dependency versions given in the composer.json always
provide a working set of packages.

Also make the development dependencies more lenient.

Fixes #394

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,21 +2,34 @@ sudo: false
 
 language: php
 
+php:
+- 5.5
+- 5.6
+- 7.0
+- 7.1
+- 7.2
+
 cache:
   directories:
   - vendor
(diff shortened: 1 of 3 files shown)


oliverklee:
@jjriv, @zoliszabo, could you please review and merge? Thanks!

PR #471, opened 2018-01-07
[TASK] Reorder the authors/maintainers in composer.json

oliverklee:
Now the order reflects the current activity level.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,14 @@
     "license": "MIT",
     "authors": [
         {
+            "name": "Oliver Klee",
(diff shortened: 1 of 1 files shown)


oliverklee:
@jjriv Should we keep Roman, Carmeron and Jaime in the composer.json, or should we remove them? They haven't been active for the project for quite some time.

jjriv:
I'd like to keep Cameron and Jaime because they were the original authors. But we can remove Roman.

oliverklee:
> But we can remove Roman.

Done and repushed.

Issue #252, opened 2015-10-12
Invisible node removal: preheaders and responsive design


sanderkruger:
Invisible node removal is nice but it creates problems for 2 cases: preheaders and responsive design.

Preheaders are hidden paragraphs at the start of an email to force a digest of the email to show a certain text. When an email is nicely designed with some links at the top or the subject repeated as a header, this is not the most important content you want recipients to see in the email list of their email client. Most email clients show the first 80-something readable characters of an email.

Typically, you put a `<span class="preheader">blablabla</span>` as the first element in the body and attach display:none, font-size=1px and some other styles to it to make sure it's hidden.

It would be good to be able to exclude certain selectors from invisible node removal.

The second issue is with responsive design. You may want to have 2 alternative representations in the document, by default one is shown and the other is hidden. A media query hides the one and shows the other.

invisible node removal should check media queries to see if a hidden node might be shown under certain circumstances.

One interesting way to handle both cases and open a route to more generic control over emogrifier is to introduce emogrifier classes. The idea is that by applying a class to a DOM element, that element gets special treatment from emogrifier. For example:

```
<span class="emogrifier-show preheader">My preheader</span>
```

Emogrifier would see the class and keep the node, despite being a candidate for invisible node removal. The emogrified result HTML would remove all classes starting with `emogrifier-`.

oliverklee:
Sounds good! :+1:

oliverklee:
@jjriv, @zoliszabo I'd like to completely remove the removal of invisible nodes. Minifying the HTML should not be the job of Emogrifier, but the job of the person/component creating the input HTML in the first place, and removing those nodes is tricky. Thoughts?

zoliszabo:
Since the feature is already implemented, instead of removing it, wouldn't it be better to only change the default behavior to not remove invisible nodes by default?

Secondly, the current option flag name is not OK: it does exactly the opposite to how it is called. https://github.com/MyIntervals/emogrifier/blob/73219f1830190677b9773eef0f019c7e3f964335/Classes/Emogrifier.php#L408-L410 (PR https://github.com/MyIntervals/emogrifier/pull/472)

oliverklee:
I find it very important that we have the courage to remove code or features once it does not provide enough benefit anymore or creates problems which are not easily solvable. Every line of code is a line we need to maintain and test.

[...]

PR #473, opened 2018-01-07
[TASK] Deprecate the removal of invisible nodes

oliverklee:
Relates to #252


oliverklee:
@jjriv @zoliszabo I'd like to have feedback from both of your before this gets merged.

Note: I'm thinking about creating a new class HtmlMinifier that then will provide methods for removing invisible nodes and removing class attributes (both with separate methods).

oliverklee:
Now that we agree on the roadmap, a review and merge would be nice. :smile:

Issue #380, opened 2017-04-18
remove the class="xxx" from html after converting to inline-css


xenDE:
the classes are perfect converted, but the class is still present in the html tag:

`<span class="f3" style="color: #AA5500;">`

this blows up the output size.

A switch like

`private $shouldClassesBeRemoved = true;`

would be fine.

oliverklee:
Or maybe automatically remove all the classes that are not used in the CSS that is kept?

What do you think?

oliverklee:
Removing the CSS classes from HTML will break any media queries that rely on these classes. So I propose we always keep the classes.

@jjriv, @zoliszabo, @xenDE Thoughts?

zoliszabo:
IMHO, if to be implemented, only as an option/switch (which would default to `FALSE`, to keep backwards compatibility) would be reasonable. Automatic removal - without the option to disable it - may break some use-cases. For example when media query blocks are added to the final email HTML _after_ emogrification.

oliverklee:
I was more thinking along the lines of "removing the automatic removal option", i.e., to always keep existing class attributes.

zoliszabo:
@oliverklee, of course. I just cannot make up my mind whether implementing this feature or not would better suit Emogrifier.

Implementing the feature with a `FALSE` default would have the benefit of optionally offering smaller final HTML size, with the drawback of making the Emogrifier class bigger/fatter. (I am not sure what is the general philosophy about adding more and more more-or-less tangent features to Emogrifier.)

oliverklee:
I'd like Emogrifier to be (or become) a tool that does one thing, and does it well, with good, opiniated default behavior. I'd like to have only those options where there are really big benefits for having different behaviors. Having as few options as possible will make usage simpler, and it will keep the code simple and the different test cases few.

oliverklee:
Minimizing the HTML should go into a separate tool (or at least a separate class).

jjriv:
I agree that Emogrifier should be a tool that does one thing. It's definitely better to have fewer options for usability, but that means saying no to some features that may seem desirable by a vocal minority. That said, I agree, let's keep the minimizing of the HTML out of this. 

One argument in favor of this is that other libraries may handle minimization better,. And when those libraries are updated do we also update Emogrifier to have the latest and greatest?  

Another question to ask is this. How much complexity does this feature add? And how much effort does it take to support this feature and keep it working correctly? If this feature is easily broken and a pain to maintain, I say get rid of it. 

[...]

PR #474, opened 2018-01-07
[TASK] Deprecate enableCssToHtmlMapping

oliverklee:
Nowadays, we can expect email clients to understand CSS reasonably well.
So converting CSS non-CSS HTML attributes will not be needed anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Deprecated
 - Support for PHP 5.5 will be removed in Emogrifier 3.0.
 - Support for PHP 5.6 will be removed in Emogrifier 4.0.
+- Converting CSS styles to (non-CSS) HTML attributes will be removed
+  in Emogrifier 3.0. ([#474](https://github.com/jjriv/emogrifier/pull/474))
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


oliverklee:
@jjriv @zoliszabo I'd like to have feedback from both of your before this gets merged.

Note: I'm thinking about creating a new class HtmlMinifier that then will provide methods for removing invisible nodes and removing class attributes (both with separate methods).

zoliszabo:
After today's discussions about invisible node removal (https://github.com/MyIntervals/emogrifier/issues/252), class removal (https://github.com/MyIntervals/emogrifier/issues/380) and seeing this PR, I was having similar thoughts: deprecated or not to be implemented features should not be lost.

My first thought was some kind of a plug-in system, where these tangent features would be implemented as separate plugins and included/activated when needed.
Then I thought that some of these features may have sense to be used even stand-alone and the plug-in system would not make that possible. Instead, other classes with well-defined responsibilities should be worked out.

jjriv:
I am coming to this conversation a little later but I agree with both of you. I like where this is heading. The addition of other classes seems like a good way to handle these extraneous features and focus more on the Emogrifier core.

oliverklee:
Now that we agree on the roadmap, a review and merge would be nice. :smile:

PR #486, opened 2018-01-12
[FEATURE] Add an AbstractHtmlProcessor class

oliverklee:
This class will be the base class for other classes that can remove
invisible nodes, convert CSS to non-CSS attributes etc.

The folder structure will be cleaned up once the Emogrifier class has
been renamed to CssInliner and moved from the Pelago\ namespace to the
Pelage\Emogrifier namespace (which will happen in Emogrifier 4.0.0).

### DIFF
--- a/Classes/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
+++ b/Classes/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
@@ -1,0 +1,157 @@
+<?php
+
+namespace Pelago\Emogrifier\HtmlProcessor;
+
+/**
+ * Base class for HTML processor that e.g., can remove, add or modify nodes or attributes.
+ *
+ * @author Oliver Klee <github@oliverklee.de>
+ */
+abstract class AbstractHtmlProcessor
+{
+    /**
+     * @var string
+     */
+    const DEFAULT_DOCUMENT_TYPE = '<!DOCTYPE html>';
+
+    /**
+     * @var string
+     */
+    const CONTENT_TYPE_META_TAG = '<meta http-equiv="Content-Type" content="text/html; charset=utf-8">';
+
+    /**
+     * @var \DOMDocument
+     */
+    protected $xmlDocument = null;
+
+    /**
+     * @param string $unprocessedHtml raw HTML, will get heavily normalized
+     *
+     * @throws \InvalidArgumentException if $unprocessedHtml is anything other than a non-empty string
+     */
+    public function __construct($unprocessedHtml)
+    {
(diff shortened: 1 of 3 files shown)


jjriv:
I'm going to let @zoliszabo do the review on this one. I'm not familiar enough with the code to approve it.

oliverklee:
Note concerning the review: Most of this code is copied from the Emogrifier class and its test case.

Classes/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -0,0 +1,157 @@
...
+
+        $this->xmlDocument = $this->createXmlDocument($unprocessedHtml);
+    }
+
+    /**
+     * Renders the normalized and processed HTML.
+     *
+     * @return string
+     */
+    public function render()
```
zoliszabo: As far as I can conclude, child classes of this abstract class will need to override public method `render()` and call `parent::render()` in each implementation.

I would introduce a protected method `process()`, abstract in this base class, which would be responsible for holding the variable logic of each subclass (I think this is called the "template design pattern", but it's not that important):

(And maybe even declare the `render()` method `final`.)

```php
abstract protected function process();

...

final public function render()
{
    $this->process();
    return $this->xmlDocument->saveHTML();
}
```

What do you think?

Classes/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * @var string
+     */
+    const CONTENT_TYPE_META_TAG = '<meta http-equiv="Content-Type" content="text/html; charset=utf-8">';
+
+    /**
+     * @var \DOMDocument
+     */
+    private $xmlDocument = null;
```
zoliszabo: I think this property should be protected instead of private, so child classes have access to it. Or getter method must be implemented.

oliverklee: Will do.

oliverklee:
Made the property protected and repushed.

Classes/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
zoliszabo: You are right, it makes sense. For some reason, I was under the impression that the `convertCssToAttributes()`, `removeClassAttributes()` etc. methods will only set flags for what there is to do and then the `render()` method will do the actual job.

Issue #489, opened 2018-01-15
Change the directory structure to match the PHP PDS


oliverklee:
http://php-pds.com/
https://github.com/php-pds/skeleton_research

Let's do this one (top-level) directory per PR at a time.

PR #494, opened 2018-01-21
[CLEANUP] PHP-PDS directory structure: Tests -> tests

zoliszabo:
Renamed "Tests" top-level directory to "tests" to match PHP-PDS recommendation.
https://github.com/MyIntervals/emogrifier/issues/489


oliverklee:
For the last remaining directory, I think we can drop the "PhpCodeSniffer/Standards/Emogrifier/" segments and rename the file to PhpCodeSniffer.xml.

PR #496, opened 2018-01-21
[ENHANCEMENT] Use composer script to run unit tests

zoliszabo:
Command is defined in a single place (composer.json) and if it ever needs
to be modified, it has to be modified in one place.
To run PHPUnit tests simply use command `composer test`.

(Next: composer command for phpcs too.)


oliverklee:
Thanks!

In the commit message, I've changed [ENHANCEMENT] to [FEATURE] so the code is consistent with the other commits.

PR #501, opened 2018-01-27
[TASK] Make the Composer CI sniffs more future-proof

oliverklee:
The script names should be named so that we can add integration tests
and configuration file linting while keeping consistent script names
without having to rename the existing scripts.


oliverklee:
(And this also will make the composer script names consistent to the names I use in my other projects.)

PR #502, opened 2018-02-06
[BUGFIX] Selector precedence pseudos/attributes

JakeQZ:
Corrected selector precedence calculation for selectors involving
pseudo-classes and/or attributes (ref:
https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity).

Increased multiplier for selector type to allow up to 100 of each type.

Added PHPUnit tests for selector precedence.  Moved common testing HTML to a
constant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
+- Calculation of selector precedence for selectors involving pseudo-classes
+  and/or attributes.
 
 
 ## 2.0.0
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -1404,18 +1404,24 @@ class Emogrifier
         $selectorKey = md5($selector);
         if (!isset($this->caches[static::CACHE_KEY_SELECTOR][$selectorKey])) {
             $precedence = 0;
-            $value = 100;
-            // ids: worth 100, classes: worth 10, elements: worth 1
-            $search = ['\\#', '\\.', ''];
+            $value = 10000;
+            $search = [
+                // ids: worth 10000
+                '\\#',
+                // classes, attributes, pseudo-classes (not pseudo-elements) except `:not`: worth 100
+                '(?:\\.|\\[|(?<!:):(?!not\\())',
+                // elements (not attribute values or `:not`), pseudo-elements: worth 1
+                '(?:(?<![="\':\\w-])|::)'
+            ];
 
             foreach ($search as $s) {
                 if (trim($selector) === '') {
                     break;
                 }
                 $number = 0;
-                $selector = preg_replace('/' . $s . '\\w+/', '', $selector, -1, $number);
+                $selector = preg_replace('/' . $s . '[\\w-]++/', '', $selector, -1, $number);
                 $precedence += ($value * $number);
-                $value /= 10;
+                $value /= 100;
             }
             $this->caches[static::CACHE_KEY_SELECTOR][$selectorKey] = $precedence;
         }
--- a/tests/Unit/EmogrifierTest.php
+++ b/tests/Unit/EmogrifierTest.php
@@ -18,6 +18,22 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
     const LF = "\n";
 
     /**
+     * @var string Common HTML markup with a variety of elements and attributes for testing with
+     */
+    const COMMON_TEST_HTML = '
+        <html>
+            <body>
+                <p class="p-1"><span>some text</span></p>
+                <p class="p-2"><span title="bonjour">some</span> text</p>
+                <p class="p-3"><span title="buenas dias">some</span> more text</p>
+                <p class="p-4" id="p4"><span title="avez-vous">some</span> more text</p>
+                <p class="p-5 additional-class"><span title="buenas dias bom dia">some</span> more text</p>
+                <p class="p-6"><span title="title: subtitle; author">some</span> more text</p>
+            </body>
+        </html>
+    ';
+
+    /**
      * @var string
      */
     private $html5DocumentType = '<!DOCTYPE html>';
@@ -572,19 +588,7 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
     {
         $cssDeclaration1 = 'color: red;';
         $cssDeclaration2 = 'text-align: left;';
-        $html = '
-            <html>
-                <body>
-                    <p class="p-1"><span>some text</span></p>
-                    <p class="p-2"><span title="bonjour">some</span> text</p>
-                    <p class="p-3"><span title="buenas dias">some</span> more text</p>
-                    <p class="p-4" id="p4"><span title="avez-vous">some</span> more text</p>
-                    <p class="p-5 additional-class"><span title="buenas dias bom dia">some</span> more text</p>
-                    <p class="p-6"><span title="title: subtitle; author">some</span> more text</p>
-                </body>
-            </html>
-            ';
-        $this->subject->setHtml($html);
+        $this->subject->setHtml(static::COMMON_TEST_HTML);
         $this->subject->setCss(sprintf($css, $cssDeclaration1, $cssDeclaration2));
 
         $result = $this->subject->emogrify();
@@ -672,19 +676,93 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
     {
         $cssDeclaration1 = 'color: red;';
         $cssDeclaration2 = 'text-align: left;';
-        $html = '
-            <html>
-                <body>
-                    <p class="p-1"><span>some text</span></p>
-                    <p class="p-2"><span title="bonjour">some</span> text</p>
-                    <p class="p-3"><span title="buenas dias">some</span> more text</p>
-                    <p class="p-4" id="p4"><span title="avez-vous">some</span> more text</p>
-                    <p class="p-5 additional-class"><span title="buenas dias bom dia">some</span> more text</p>
-                    <p class="p-6"><span title="title: subtitle; author">some</span> more text</p>
-                </body>
-            </html>
-            ';
-        $this->subject->setHtml($html);
+        $this->subject->setHtml(static::COMMON_TEST_HTML);
+        $this->subject->setCss(sprintf($css, $cssDeclaration1, $cssDeclaration2));
+
+        $result = $this->subject->emogrify();
+
+        static::assertContains(sprintf($expectedHtml, $cssDeclaration1, $cssDeclaration2), $result);
+    }
+
+    /**
+     * Data provider for {@see EmogrifierTest::emogrifyAppliesMostSpecificSelectorCssToMatchingElements}
+     *
+     * @return string[][]
+     */
+    public function selectorSpecificityCssDataProvider()
+    {
+        /**
+         * @var string[][] $selectorData Data for constructing the parameter sets, whose keys are a test description,
+         *      and values are an array with the following keys:
+         *      0 - first selector expression
+         *      1 - second selector expression
+         *      2 - operator, '<', '==' or '>', defining whether the first selector should respectively have lower,
+         *          equal or higher specificity than the second
+         *      3 - tag (in {@see EmogrifierTest::COMMON_TEST_HTML}) expected to be matched by both selectors, without
+         *          the closing '>', e.g. '<p class="p=1"'
+         */
+        $selectorData = [
+            'class more specific than types' => ['.p-1', 'html body p', '>', '<p class="p-1"'],
+            'id more specific than class and types' => ['#p4', 'html body .p-4', '>', '<p class="p-4" id="p4"'],
+            'id more specific than 20 classes and types'
+                => ['#p4', 'html body ' . str_repeat('.p-4', 20), '>', '<p class="p-4" id="p4"'],
+            'psuedo-class as specific as class' => ['.p-1', '*:first-child', '==', '<p class="p-1"'],
+            'attribute as specific as class' => ['span[title="bonjour"]', '.p-2 span', '==', '<span title="bonjour"'],
+            ':not alone does not increase specificity' => ['p:not(* + *)', 'p', '==', '<p class="p-1"'],
+            ':not with type more specific than nothing' => ['.p-1:not(h1)', '.p-1', '>', '<p class="p-1"'],
+            ':not with type as specific as type' => ['*:not(h1)', 'p', '==', '<p class="p-1"'],
+            ':not with several types more specific than single type'
+                => ['*:not(html body h1)', 'p', '>', '<p class="p-1"'],
+            ':not with class more specific than types'
+                => ['*:not(.p-2)', 'html body p', '>', '<p class="p-1"'],
+            ':not with class more specific than :not with types'
+                => ['*:not(.p-2)', '*:not(html body h1)', '>', '<p class="p-1"'],
+            'class more specific than :not with types' => ['.p-1', '*:not(html body h1)', '>', '<p class="p-1"'],
+        ];
+
+        /**
+         * @var string[][] $parameterSets Sets of parameters returned by this data provider for
+         *      {@see EmogrifierTest::emogrifyAppliesMostSpecificSelectorCssToMatchingElements}.
+         *      The sprintf placeholders %1$s and %2$s will automatically be replaced with CSS declarations
+         *      like 'color: red;' or 'color: blue;'.
+         */
+        $parameterSets = [];
+        foreach ($selectorData as $name => $data) {
+            $firstRuleBlock = $data[0] . ' { %1$s }';
+            $secondRuleBlock = $data[1] . ' { %2$s }';
+            $firstRuleBlockWasApplied = $data[3] . ' style="%1$s">';
+            $secondRuleBlockWasApplied = $data[3] . ' style="%2$s">';
+
+            // Check selectors actually match on their own (i.e. test is valid)
+            $parameterSets[$name . ' (1st selector valid for test)'] = [$firstRuleBlock, $firstRuleBlockWasApplied];
+            $parameterSets[$name . ' (2nd selector valid for test)'] = [$secondRuleBlock, $secondRuleBlockWasApplied];
+
+            // Check first selector matches when its rule is last unless the second has higher specificity
+            $parameterSets[$name . ' (1st selector in last rule)'] = [
+                $secondRuleBlock . ' ' . $firstRuleBlock,
+                $data[2] === '<' ? $secondRuleBlockWasApplied : $firstRuleBlockWasApplied
+            ];
+
+            // Check second selector matches when its rule is last unless the first has higher specificity
+            $parameterSets[$name . ' (2nd selector in last rule)'] = [
+                $firstRuleBlock . ' ' . $secondRuleBlock,
+                $data[2] === '>' ? $firstRuleBlockWasApplied : $secondRuleBlockWasApplied
+            ];
+        }
+        return $parameterSets;
+    }
+
+    /**
+     * @test
+     * @param string $css CSS statements, potentially with %1$s and $2$s placeholders for a CSS declaration
+     * @param string $expectedHtml HTML, potentially with %1$s and $2$s placeholders for a CSS declaration
+     * @dataProvider selectorSpecificityCssDataProvider
+     */
+    public function emogrifyAppliesMostSpecificSelectorCssToMatchingElements($css, $expectedHtml)
+    {
+        $cssDeclaration1 = 'color: red;';
+        $cssDeclaration2 = 'color: blue;';
+        $this->subject->setHtml(static::COMMON_TEST_HTML);
         $this->subject->setCss(sprintf($css, $cssDeclaration1, $cssDeclaration2));
 
         $result = $this->subject->emogrify();


oliverklee:
Hi @JakeQZ, thanks for the PR! :heart: 

Would you be willing to create another PR for the testing HTMLconstant? Then we can have the cleanup nicely separated from the actual bugfix, which will help keep the Git history (and line-based history) clean and would allow for separate reverts (should we every need to revert any of those commits)?

tests/Unit/EmogrifierTest.php
```diff
@@ -672,19 +676,93 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+         *          equal or higher specificity than the second
+         *      3 - tag (in {@see EmogrifierTest::COMMON_TEST_HTML}) expected to be matched by both selectors, without
+         *          the closing '>', e.g. '<p class="p=1"'
+         */
+        $selectorData = [
+            'class more specific than types' => ['.p-1', 'html body p', '>', '<p class="p-1"'],
+            'id more specific than class and types' => ['#p4', 'html body .p-4', '>', '<p class="p-4" id="p4"'],
+            'id more specific than 20 classes and types'
+                => ['#p4', 'html body ' . str_repeat('.p-4', 20), '>', '<p class="p-4" id="p4"'],
+            'psuedo-class as specific as class' => ['.p-1', '*:first-child', '==', '<p class="p-1"'],
```
oliverklee: typo: psuedo -> pseudo

JakeQZ:
Ok, have created new PR for the testing HTML constant.

tests/Unit/EmogrifierTest.php
JakeQZ: Have corrected typo.

JakeQZ:
Have corrected typo, and merged the now separate HTML constant cleanup and changed the opening comment of this PR to reflect that it doesn't include that.

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
+- Calculation of selector precedence for selectors involving pseudo-classes
+  and/or attributes.
```
oliverklee: Could you please add a link to the PR (so this is consistent with the other changelog entries)? Then we don't need to edit your PR before we merge it. Thanks!

src/Emogrifier.php
```diff
@@ -1404,18 +1404,24 @@ private function getCssSelectorPrecedence($selector)
...
-            $search = ['\\#', '\\.', ''];
+            $value = 10000;
+            $search = [
+                // ids: worth 10000
+                '\\#',
+                // classes, attributes, pseudo-classes (not pseudo-elements) except `:not`: worth 100
+                '(?:\\.|\\[|(?<!:):(?!not\\())',
+                // elements (not attribute values or `:not`), pseudo-elements: worth 1
+                '(?:(?<![="\':\\w-])|::)'
+            ];
```
oliverklee: I propose putting the values (1, 100, 10000) in code, e.g. as array keys. Then this array does not depend on the order of the entries.

src/Emogrifier.php
```diff
@@ -1404,18 +1404,24 @@ private function getCssSelectorPrecedence($selector)
...
 
             foreach ($search as $s) {
                 if (trim($selector) === '') {
                     break;
                 }
                 $number = 0;
-                $selector = preg_replace('/' . $s . '\\w+/', '', $selector, -1, $number);
+                $selector = preg_replace('/' . $s . '[\\w-]++/', '', $selector, -1, $number);
```
oliverklee: Is this related to this PR? And if it is not, could you please undo this change?

And if this actually fixes a but, I'd prefer this change to have its own PR (including changelog entry and unit tests).

tests/Unit/EmogrifierTest.php
```diff
@@ -684,6 +684,92 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
         static::assertContains(sprintf($expectedHtml, $cssDeclaration1, $cssDeclaration2), $result);
     }
 
+    /**
+     * Data provider for {@see EmogrifierTest::emogrifyAppliesMostSpecificSelectorCssToMatchingElements}
```
oliverklee: I don't think we need the description and the `@see` here as the method name is "speaking" enough.

tests/Unit/EmogrifierTest.php
```diff
@@ -684,6 +684,92 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+     *
+     * @return string[][]
```
oliverklee: Good type!

tests/Unit/EmogrifierTest.php
```diff
@@ -684,6 +684,92 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+            $parameterSets[$name . ' (1st selector in last rule)'] = [
+                $secondRuleBlock . ' ' . $firstRuleBlock,
+                $data[2] === '<' ? $secondRuleBlockWasApplied : $firstRuleBlockWasApplied
+            ];
+
+            // Check second selector matches when its rule is last unless the first has higher specificity
+            $parameterSets[$name . ' (2nd selector in last rule)'] = [
+                $firstRuleBlock . ' ' . $secondRuleBlock,
+                $data[2] === '>' ? $firstRuleBlockWasApplied : $secondRuleBlockWasApplied
+            ];
```
oliverklee: Instead of having code blocks with comments explaining what they do, please move these code blocks to private methods with "speaking" names. This will make it easier to get an overview of what happens here.

tests/Unit/EmogrifierTest.php
```diff
@@ -684,6 +684,92 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+            ':not alone does not increase specificity' => ['p:not(* + *)', 'p', '==', '<p class="p-1"'],
+            ':not with type more specific than nothing' => ['.p-1:not(h1)', '.p-1', '>', '<p class="p-1"'],
+            ':not with type as specific as type' => ['*:not(h1)', 'p', '==', '<p class="p-1"'],
+            ':not with several types more specific than single type'
+                => ['*:not(html body h1)', 'p', '>', '<p class="p-1"'],
+            ':not with class more specific than types'
+                => ['*:not(.p-2)', 'html body p', '>', '<p class="p-1"'],
+            ':not with class more specific than :not with types'
+                => ['*:not(.p-2)', '*:not(html body h1)', '>', '<p class="p-1"'],
+            'class more specific than :not with types' => ['.p-1', '*:not(html body h1)', '>', '<p class="p-1"'],
```
oliverklee: Great test case descriptions!

tests/Unit/EmogrifierTest.php
```diff
@@ -684,6 +684,92 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+
+        /**
+         * @var string[][] $parameterSets Sets of parameters returned by this data provider for
+         *      {@see EmogrifierTest::emogrifyAppliesMostSpecificSelectorCssToMatchingElements}.
+         *      The sprintf placeholders %1$s and %2$s will automatically be replaced with CSS declarations
+         *      like 'color: red;' or 'color: blue;'.
+         */
+        $parameterSets = [];
+        foreach ($selectorData as $name => $data) {
+            $firstRuleBlock = $data[0] . ' { %1$s }';
```
oliverklee: Maybe also move the code in the loop into a well-named method.

oliverklee:
(Just removed some comments that don't apply anymore so we keep this PR tidy.)

CHANGELOG.md
JakeQZ: Done

src/Emogrifier.php
JakeQZ: I've put the values as array values, so that it would be possible in future to add additional regular expressions for the same value if desired.  There is still some dependency on the order of the entries, as the matcher for element type relies on the ID and class matches having been removed.

In fact, it could still match after a space in an attribute value expression, such as `[title="buenas dias"]` - maybe this should also be addressed while we're here...?

src/Emogrifier.php
JakeQZ: I've reverted this change, as none of the tests I added fail without it.  When removing matched parts, I wanted to see `:first-child` rather than just `:first` removed, but maybe that isn't necessary.  I'll see if I can come up with a valid test case that would require it...

tests/Unit/EmogrifierTest.php
JakeQZ: Removed

tests/Unit/EmogrifierTest.php
JakeQZ: Refactored loop body into new private method `getSelectorSpecificityCssDataSets`

tests/Unit/EmogrifierTest.php
JakeQZ: Refactored into new private methods `getMatchedCssDataSet` and `getSelectorSpecificityCssDataSet`

JakeQZ:
I'm updating the branch with the requested changes, but will also need to rebase because of CHANGELOG.md.  I've responded to some of the individual comments with more detail.

I've also spotted that, I think, a space in an attribute value e.g. `[title="buenas dias"]` will still result in an additional match as an element type (`dias`) - maybe this should also be addressed while we're here...?

JakeQZ:
Have rebased and merged, and confirmed that `[title="buenas dias"]` is being incorrectly evaluated as more specific becuase of the space...

oliverklee:
> I've also spotted that, I think, a space in an attribute value e.g. [title="buenas dias"] will still result in an additional match as an element type (dias) - maybe this should also be addressed while we're here...?

Please let's fix this in a separate PR.

tests/Unit/EmogrifierTest.php
```diff
@@ -690,6 +690,143 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+        // The sprintf placeholders %1$s and %2$s will automatically be replaced with CSS declarations
+        // like 'color: red;' or 'color: blue;'.
+        return [$selector . ' { %1$s }', $matchedTagPart . ' style="%1$s">'];
+    }
+
+    /**
+     * @param string $firstSelector A selector expression to be used in an earlier CSS declaration
+     * @param string $secondSelector Some other selector expression to be used in a later CSS declaration
+     * @param bool $firstSelectorIsMoreSpecific Whether the first selector's rules are expected to prevail
+     * @param string $matchedTagPart Tag (in {@see EmogrifierTest::COMMON_TEST_HTML}) expected to be matched by both
```
oliverklee: I don't think we need the `@see` here.

tests/Unit/EmogrifierTest.php
```diff
@@ -690,6 +690,143 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
         static::assertContains(sprintf($expectedHtml, $cssDeclaration1, $cssDeclaration2), $result);
     }
 
+    /**
+     * @param string $selector A selector expression
+     * @param string $matchedTagPart Tag (in {@see EmogrifierTest::COMMON_TEST_HTML}) expected to be matched by the
```
oliverklee: I don't think we need the `@see` here.

tests/Unit/EmogrifierTest.php
```diff
@@ -690,6 +690,143 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+     * @param bool $firstSelectorIsMoreSpecific Whether the first selector's rules are expected to prevail
+     * @param string $matchedTagPart Tag (in {@see EmogrifierTest::COMMON_TEST_HTML}) expected to be matched by both
+     *                               selectors, without the closing '>', e.g. '<p class="p-1"'
+     *
+     * @return string[][]
+     */
+    private function getSelectorSpecificityCssDataSet(
+        $firstSelector,
+        $secondSelector,
+        $firstSelectorIsMoreSpecific,
```
oliverklee: Instead of having this "toggle" parameter (that toggles between two different behaviors), we should have two separate, well-named methods.

tests/Unit/EmogrifierTest.php
```diff
@@ -690,6 +690,143 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+    }
+
+    /**
+     * @return string[][]
+     */
+    public function selectorSpecificityCssDataProvider()
+    {
+        $selectorData = [
+            ['class more specific than types', '.p-1', 'html body p', '>', '<p class="p-1"'],
+            ['id more specific than class and types', '#p4', 'html body .p-4', '>', '<p class="p-4" id="p4"'],
```
oliverklee: id -> ID (in human-readable text, not in code)

tests/Unit/EmogrifierTest.php
```diff
@@ -690,6 +690,143 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+                '*:not(html body h1)',
+                '>',
+                '<p class="p-1"'
+            ],
+            ['class more specific than :not with types', '.p-1', '*:not(html body h1)', '>', '<p class="p-1"'],
+        ];
+
+        $datasets = [];
+        foreach ($selectorData as $data) {
+            $datasets += call_user_func_array([$this, 'getSelectorSpecificityCssDataSets'], $data);
```
oliverklee: Why not call $this->getSelectorSpecificityCssDataSets directly?

tests/Unit/EmogrifierTest.php
```diff
@@ -690,6 +690,143 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+                '*:not(.p-2)',
+                '*:not(html body h1)',
+                '>',
+                '<p class="p-1"'
+            ],
+            ['class more specific than :not with types', '.p-1', '*:not(html body h1)', '>', '<p class="p-1"'],
+        ];
+
+        $datasets = [];
+        foreach ($selectorData as $data) {
```
oliverklee: I think we need different variable names here because currently both "data" and "selectorData" basically mean the same thing, but the names should make the difference clear.

tests/Unit/EmogrifierTest.php
```diff
@@ -690,6 +690,143 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+                $secondSelector,
+                $firstSelector,
+                $operator === '<',
+                $matchedTagPart
+            ),
+            $name . ' (2nd selector in last rule)' => $this->getSelectorSpecificityCssDataSet(
+                $firstSelector,
+                $secondSelector,
+                $operator === '>',
+                $matchedTagPart
```
oliverklee: We can simplify this by having 2 or 3 separate tests (with shared or separate data providers): one for selectors that have the same specificity (what a word!), and one or two for different specificities (depending on whether you think that we order matters).

What do you think?

src/Emogrifier.php
```diff
@@ -1404,18 +1404,26 @@ private function getCssSelectorPrecedence($selector)
...
-            // ids: worth 100, classes: worth 10, elements: worth 1
-            $search = ['\\#', '\\.', ''];
-
-            foreach ($search as $s) {
+            /**
+             * @var int[] $search Keys are a regular expression part to match before a CSS name,
+             *                    values are a multiplier factor per match to weight specificity
+             */
+            $search = [
+                // ids: worth 10000
```
oliverklee: ids -> IDs

JakeQZ:
> We can simplify this by having 2 or 3 separate tests (with shared or separate data providers): one for selectors that have the same specificity (what a word!), and one or two for different specificities (depending on whether you think that we order matters).
> 
> What do you think?

I think doing that would also resolve the other issues with the testing code you commented on.  So I will refactor with that approach...

tests/Unit/EmogrifierTest.php
JakeQZ: Have refactored along those lines.

tests/Unit/EmogrifierTest.php
JakeQZ: Changed

tests/Unit/EmogrifierTest.php
JakeQZ: No longer applicable after refactoring

tests/Unit/EmogrifierTest.php
JakeQZ: No longer applicable after refactoring

src/Emogrifier.php
JakeQZ: Changed

JakeQZ:
I've refactored as suggested, and changed "id" to "ID" in comments/dataset names.  This has also reduced the amount of code a bit, and the number of distinct tests (pairs of tests now combined, and checking the validity of the dataset parameters was only necessary in 1/4 cases).

src/Emogrifier.php
```diff
@@ -1404,18 +1404,26 @@ private function getCssSelectorPrecedence($selector)
...
-            $value = 100;
-            // ids: worth 100, classes: worth 10, elements: worth 1
-            $search = ['\\#', '\\.', ''];
-
-            foreach ($search as $s) {
+            /**
+             * @var int[] $search Keys are a regular expression part to match before a CSS name,
+             *                    values are a multiplier factor per match to weight specificity
+             */
+            $search = [
```
oliverklee: This now even can be a constant instead of a local variable.

oliverklee: (which would make the method more concise and move fixed data out of it)

JakeQZ: Have moved it to a constant.

src/Emogrifier.php
```diff
@@ -1404,18 +1420,13 @@ private function getCssSelectorPrecedence($selector)
         $selectorKey = md5($selector);
         if (!isset($this->caches[static::CACHE_KEY_SELECTOR][$selectorKey])) {
             $precedence = 0;
-            $value = 100;
-            // ids: worth 100, classes: worth 10, elements: worth 1
-            $search = ['\\#', '\\.', ''];
-
-            foreach ($search as $s) {
+            foreach ($this->selectorPrecedenceMatchers as $s => $value) {
```
oliverklee: While we're touching the foreach anyway, please let's rename `$s` to something more expressive, e.g. `$matcher`.

tests/Unit/EmogrifierTest.php
```diff
@@ -690,6 +690,118 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
 
+    /**
+     * @return string[][]
+     */
+    public function differentCssSelectorSpecificityDataProvider()
+    {
+        return [
+            'class more specific than types' => ['html body p', '.p-1', '<p class="p-1"'],
+            'ID more specific than class and types' => ['html body .p-4', '#p4', '<p class="p-4" id="p4"'],
+            'ID more specific than 20 classes and types' => [
```
oliverklee: Why exactly 20? Could this also be a different number? Maybe a comment would be helpful why this number is here.

tests/Unit/EmogrifierTest.php
```diff
@@ -690,6 +690,118 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+     * @test Confirms the dataset is valid for the test, and the less specific selector does in fact match
+     * @param string $lessSpecificSelector A selector expression
+     * @param string $moreSpecificSelector Unused parameter from dataset
+     * @param string $matchedTagPart Tag expected to be matched by selector, without the closing '>',
+     *                               e.g. '<p class="p-1"'
+     * @dataProvider differentCssSelectorSpecificityDataProvider
+     */
+    public function emogrifyWouldApplyLessSpecificCssSelectorToMatchingElements( // phpcs:ignore
+        $lessSpecificSelector,
+        // PHP_CodeSniffer warning ignored: The method parameter $moreSpecificSelector is never used
```
oliverklee: You can switch the order of parameters and then make this test method have one argument less than the others (having surplus parameters in a method call is fine in PHP).

I propose this order:
`$matchedTagPart, $lessSpecificSelector, $moreSpecificSelector`

tests/Unit/EmogrifierTest.php
```diff
@@ -690,6 +690,118 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+        $lessSpecificSelector,
+        $moreSpecificSelector,
+        $matchedTagPart
+    ) {
+        $this->subject->setHtml(static::COMMON_TEST_HTML);
+        $this->subject->setCss(
+            $lessSpecificSelector . ' { color: red } ' .
+            $moreSpecificSelector . ' { color: green } ' .
+            $moreSpecificSelector . ' { background-color: green } ' .
+            $lessSpecificSelector . ' { background-color: red }'
```
oliverklee: Any particular reason why you're leaving out the closing semicolon (`color: red;`)?

src/Emogrifier.php
JakeQZ: Done

tests/Unit/EmogrifierTest.php
JakeQZ: 20 was arbitrary.  Have bumped this to 99 which is the maximum supported (without rewriting the comparator, and I don't think likely to be a real-world limitation).  Also added a comment suggesting a similar test for classes vs 99 types, which would require support for chaining pseudo-classes, e.g. `:not(h1):not(h1)...`

tests/Unit/EmogrifierTest.php
JakeQZ: No particular reason.  Have added it.

tests/Unit/EmogrifierTest.php
JakeQZ: Good thinking (it took some coaxing to get CodeSniffer to suppress the warning without introducing a different one!).  Agree with your proposed order as being most logical and have changed this and the other methods for consistency.

JakeQZ:
Have made the suggested changes and rebased.  More detail in reply to individual comments.

tests/Unit/EmogrifierTest.php
```diff
@@ -737,6 +737,117 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
     }
 
+    /**
+     * @return string[][]
+     */
+    public function differentCssSelectorSpecificityDataProvider()
+    {
+        return [
+            'class more specific than types' => ['<p class="p-1"', 'html body p', '.p-1'],
+            'ID more specific than class and types' => ['<p class="p-4" id="p4"', 'html body .p-4', '#p4'],
```
oliverklee: Do we need two types here, or would one suffice? (Test cases should be minimal so only relevant things are there, making it easier to find out what's important when reading the test.)

tests/Unit/EmogrifierTest.php
```diff
@@ -737,6 +737,117 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+    }
+
+    /**
+     * @return string[][]
+     */
+    public function equalCssSelectorSpecificityDataProvider()
+    {
+        return [
+            'pseudo-class as specific as class' => ['<p class="p-1"', '.p-1', '*:first-child'],
+            'attribute as specific as class' => ['<span title="bonjour"', 'span[title="bonjour"]', '.p-2 span'],
```
oliverklee: Should we also have a case comparing a pseudo-class comparing an attribute, or would that be redundant?

oliverklee:
> I'll go through them with a finer-toothed comb - probably won't be until early next week now...

Okay, I'll do the review then.

Could you also please rebase?

JakeQZ:
I've reworked the test data as planned, and rebased.

JakeQZ:
> I've reworked the test data as planned, and rebased.

Actually I think the test data for selector specificity inequality/ordering can be simplified - if all selectors can target the same element, then it can be generated from an array of selectors of increasing specificity, avoiding repetition.

Please hold off reviewing this for now, and I'll either commit another change, or advise that it can't be simplified in this way.

JakeQZ:
> Actually I think the test data for selector specificity inequality/ordering can be simplified - if all selectors can target the same element, then it can be generated from an array of selectors of increasing specificity, avoiding repetition.

I have now changed this as described; the presentation of the test data in the code should now be clearer.

tests/Unit/EmogrifierTest.php
```diff
@@ -737,6 +737,198 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+     * Specificity ordering for selectors involving pseudo-classes, attributes and `:not` is covered through the
+     * combination of these tests and the equal specificity tests and thus does not require explicit separate testing.
+     *
+     * @return string[][]
+     */
+    public function differentCssSelectorSpecificityDataProvider()
+    {
+        // broken: testing selectors with 99 types or 99 IDs requires support for chaining `:not(h1):not(h1)...` or
+        // `#p4#p4...`, so currently 2 types and/or 2 IDs are tested as 'many' instead of 99 of each.
+        $classP4Repeated = str_repeat('.p-4', 99);
```
oliverklee: By the way, please let's reduce the 99 to the minimal number that's relevant for the effect. My educated guess is that it's 2 or 3 (as this is not a performance stress test).

JakeQZ: I've changed this from 99 to 2 in these tests, as testing that 'many' is more specific than one is sufficient.

To confirm the increase in selector type weighting I've added a single separate dataset with 99.

JakeQZ:
I've changed 99 to 2 in the selector inequality test data for confirming 'many' is more specific than 'one', and added a single separate test with 99 classes to confirm the increase in selector weighting multiplier.

tests/Unit/EmogrifierTest.php
```diff
@@ -737,6 +737,204 @@ public function emogrifyNotAppliesCssToNonMatchingElements($css, $expectedHtml)
...
+    }
+
+    /**
+     * @test Confirms the dataset is valid for the test, and the less specific selector does in fact match
+     * @param string $matchedTagPart Tag expected to be matched by selector, without the closing '>',
+     *                               e.g. '<p class="p-1"'
+     * @param string $lessSpecificSelector A selector expression
+     * @dataProvider differentCssSelectorSpecificityDataProvider
+     */
+    public function emogrifyWouldApplyLessSpecificCssSelectorToMatchingElements(
```
oliverklee: The way I read this test is that is doesn't really check for less/more specific selectors, but tests that the given selectors work (without regard to specificity), i.e., we can remove this test as we already have tests for the individual selectors matching. Or am I missing anything here?

JakeQZ: Yes, this was just to confirm that the test datasets did in fact represent valid tests, and the lower precedence selector provided would in fact match the intended target if used alone.  However, since the test data is now written such that the lower precedence selector will be the same as the higher precedence one from the previous dataset, this is now unnecessary apart from the first dataset - but that uses the universal selector which obviously matches - and the final separate dataset with 99 classes - though for just a single dataset this could be confirmed 'by inspection'.

So I have removed this.

PR #504, opened 2018-02-06
[BUGFIX] CSS property values with newlines dropped

JakeQZ:
Added the 's' (PCRE_DOTALL) pattern modifier for `preg_match` in
`parseCssDeclarationsBlock` so that property values containing newlines are also
matched.  Added PHPUnit tests for such property values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
-- CSS rules with property values containing newlines are dropped.
+- Allow CSS property values containing newlines
+  ([#504](https://github.com/MyIntervals/emogrifier/pull/504))
 
 
 ## 2.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
+- CSS rules with property values containing newlines are dropped.
 
 
 ## 2.0.0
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
+- CSS rules with property values containing newlines are dropped.
```
oliverklee: Could you also please link the PR (as in the other changelog entries)? Then we don't need to edit the changelog to merge this PR.

oliverklee: Could you please reword this so it's about what's fixed now, for example like this (also in the commit message):

Allow property values with newlines

tests/Unit/EmogrifierTest.php
```diff
@@ -758,6 +758,20 @@ public function formattedCssDeclarationDataProvider()
...
+                    '1px 1px 1px #000;',
+                'text-shadow: 1px 1px 3px #000,' . static::LF .
+                    '1px 1px 1px #000;',
+            ],
+            'one declaration with Windows line ending in property value' => [
+                'text-shadow:' . "\r\n" .
+                    '1px 1px 3px #000,' . "\r\n" .
+                    '1px 1px 1px #000;',
+                'text-shadow: 1px 1px 3px #000,' . "\r\n" .
+                    '1px 1px 1px #000;',
```
oliverklee: You can shorten this a bit by using double quotes for the complete string (without using concatenation and without using the constant). (The rule is to have single quotes unless double quotes are needed.)

CHANGELOG.md
JakeQZ: Done & done.

tests/Unit/EmogrifierTest.php
JakeQZ: Done

JakeQZ:
Have updated the branch with the requested changes.

oliverklee:
Merged. Thanks for contributing to Emogrifier, @JakeQZ!  ❤️

PR #505, opened 2018-02-07
 [BUGFIX] Minified adjacent sibling combinator

JakeQZ:
Changed `$xPathRules` regular expression for adjacent sibling combinator to
match zero or more spaces either side, rather than one or more, allowing
selectors with this combinator in minified CSS to be matched.  Added PHPUnit
tests for such CSS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,10 +24,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
-- Allow CSS property values containing newlines
-  ([#504](https://github.com/MyIntervals/emogrifier/pull/504))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Fixed
 - Allow CSS property values containing newlines
   ([#504](https://github.com/MyIntervals/emogrifier/pull/504))
+- Allow adjacent sibling CSS selector combinator in minified CSS
+  ([#505](https://github.com/MyIntervals/emogrifier/pull/505))
```
oliverklee: The change log entries should be in reverse chronological oder (i.e., newest on top). I'll change that in a minute.

Issue #443, opened 2017-12-27
CSS selector to XPath: Descendant combinator not converted when descendant is an attribute selector


zoliszabo:
In the case of `A B[attr...]` CSS selectors (element B is descendant of element A and element B also has some attribute constraint specified), the resulting XPath is not correct.

For example:
Original CSS selector: `p a[href^=tel]`
The (wrong) current result of `translateCssToXpath`:  `//p a[starts-with(@href, "tel")]`
Correct XPath would be: `//p//a[starts-with(@href, "tel")]`

The source of this issue must be the [regex of the descendant combinator](https://github.com/MyIntervals/emogrifier/blob/097c5841b66cada54d3e9bd960b52c659f447d6b/Classes/Emogrifier.php#L167), which does not accept `[]` on the descendant element.

Issue #381, opened 2017-04-26
XPath error with input[type=text]


laurensschuitemaker:
Hi,

I noticed when my CSS contains input[type=text], it throws this error:
DOMXPath::query(): Invalid expression in selector >> div.form-field input[type="submit"] << in Emogrifier.php on line 370

How can I solve this?

Thanks

oliverklee:
Hi @laurensschuitemaker, this looks like a bug. Could you please post the full CSS rule that triggers this error?

laurensschuitemaker:
Hi @oliverklee, sure:
div.form-field input[type=text],
div.form-field input[type=password],
div.form-field textarea,
div.form-field select{
	width: 200px;
	padding: 3px;
	border: 1px solid #B6B6B6;
	outline-color: #048399;
}

This also doesn't work (same error):
.form-columns-switcher .column-switcher:nth-child(3n + 1){}

Thanks!

oliverklee:
Thanks!

oliverklee:
Fixed by 2c2b7853e35c8d03b301661bbe3732d7569d5ba2.

PR #327, opened 2016-07-24
[BUGFIX] Ignore value with words more than one in the attribute selector

taiyonoryoshy:
Fixes #327

PR #506, opened 2018-02-08
[BUGFIX] Descendent attribute selectors

JakeQZ:
Changed CSS-to-XPath translation matcher for descendant combinator to exclude
only spaces within `[...]`, i.e. when not followed by a closing `]` unless there
is an intervening opening `[` (rather than than excluding any space where the
remaining part of the expression ends with a closing `]`).  This matches spaces
in the original CSS selector but not spaces in already-translated XPath
predicates, and allows attribute selectors to be used in descendants.

Added PHPUnit tests for some attribute selectors as descendants, particularly
where the translated XPath predicate will contain a space.  Also added tests for
the converse (attribute selectors with descendants), and comments for relevant
tests that should be added but would currently fail due to other issues.

This should fix Issue
[#381](https://github.com/MyIntervals/emogrifier/issues/381) and at least
partially address Issue
[#443](https://github.com/MyIntervals/emogrifier/issues/443).

For reference, the original space-exclusion regex was added in
[#327](https://github.com/MyIntervals/emogrifier/pull/327) to avoid matching
spaces in selector attributes.  The purpose changed in
[#458](https://github.com/MyIntervals/emogrifier/pull/458), when the
CSS-to-XPath translation rules were reordered, to avoid matching spaces in XPath
predicates, though the requirement not to match spaces within `[...]` was
unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
+- Allow attribute selectors in descendants
+  ([#506](https://github.com/MyIntervals/emogrifier/pull/505),
+  [#381](https://github.com/MyIntervals/emogrifier/issues/381))
 - Allow adjacent sibling CSS selector combinator in minified CSS
   ([#505](https://github.com/MyIntervals/emogrifier/pull/505))
 - Allow CSS property values containing newlines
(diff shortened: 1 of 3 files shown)


oliverklee:
Merged. Thanks for contributing this PR to Emogrifier, @JakeQZ! :heart: :cookie:

Issue #296, opened 2016-01-28
Support @charset in the CSS


visioncan:
My css file with @charset "UTF-8" like:

``` css
@charset "UTF-8";
body {...} 
```

will output wraning:

``` sh
Warning: DOMXPath::query(): Invalid expression in /Users/../vendor/pelago/emogrifier/Classes/Emogrifier.php on line 353
```

oliverklee:
https://developer.mozilla.org/en/docs/Web/CSS/@charset

Currently, Emogrifier does not support charset in the CSS yet. If you are using only ASCII characters in your CSS, you can safely remove the charset.

JakeQZ:
We added support for dropping `@charset` rules without impacting CSS parsing in 2.0.

I don't think we'll ever support alternative charsets to UTF-8, which is the norm these days, unless that norm miraculously changes.

Can we close this as either 'fixed' or 'wontfix'?

oliverklee:
Yes, let's close this.

PR #507, opened 2018-02-08
[BUGFIX] Allow @charset without error

JakeQZ:
Filter out `@charset` at-rules from the CSS before processing, to avoid an error
when presented with CSS containing this at-rule.

Obviously this does not mean `@charset` is supported - its value is still
ignored - but does address the immediate error in #296.

Added PHPUnit test like that for `@import`, and also test that such CSS content
does not prevent Emogrifier working correctly on the remainder of the CSS.

### DIFF
--- a/tests/Unit/EmogrifierTest.php
+++ b/tests/Unit/EmogrifierTest.php
@@ -1091,7 +1091,7 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
      *
      * @dataProvider unneededCssThingsDataProvider
      */
-    public function emogrifyUnaffectedByUnneededCssThings($css)
+    public function emogrifyMatchesRuleAfterUnneededCssThing($css)
     {
         $this->subject->setHtml('<html><body></body></html>');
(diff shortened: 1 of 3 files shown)


tests/Unit/EmogrifierTest.php
```diff
@@ -1083,6 +1084,23 @@ public function emogrifyFiltersUnneededCssThings($css, $markerNotExpectedInHtml)
...
     }
 
+    /**
+     * @test
+     *
+     * @param string $css
+     *
+     * @dataProvider unneededCssThingsDataProvider
+     */
+    public function emogrifyUnaffectedByUnneededCssThings($css)
```
oliverklee: Please let's rename the test to be more specific. What does "unaffected" mean?

(And please add a verb so that the test name is a (possible pidgin) sentence.)

JakeQZ: Have renamed the method.

oliverklee:
Merged. Thanks for the bug fix, @JakeQZ! :heart:

Issue #508, opened 2018-02-14
Incorrect style property order with multiple mixed individual and shorthand properties


JakeQZ:
The following CSS
```
p { margin: 1em; margin-top: 0; }
.big p { margin: 2em; }
```
when applied to the HTML
```
<div class="big">
  <p>Some text</p>
</div>
```
results in
```
  <p style="margin: 2em; margin-top: 0;">Some text</p>
```
The `margin-top: 0` should either be before `margin: 2em` or (as an optimization in this case) not present at all.

JakeQZ:
The issue arises from the behaviour of `array_merge` in `generateStyleStringFromDeclarationsArrays`.  When a property is present in both `$oldStyles` and `$newStyles`, its position in the resultant array comes from its position in `$oldStyles`, though there may be a later shorthand or individual property variant it should now be overriding.

To resolve this and maintain correct ordering, I propose:
- Move `$combinedStyles = array_merge(...)` to after the first `foreach` loop;
- In the loop either `unset($newStyles[$attributeName])` or `unset($oldStyles[$attributeName])` depending which should prevail;
- Change the cache key to `serialize([$oldStyles, $newStyles])`, since the result of the `array_merge` would no longer be available initially.

I'd suggest the possible optimization of eliminating individual styles that are overridden with shorthand styles be addressed as a separate issue.

oliverklee:
@JakeQZ Sounds good. Would you be willing to tackle this bug?

JakeQZ:
@oliverklee Yes, I have already confirmed the proposed fix works...

oliverklee:
@JakeQZ #511 fixed this completely, so we can close this ticket, right?

JakeQZ:
Yes, am closing.  I'll create a separate issue for the optimization.

PR #510, opened 2018-02-14
[CLEANUP] Boilerplate HTML for tests

JakeQZ:
Corrected duplicated missing ">" typo in boilerplate HTML used by several tests,
adding common method for setting such HTML in the test subject.

Needed before adding additional test using this HTML for #508.


oliverklee:
Merged. Thanks, @JakeQZ! :heart:

PR #511, opened 2018-02-14
[BUGFIX] Individual/shorthand style property order

JakeQZ:
Fix for style property ordering when multiple mixed individual and shorthand
properties apply (#508).

In `generateStyleStringFromDeclarationsArrays`, an overridden style was
retaining its original position in the property list from the result of
`array_merge`, which could be before a shorthand/individual variant it should
apply after.

The properties that are overridden are now `unset` from the arrays before the
`array_merge`, so that the correct order is maintained.

### DIFF
--- a/tests/Unit/EmogrifierTest.php
+++ b/tests/Unit/EmogrifierTest.php
@@ -1739,12 +1739,10 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
 
         $result = $this->subject->emogrify();
 
-        // Property order in style attribute now follows precedence order, though used to follow application order
-        // Either is valid in this case
-        static::assertThat($result, static::logicalOr(
-            static::stringContains('<p style="margin: 0; padding-top: 1px; padding-bottom: 3px; text-align: center;">'),
-            static::stringContains('<p style="margin: 0; padding-bottom: 3px; padding-top: 1px; text-align: center;">')
-        ));
+        static::assertContains(
+            '<p style="margin: 0; padding-bottom: 3px; padding-top: 1px; text-align: center;">',
+            $result
+        );
     }
 
     /**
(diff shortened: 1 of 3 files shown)


tests/Unit/EmogrifierTest.php
```diff
@@ -1739,10 +1739,12 @@ public function emogrifyMergesCssWithMixedCaseAttribute()
 
         $result = $this->subject->emogrify();
 
-        static::assertContains(
-            '<p style="margin: 0; padding-top: 1px; padding-bottom: 3px; text-align: center;">',
-            $result
-        );
+        // Property order in style attribute now follows precedence order, though used to follow application order
+        // Either is valid in this case
+        static::assertThat($result, static::logicalOr(
```
zoliszabo: I think using only the new version output would be enough here.

JakeQZ: Ok, have changed this.

tests/Unit/EmogrifierTest.php
```diff
@@ -2015,6 +2015,54 @@ public function secondNonImportantStyleNotOverwritesFirstImportantOne()
         );
     }
 
+    /**
+     * @test
+     */
+    public function emogrifyAppliesLaterOverridingStyleAfterStyleAfterOverriddenStyle()
+    {
+        $this->setSubjectBoilerplateHtml();
+        $this->subject->setCss('p { margin-top: 1px; } p { margin: 2px; } p { margin-top: 3px; }');
```
oliverklee: These should be two separate tests: One that the (first) margin overrides the margin-top, and one that the (other) margin-top gets applies after the margin.

JakeQZ: Ok, I've added a separate test, which can be updated if/when elimination of individual properties overridden by shorthand properties is supported.

PR #513, opened 2018-02-15
[FEATURE] Add a HtmlNormalizer class

oliverklee:
This class is a "vanilla" subclass of the AbstractHtmlProcessor class. It
provides the following features:

- add a document type (HTML5) if missing
- disentangle incorrectly nested tags
- add HEAD and BODY elements (if they are missing)
- reformat the HTML


tests/Unit/Emogrifier/HtmlProcessor/HtmlNormalizerTest.php
```diff
@@ -0,0 +1,44 @@
...
+            '<head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head>' .
+            '<body></body>' .
+            '</html>';
+        $formattedHtml = '<!DOCTYPE HTML>' . "\n" .
+            '<html>' . "\n" .
+            '<head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head>' . "\n" .
+            '<body></body>' . "\n" .
+            '</html>' . "\n";
+
+        $subject = new TestingHtmlProcessor($rawHtml);
```
zoliszabo: I think this line should read `$subject = new HtmlNormalizer($rawHtml);`.

oliverklee: Oh, right. I'm going to fix this in a minute. Good catch!

oliverklee:
> Since this new HtmlNormalizer is just the non-abstract version of its base AbstractHtmlProcessor (am I getting this right?), shouldn't the unit tests AbstractHtmlProcessorTest.php and HtmlNormalizerTest.php combined into a single test (being somewhat redundant right now)?

The idea is that the test for AbstractHtmlProcessor tests what the `AbstractHtmlProcessor` does, and the tests for the subclasses then only need to test that the class is an `AbstractHtmlProcessor`, that it has the added functionality, and maybe some smoke tests.

tests/Unit/Emogrifier/HtmlProcessor/HtmlNormalizerTest.php
oliverklee: Fixed and re-pushed.

zoliszabo:
>> Since this new HtmlNormalizer is just the non-abstract version of its base AbstractHtmlProcessor (am I getting this right?), shouldn't the unit tests AbstractHtmlProcessorTest.php and HtmlNormalizerTest.php combined into a single test (being somewhat redundant right now)?

>The idea is that the test for AbstractHtmlProcessor tests what the AbstractHtmlProcessor does, and the tests for the subclasses then only need to test that the class is an AbstractHtmlProcessor, that it has the added functionality, and maybe some smoke tests.

Right. Maybe I am missing something here, but it's just that `HtmlNormalizer` in its current form does not add any new functionality, only inherits existing functionality implemented & covered by tests from `AbstractHtmlProcessor`. So - I think - only the  `instanceOf` test should be in the new test class. (This is also somewhat confirmed by the fact that test `reformatsHtml` worked just fine for the abstract class before you corrected it).

oliverklee:
> So - I think - only the instanceOf test should be in the new test class. 

Okay, I'll change the test accordingly. Hold on …

oliverklee:
Done and repushed.

PR #514, opened 2018-02-16
[TASK] Add a CssInliner class

oliverklee:
For Emogrifier 3.0.0, this will be the successor of the
\Pelago\Emogrifer class.

Currently, the new class is a plain copy of the Emogrifier class,
including all deprecated methods and without and API changes.


JakeQZ:
I'm guessing that `git`/GitHub does not support branching a single file within a repository.  If it did (as most other VCSs do), maintenance until 2.1 would probably be a lot easier, with semi-automatic merging between the two copies.

Issue #280, opened 2015-11-06
Process CSS selectors with states and add as style block


peterfox:
Currently if you process CSS with state CSS selectors e.g. :hover these selectors get lost in the inline process. Instead it would be ideal if they're added to the HTML as a style block instead in the same way the media is.

roman-1983:
+1

JakeQZ:
I'll be proposing a PR for this next week, so if anyone has had any more thoughts, or tried any code changes, let us know here.

oliverklee:
Sounds good to me. I say: Go for it! @jjriv, @zoliszabo: More thoughts?

zoliszabo:
> Sounds good to me. I say: Go for it! @jjriv, @zoliszabo: More thoughts?

Sounds good to me too! 
By the way: Nice PRs @JakeQZ!

jjriv:
I'm with @zoliszabo and @oliverklee

JakeQZ:
Proposed solution...

Currently we have in `process()`:

```
    $cssParts = $this->splitCssAndMediaQuery($allCss);
    //...
    $cssRules = $this->parseCssRules($cssParts['css']);
    foreach ($cssRules as $cssRule) {
        //...
```

Here `$cssRules` does not currently contain rules from media queries or with selectors involving pseudo -classes or -elements whose rules cannot be inlined.

I suggest changing this so that `$cssRules` contains all rules, with each entry having additional keys:
- "media" - Media query string if rule is from within a media query block;
- "hasPseudo" - Boolean indicating whether the selector involves a pseudo -class or -element such that the rules cannot be inlined;
- "uninlineable" - Convenience Boolean indicating either "media" is nonempty or "hasPseudo" is `true` (or both).

Alternatively, `$cssRules` could contain two arrays:
- Inlineable rules (as before), sorted by selector precedence;
- Uninlineable rules, i.e. those from media queries and/or with selectors involving psuedo-elements or behavioural pseudo-classes, with additional keys "media" and "hasPseudo" described above, and which do not need sorting by selector precedence.

The rationale is that original 'line number' information from media queries will also be required so that in the added style element they can be correctly ordered with respect to other rules for selectors involving pseudo-elements or behavioural psuedo-classes (and their rule blocks are currently parsed later anyway).

The loop which follows would ignore uninlineable rules.

[...]

PR #515, opened 2018-02-16
[CLEANUP] `splitCssAndMediaQuery`: retain ordering

JakeQZ:
Return from `splitCssAndMediaQuery` an array of split CSS parts for different
media queries, along with the parts of intervening CSS not within a media query,
in order.

The ordering will need to be retained for implementing #280.

Currently in `process` the return value is converted back to `$cssParts` as it
was before, as a temporary measure, pending further refactoring, of
`parseCssRules` and `copyCssWithMediaToStyleNode`.


src/Emogrifier.php
```diff
@@ -375,7 +375,27 @@ protected function process(\DOMDocument $xmlDocument)
...
 
-        $cssParts = $this->splitCssAndMediaQuery($allCss);
+        $splitCss = $this->splitCssAndMediaQuery($allCss);
+        // this is temporary pending refactoring `parseCssRules` and `copyCssWithMediaToStyleNode`
+        $cssParts = [
+            'css' => implode('', array_map(
+                function ($splitCssPart) {
+                    if ($splitCssPart['media'] === '') {
+                        return $splitCssPart['css'];
+                    }
```
zoliszabo: I know this is temporary, but please add `return '';`, so that the callback always returns a value (just for being explicit and have no IDE warnings ;) ).

src/Emogrifier.php
```diff
@@ -375,7 +375,27 @@ protected function process(\DOMDocument $xmlDocument)
...
+                },
+                $splitCss
+            )),
+            'media' => implode('', array_map(
+                function ($splitCssPart) {
+                    if ($splitCssPart['media'] !== '') {
+                        return $splitCssPart['media'] . '{' . $splitCssPart['css'] . '}';
+                    }
```
zoliszabo: Same here: `return '';`.

src/Emogrifier.php
```diff
@@ -1267,24 +1299,41 @@ private function splitCssAndMediaQuery($css)
...
+        // filter the CSS outside/between allowed @media rules
+        $cssCleaningMatchers = [
             'import/charset directives' => '/^\\s*@(?:import|charset)\\s[^;]+;/misU',
             'remaining media enclosures' => '/^\\s*@media\\s[^{]+{(.*)}\\s*}\\s/misU',
         ];
 
-        $cleanedCss = preg_replace($search, '', $cssForAllowedMediaTypes);
-
-        return ['css' => $cleanedCss, 'media' => $media];
+        $splitCss = [];
```
oliverklee: Even if the code may be temporary, please let's move it into a well-named private method instead of making this existing method even longer.

src/Emogrifier.php
```diff
@@ -1267,24 +1299,41 @@ private function splitCssAndMediaQuery($css)
...
+                $ruleBodyStartPosition = strpos($cssPart, '{');
+                $media = substr($cssPart, 0, $ruleBodyStartPosition);
+
+                // advance past "{"
+                ++$ruleBodyStartPosition;
+                $css = substr(
+                    $cssPart,
+                    $ruleBodyStartPosition,
+                    strrpos($cssPart, '}') - $ruleBodyStartPosition
+                );
```
oliverklee: I think a regular expression would make the intent a lot more clear for this.

src/Emogrifier.php
```diff
@@ -375,7 +375,27 @@ protected function process(\DOMDocument $xmlDocument)
             $allCss .= $this->getCssFromAllStyleNodes($xPath);
         }
 
-        $cssParts = $this->splitCssAndMediaQuery($allCss);
+        $splitCss = $this->splitCssAndMediaQuery($allCss);
+        // this is temporary pending refactoring `parseCssRules` and `copyCssWithMediaToStyleNode`
```
oliverklee: Even if the code is temporary, please let's move it into a well-named private method instead of making this existing method even longer.

src/Emogrifier.php
```diff
@@ -1267,24 +1299,41 @@ private function splitCssAndMediaQuery($css)
...
             'import/charset directives' => '/^\\s*@(?:import|charset)\\s[^;]+;/misU',
             'remaining media enclosures' => '/^\\s*@media\\s[^{]+{(.*)}\\s*}\\s/misU',
         ];
 
-        $cleanedCss = preg_replace($search, '', $cssForAllowedMediaTypes);
-
-        return ['css' => $cleanedCss, 'media' => $media];
+        $splitCss = [];
+        foreach ($cssSplitForAllowedMediaTypes as $i => $cssPart) {
+            if ($i & 1) {
```
oliverklee: If find this code not very "speaking". Can we make this more clear?

oliverklee:
Can we have tests for that the ordering is kept?

JakeQZ:
I've made the changes suggested, rebased, and merged the changes into `CssInliner` and `CssInlinerTest`.

oliverklee:
I'll have a look at the changes. @zoliszabo, please don't merge yet.

src/Emogrifier.php
```diff
@@ -375,7 +375,10 @@ protected function process(\DOMDocument $xmlDocument)
             $allCss .= $this->getCssFromAllStyleNodes($xPath);
         }
 
-        $cssParts = $this->splitCssAndMediaQuery($allCss);
+        $splitCss = $this->splitCssAndMediaQuery($allCss);
+        // this is temporary pending refactoring `parseCssRules` and `copyCssWithMediaToStyleNode`
```
oliverklee: I don't think we need this comment anymore. Even if we stopped after this PR, the refactoring would make sense.

src/Emogrifier.php
```diff
@@ -1267,24 +1282,70 @@ private function splitCssAndMediaQuery($css)
...
+                $css = preg_replace($cssCleaningMatchers, '', $cssPart);
+            }
+            $splitCss[] = compact('css', 'media');
+        }
+        return $splitCss;
+    }
 
-        return ['css' => $cleanedCss, 'media' => $media];
+    /**
+     * Temporary method pending refactoring of `parseCssRules` and `copyCssWithMediaToStyleNode`.
```
oliverklee: I don't think we need this comment anymore. Even if we stopped after this PR, the refactoring would make sense.

src/Emogrifier.php
```diff
@@ -1251,12 +1256,22 @@ private function getBodyElement(\DOMDocument $document)
...
+     *   0 => [
+     *     "css" => " h1 { color:red; } ",
+     *     "media" => ""
+     *   ],
+     *   1 => [
+     *     "css" => " h1 {}",
+     *     "media" => "@media "
+     *   ],
+     *   2 => [
+     *     "css" => " "
```
oliverklee: Will/should this really be a string that just contains a space? Should we trim this?

src/Emogrifier.php
```diff
@@ -1267,24 +1282,70 @@ private function splitCssAndMediaQuery($css)
...
+     * Temporary method pending refactoring of `parseCssRules` and `copyCssWithMediaToStyleNode`.
+     * Combines the parts returned by `splitCssAndMediaQuery` into just two strings.
+     *
+     * @param string[][] As returned by `splitCssAndMediaQuery`
+     *
+     * @return string[] Array with the following keys:
+     *                  "css" => the cleaned CSS without any @media rules
+     *                  "media" => the valuable @media rules from the original CSS
+     */
+    private function combineSplitCssAndMediaQueryParts($splitCss)
```
oliverklee: `$splitCss` should have an `array` type hint.

src/Emogrifier.php
```diff
@@ -1267,24 +1282,70 @@ private function splitCssAndMediaQuery($css)
...
-        return ['css' => $cleanedCss, 'media' => $media];
+    /**
+     * Temporary method pending refactoring of `parseCssRules` and `copyCssWithMediaToStyleNode`.
+     * Combines the parts returned by `splitCssAndMediaQuery` into just two strings.
+     *
+     * @param string[][] As returned by `splitCssAndMediaQuery`
+     *
+     * @return string[] Array with the following keys:
+     *                  "css" => the cleaned CSS without any @media rules
+     *                  "media" => the valuable @media rules from the original CSS
```
oliverklee: "valuable" sound a bit like "depends on the reader's judgement". ;-) What about "relevant" or "important" or "applicable"?

src/Emogrifier.php
```diff
@@ -1267,24 +1282,70 @@ private function splitCssAndMediaQuery($css)
...
+     */
+    private function combineSplitCssAndMediaQueryParts($splitCss)
+    {
+        return [
+            'css' => implode('', array_map(
+                function ($splitCssPart) {
+                    if ($splitCssPart['media'] === '') {
+                        return $splitCssPart['css'];
+                    } else {
+                        return '';
```
oliverklee: Shorter:

```php
return $splitCssPart['media'] === '' ? $splitCssPart['css'] : '';
```

src/Emogrifier.php
```diff
@@ -1267,24 +1282,70 @@ private function splitCssAndMediaQuery($css)
...
+                    }
+                },
+                $splitCss
+            )),
+            'media' => implode('', array_map(
+                function ($splitCssPart) {
+                    if ($splitCssPart['media'] !== '') {
+                        return $splitCssPart['media'] . '{' . $splitCssPart['css'] . '}';
+                    } else {
+                        return '';
```
oliverklee: Shorter:

```php
return $splitCssPart['media'] !== '' ? ($splitCssPart['media'] . '{' . $splitCssPart['css'] . '}') : '';
```

src/Emogrifier.php
```diff
@@ -1267,24 +1282,70 @@ private function splitCssAndMediaQuery($css)
...
+        // filter the CSS outside/between allowed @media rules
+        $cssCleaningMatchers = [
             'import/charset directives' => '/^\\s*@(?:import|charset)\\s[^;]+;/misU',
             'remaining media enclosures' => '/^\\s*@media\\s[^{]+{(.*)}\\s*}\\s/misU',
         ];
 
-        $cleanedCss = preg_replace($search, '', $cssForAllowedMediaTypes);
+        $splitCss = [];
+        foreach ($cssSplitForAllowedMediaTypes as $splitPartIndexWhereOddIsMediaRule => $cssPart) {
+            if ($splitPartIndexWhereOddIsMediaRule % 2 !== 0) {
```
oliverklee: We can simplify this (or make the intention more clear):

1. use `$index` or `$i` for the index variable (again)
2. `$isMediaRule = $i % 2 !== 0`
3. use `$isMediaRule` in the `if` (then we don't need the comments there anymore)

src/Emogrifier.php
```diff
@@ -1267,24 +1282,70 @@ private function splitCssAndMediaQuery($css)
...
+                // @media rule
+                preg_match('/^([^{]*+){(.*)}[^}]*+$/s', $cssPart, $matches);
+                $media = $matches[1];
+                $css = $matches[2];
+            } else {
+                // CSS outside/between allowed @media rules
+                $media = '';
+                $css = preg_replace($cssCleaningMatchers, '', $cssPart);
+            }
+            $splitCss[] = compact('css', 'media');
```
oliverklee: I propose setting `$splitCss[]` directly in the `if/else`. This will remove another layer of indirection, i.e., will make the code more straightforward to read.

src/Emogrifier.php
JakeQZ: Removed

src/Emogrifier.php
JakeQZ: Removed

src/Emogrifier.php
JakeQZ: I've added `trim` for the CSS outside of @media rules, and these parts are now only included if the result of `trim` is a nonempty string.

src/Emogrifier.php
JakeQZ: Added

src/Emogrifier.php
JakeQZ: Changed to "allowed" which is consistent with method names `addAllowedMediaType` and `removeAllowedMediaType`

src/Emogrifier.php
JakeQZ: Changed

src/Emogrifier.php
JakeQZ: Changed; however as this would result in exceeding the 120 character line length limit, and CodeSniffer does not allow ternary conditionals to be split onto multiple lines, I have added `extract` to assign the array values to local variables and allow the ternary conditional to be shorter.

src/Emogrifier.php
JakeQZ: Changed as suggested (using `$index`)

src/Emogrifier.php
JakeQZ: Changed as suggested, and renamed `$css` -> `$cleanedCss` in the non-media branch so that the original parameter `$css` is not overwritten (and thus remains available for debugging etc).

JakeQZ:
Have made suggested changes (more detail in reply to specific comments), rebased and merged changes into `CssInliner`.

oliverklee:
Merged. Thanks for the neat refactoring, @JakeQZ! :heart:

PR #517, opened 2018-02-17
[TASK] Remove CSS-to-attribute mapping from CssInliner

oliverklee:
This functionality is deprecated and should not be part of the new
CssInliner class.

This is the first step to cleaning up the CssInliner class after
copying it from the Emogrifier class.


zoliszabo:
@oliverklee Are you working on putting the code removed in this PR in a new `HtmlProcessor` class?

oliverklee:
> Are you working on putting the code removed in this PR in a new HtmlProcessor class?

Yes.

PR #521, opened 2018-02-20
[CLEANUP] Allow whitespace removal in @media tests

JakeQZ:
Added test assertion methods similar to `assertContains` and `assertNotContains`
but which also allow for removal of some unnecessary whitespace from the CSS,
without having to make the test data less humanly readable or change the test
inputs.

Changes for #280 will inevitably mean that during processing, due to the way
that some components of the CSS are `trim`med, some unnecessary whitespace is
removed from the CSS for media queries added to the style element.  Relevant
tests have been changed to used the new assertion methods.

### DIFF
--- a/tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
+++ b/tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
@@ -1290,16 +1290,16 @@ class CssInlinerTest extends \PHPUnit_Framework_TestCase
      * that allows for such whitespace removal, so that the tests themselves do not need to be written less humanly
      * readable and can use inputs containing extra whitespace.
      *
-     * @param string $needle Needle that would be used with `assertContains` or `assertNotContains`
+     * @param string $needle Needle that would be used with `assertContains` or `assertNotContains`.
      *
-     * @return string Needle to use with `assertRegExp` or `assertNotRegExp` instead
+     * @return string Needle to use with `assertRegExp` or `assertNotRegExp` instead.
      */
-    private static function getNeedleRegExpAllowingRemovalOfUnnecessaryCssWhitespace($needle)
+    private static function getCssNeedleRegExp($needle)
     {
         $needleMatcher = preg_replace_callback(
             '/\\s*+([{}])\\s*+|(?:(?!\\s*+[{}]).)++/',
-            function ($matches) {
-                if (!empty($matches[1])) {
+            function (array $matches) {
+                if (isset($matches[1]) && $matches[1] !== '') {
                     // matched possibly some whitespace, followed by "{" or "}", then possibly more whitespace
                     return '\\s*+' . preg_quote($matches[1], '/') . '\\s*+';
                 } else {
@@ -1313,30 +1313,31 @@ class CssInlinerTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
-     * Like `assertContains` but allows for removal of some unnecessary whitespace from the CSS
+     * Like `assertContains` but allows for removal of some unnecessary whitespace from the CSS.
      *
      * @param string $needle
      * @param string $haystack
      */
-    private static function assertContainsAllowingRemovalOfUnnecessaryCssWhitespace($needle, $haystack)
+    private static function assertContainsCss($needle, $haystack)
     {
         static::assertRegExp(
-            static::getNeedleRegExpAllowingRemovalOfUnnecessaryCssWhitespace($needle),
+            static::getCssNeedleRegExp($needle),
             $haystack,
             'Plain text needle: "' . $needle . '"'
         );
     }
 
     /**
-     * Like `assertNotContains` and also enforces the assertion with removal of some unnecessary whitespace from the CSS
+     * Like `assertNotContains` and also enforces the assertion with removal of some unnecessary whitespace from the
+     * CSS.
      *
      * @param string $needle
      * @param string $haystack
      */
-    private static function assertNotContainsAllowingRemovalOfUnnecessaryCssWhitespace($needle, $haystack)
+    private static function assertNotContainsCss($needle, $haystack)
     {
         static::assertNotRegExp(
-            static::getNeedleRegExpAllowingRemovalOfUnnecessaryCssWhitespace($needle),
+            static::getCssNeedleRegExp($needle),
             $haystack,
             'Plain text needle: "' . $needle . '"'
         );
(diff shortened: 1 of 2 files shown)


tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
```diff
@@ -1281,6 +1281,87 @@ public function emogrifyMatchesRuleAfterUnneededCssThing($css)
...
+     *
+     * @param string $needle Needle that would be used with `assertContains` or `assertNotContains`
+     *
+     * @return string Needle to use with `assertRegExp` or `assertNotRegExp` instead
+     */
+    private static function getNeedleRegExpAllowingRemovalOfUnnecessaryCssWhitespace($needle)
+    {
+        $needleMatcher = preg_replace_callback(
+            '/\\s*+([{}])\\s*+|(?:(?!\\s*+[{}]).)++/',
+            function ($matches) {
```
oliverklee: `$matches` can have the `array` type hinting.

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
```diff
@@ -1281,6 +1281,87 @@ public function emogrifyMatchesRuleAfterUnneededCssThing($css)
...
+                if (!empty($matches[1])) {
```
oliverklee: Please use `!== ''` instead of `empty`.  `empty` will also be true for expressions that cast to a boolean `false`, e.g. the non-empty string `"0"`.

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
```diff
@@ -1281,6 +1281,87 @@ public function emogrifyMatchesRuleAfterUnneededCssThing($css)
...
+        static::assertNotRegExp(
+            static::getNeedleRegExpAllowingRemovalOfUnnecessaryCssWhitespace($needle),
+            $haystack,
+            'Plain text needle: "' . $needle . '"'
+        );
+    }
+
+    /**
+     * Asserts that a string of CSS occurs exactly a certain number of times in the result, allowing for removal of some
+     * unnecessary whitespace
```
oliverklee: Please add the period at the end of all (semi-) full sentences in comments.

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
```diff
@@ -1281,6 +1281,87 @@ public function emogrifyMatchesRuleAfterUnneededCssThing($css)
...
+        return '/' . $needleMatcher . '/';
+    }
+
+    /**
+     * Like `assertContains` but allows for removal of some unnecessary whitespace from the CSS
+     *
+     * @param string $needle
+     * @param string $haystack
+     */
+    private static function assertContainsAllowingRemovalOfUnnecessaryCssWhitespace($needle, $haystack)
```
oliverklee: Maybe we can shorten this name (and the other method correspondingly) to something like "assertContainsCss"?

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
```diff
@@ -1281,6 +1281,87 @@ public function emogrifyMatchesRuleAfterUnneededCssThing($css)
...
+     */
+    private static function assertContainsCountAllowingRemovalOfUnnecessaryCssWhitespace(
+        $needle,
+        $expectedCount,
+        $haystack
+    ) {
+        static::assertSame(
+            $expectedCount,
+            preg_match_all(static::getNeedleRegExpAllowingRemovalOfUnnecessaryCssWhitespace($needle), $haystack),
+            'Plain text needle: "' . $needle . "\"\nHaystack: \"" . $haystack . '"'
```
oliverklee: If you use single quotes for the middle string, you won't need to escape the double quotes.

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
```diff
@@ -1281,6 +1281,87 @@ public function emogrifyMatchesRuleAfterUnneededCssThing($css)
...
+
+    /**
+     * Asserts that a string of CSS occurs exactly a certain number of times in the result, allowing for removal of some
+     * unnecessary whitespace
+     *
+     * @param string $needle
+     * @param int $expectedCount
+     * @param string $haystack
+     */
+    private static function assertContainsCountAllowingRemovalOfUnnecessaryCssWhitespace(
```
oliverklee: Please let's have the needle and the haystack directly after another and the `$expectedCount` first.

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
JakeQZ: But then to have the newline would require concatenating three strings:
`'"' . "\n" . 'Haystack: "'`

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
JakeQZ: Added.  I note I missed a type hint in the previous PR which you added before committing.  I've made a mental note to double-check these in future.

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
JakeQZ: Changed - this means an additional `isset` test because `preg_...` does not necessarily set the array entries for subpatterns within alternatives not matched.

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
oliverklee: Oh, right. I missed that.

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
JakeQZ: Changed 3 method descriptions and 2 parameter descriptions.

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
JakeQZ: Shortened the 4 method names accordingly.

JakeQZ:
I've made the suggested changes, apart from changing the string delimiters as the string contains a newline.

oliverklee:
Merged. Thanks for making helping make Emogrifier shine, @JakeQZ! :fireworks:

PR #526, opened 2018-02-21
[CLEANUP] Move CSS test assertion methods to trait

JakeQZ:
Added trait `AssertCss` with the methods `assertContainsCss`,
`assertNotContainsCss` and `assertContainsCssCount` moved from `CssInlinerTest`
and `EmogrifierTest`.

Added PHPUnit test case for the methods themselves.


tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
```diff
@@ -4,6 +4,7 @@
 
 use Pelago\Emogrifier;
 use Pelago\Emogrifier\CssInliner;
+use Pelago\Tests\Unit\Emogrifier\Support\AssertCss;
```
oliverklee: Please let's move the trait to the namespace `Pelago\Tests\Support\Traits` (so a relatively high level, and with `Traits` added).

tests/Unit/Emogrifier/Support/Tests/AssertCssTest.php
```diff
@@ -0,0 +1,287 @@
+<?php
+
+namespace Pelago\Tests\Unit\Emogrifier\Support\Tests;
+
+use Pelago\Tests\Unit\Emogrifier\Support\Tests\Fixtures\TestingAssertCss;
+
+/**
+ * Test case.
+ */
```
oliverklee: If you like, you can add yourself as an `@author` of the new files so proper credit is given.

tests/Unit/Emogrifier/HtmlProcessor/CssInlinerTest.php
JakeQZ: Ok.  Should the corresponding test then be in `Pelago\Tests\Unit\Tests\Support\Traits`?

oliverklee: Um … :-)

The tests for the trait should be in `Pelago\Tests\Unit\Support\Traits` (so only one ` Tests` ). (Tests for stuff in tests seems to resists namespacing that is both sane and consistent.)

tests/Unit/Emogrifier/Support/AssertCss.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $needleMatcher = preg_replace_callback(
+            '/\\s*+([{}])\\s*+|(?:(?!\\s*+[{}]).)++/',
+            function (array $matches) {
+                if (isset($matches[1]) && $matches[1] !== '') {
+                    // matched possibly some whitespace, followed by "{" or "}", then possibly more whitespace
+                    return '\\s*+' . preg_quote($matches[1], '/') . '\\s*+';
+                } else {
+                    // matched any other sequence which could not overlap with the above
+                    return preg_quote($matches[0], '/');
```
oliverklee: We don't need an else for this because the if has a return. (This is on of the changes in my "fix warnings" PR.)

JakeQZ:
I've made the suggested changes.

oliverklee:
@JakeQZ For future PRs, could you please edit your Git configuration and set your real name so the credits in the Git history are correct?

`git config --global user.name "Your full Name"`

tests/Unit/Support/Traits/AssertCssTest.php
```diff
@@ -0,0 +1,289 @@
...
+class AssertCssTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @test
+     */
+    public function getCssNeedleRegExpEscapesAllSpecialCharacters()
+    {
+        $needle = '.\\+*?[^]$(){}=!<>|:-/';
+
+        $result = TestingAssertCss::getCssNeedleRegExp($needle);
```
oliverklee: You don't need to have a separate class for this - you can directly `use` the trait in this test case class.

tests/Unit/Support/Traits/AssertCssTest.php
```diff
@@ -0,0 +1,289 @@
...
+
+        static::assertSame(
+            '/' . preg_quote($needle, '/') . '/',
+            str_replace('\\s*+', '', $result)
```
oliverklee: Why the str_replace?

tests/Unit/Support/Traits/AssertCssTest.php
```diff
@@ -0,0 +1,289 @@
...
+
+    /**
+     * @test
+     */
+    public function getCssNeedleRegExpNotEscapesNonSpecialCharacters()
+    {
+        $needle
+            = implode('', range('a', 'z'))
+            . implode('', range('A', 'Z'))
+            . implode('', range('0 ', '9 '))
```
oliverklee: Why not `implode(range(…) + range(…) + range(…))`?

tests/Unit/Support/Traits/AssertCssTest.php
```diff
@@ -0,0 +1,289 @@
...
+    {
+        $result = TestingAssertCss::getCssNeedleRegExp(
+            $otherContent . $contentToInsertAround . $otherContent
+        );
+
+        $quotedOtherContent = preg_quote($otherContent, '/');
+
+        static::assertSame(
+            '/' . $quotedOtherContent . '\\s*+' . preg_quote($contentToInsertAround, '/') . '\\s*+'
+                . $quotedOtherContent . '/',
```
oliverklee: Could you please extract this to a variable `$expectedContent`?

tests/Unit/Support/Traits/AssertCssTest.php
```diff
@@ -0,0 +1,289 @@
...
+
+    /**
+     * @test
+     *
+     * @param string $needle
+     * @param string $haystack
+     *
+     * @dataProvider needleFoundDataProvider
+     *
+     * @expectedException PHPUnit_Framework_ExpectationFailedException
```
oliverklee: Please add the missing leading Backslash to the expection class name to make it obvious to the reader that this is not namespaced.

oliverklee: Ditto for the other `@expectedException`.

tests/Unit/Support/Traits/AssertCssTest.php
JakeQZ: That did occur to me, though I left it as it was as there could be advantages to having the test and the tested less tightly coupled.  I'll change it.

tests/Unit/Support/Traits/AssertCssTest.php
JakeQZ: Because `$result` also contains inserted regex for optional whitespace, so would not match exactly.  I can introduce an intermediate variable `$resultWithOptionalWhitespaceRemoved` to make this (and similar other cases) clearer...

tests/Unit/Support/Traits/AssertCssTest.php
JakeQZ: It would need to be with `array_merge` rather than `+` as only the former concatenates arrays with equivalent numeric indexes, but can change it...

tests/Unit/Support/Traits/AssertCssTest.php
JakeQZ: Ok

tests/Unit/Support/Traits/AssertCssTest.php
JakeQZ: When I originally tried with the leading `\` it didn't work.  But now it does.

oliverklee: It's magic! ;-)

JakeQZ:
I've made the suggested changes.  Hopefully the name in the Git history credit is now correct for the latest commit.

JakeQZ:
Have merged and resolved the conflicts.

oliverklee:
Merged. Thanks for this neat cleanup, @JakeQZ! ❤️

PR #527, opened 2018-02-22
[BUGFIX] At-rules not at start of line

JakeQZ:
In `splitCssAndMediaQuery` corrected the cleaning regex to also remove at-rules
even if they are not at the start of a line (removing no-longer/not needed PCRE
modifiers, and for performance using possessive quantifiers where possible).


oliverklee:
Merged. Thanks for fixing this, @JakeQZ! ❤️

PR #528, opened 2018-02-22
[CLEANUP] Refactor `parseCssRules`: parse all CSS

JakeQZ:
Refactored `parseCssRules` so that it can now parse an entire CSS string,
including @media rules.  It now returns two arrays, in an array with the
following keys:
- "inlineable" => The inlineable rules that would have been previously returned;
- "uninlineable" => The uninlineable rules, currently those from within @media
  rules.

The original line number information from the CSS is available in both arrays,
so the original ordering of all the rules (that will be required for #280) can
be determined.

A temporary method `combineCssMediaRules` has been added to recombine the @media
rules back into a string of CSS as expected by `copyCssWithMediaToStyleNode`,
pending refactoring of that method to instead accept a parsed rule array as its
parameter.  This does mean that, for now, the rules within each @media rule are
parsed again.


JakeQZ:
As you'll see, the code in `combineCssMediaRules` is somewhat messy and I think would benefit from a `CssConcatenator` class to abstract the functionality for combining rules with the same media query, selectors or declarations block.  (This functionality will still be required after refactoring `copyCssWithMediaToStyleNode`, then by the refactored method itself.)  This I propose to add in the next incremental PR, though it could be included in this PR if desired...

src/Emogrifier.php
```diff
@@ -670,42 +675,60 @@ private function parseCssRules($css)
...
-                    $hasSupportedPseudoClass = (bool)preg_match(
-                        '/:(\\S+\\-(child|type\\()|not\\([[:ascii:]]*\\))/i',
-                        $selector
-                    );
-                    if ($hasPseudoElement || ($hasAnyPseudoClass && !$hasSupportedPseudoClass)) {
+            $splitCss = $this->splitCssAndMediaQuery($css);
+
+            $cssRules = [
+                'inlineable' => [],
+                'uninlineable' => []
```
oliverklee: Please add a trailing comma also after the last element of a multiline array. This makes appending elements possible without having to modify the existing line.

src/Emogrifier.php
```diff
@@ -670,42 +675,60 @@ private function parseCssRules($css)
...
 
-                    $cssRules[] = [
-                        'selector' => trim($selector),
-                        'declarationsBlock' => $cssDeclaration,
-                        // keep track of where it appears in the file, since order is important
-                        'line' => $key,
-                    ];
+                    $selectors = explode(',', $cssRule[1]);
+                    foreach ($selectors as $selector) {
+                        // don't process pseudo-elements and behavioral (dynamic) pseudo-classes;
```
oliverklee: Can we have this part (within the foreach loop of $selector, or most of it) in a separate private method?

src/Emogrifier.php
```diff
@@ -670,42 +675,60 @@ private function parseCssRules($css)
...
-                    );
-                    if ($hasPseudoElement || ($hasAnyPseudoClass && !$hasSupportedPseudoClass)) {
+            $splitCss = $this->splitCssAndMediaQuery($css);
+
+            $cssRules = [
+                'inlineable' => [],
+                'uninlineable' => []
+            ];
+            // keep track of where each rule appears in the file, since order is important
+            $key = 0;
```
oliverklee: Can't we have this as the key in the corresponding loop?

src/Emogrifier.php
```diff
@@ -1069,6 +1092,78 @@ private function attributeValueIsImportant($attributeValue)
...
 
+    /**
+     * Reconstructs the CSS containing only allowed @media rules from the "uninlineable" CSS rules array returned by
+     * `parseCssRules`.
+     *
+     * @param string[][] $cssRules
+     *
+     * @return string
+     */
+    private function combineCssMediaRules(array $cssRules)
```
oliverklee: Can you also please split up this method? It's far too long.

oliverklee:
As for the `CssConcatenator` class, we will only be able to have this for the `CssInliner` class, not for `Emogrifier` anymore (as the `Emogrifier` class is Composer-optional, and the class will go away for version 4.0.0).

oliverklee:
Oops, wrong button. Sorry, Reopening.

JakeQZ:
> As for the CssConcatenator class, we will only be able to have this for the CssInliner class, not for Emogrifier anymore (as the Emogrifier class is Composer-optional, and the class will go away for version 4.0.0).

Would something like 
```
        // support use without autoload
        if (!class_exists('Pelago\\Emogrifier\\CssConcatenator')) {
            require_once __DIR__ . '/Emogrifier/CssConcatenator.php';
        }
```
not be possible within the Emogrifier class?  I understand the need to support those not using autoload, but is it also a requirement to support grabbing just the `.php` file rather than the `src` directory (tree)?

src/Emogrifier.php
JakeQZ: Currently it needs to increment with the inner loop and carry on incrementing with each iteration of the outer loop.  However, it should be possible to have one double-loop doing the regex matching and `array_merge`-ing the results, then a second single loop processing the matches...

oliverklee:
No, that would be okay for me.

src/Emogrifier.php
JakeQZ: This is why I'd like to add a `CssConcatenator` class.

JakeQZ:
I've added a `CssConcatenator` class as a separate PR (#530).  I'll come back to this one when that is agreed.

src/Emogrifier.php
JakeQZ: Done.

src/Emogrifier.php
JakeQZ: I've moved the outer loop and regex matching into a separate method, so the loops are now very similar to how they were prior to this PR.

src/Emogrifier.php
JakeQZ: Changed as per comment above.

src/Emogrifier.php
JakeQZ: Now using `CssConcatenator`.

JakeQZ:
Now made the changes suggested in review.  Also rebased and reverted changes to `CssInliner.php` which can be merged as a separate PR once agreed.

src/Emogrifier.php
```diff
@@ -694,23 +701,62 @@ private function parseCssRules($css)
...
+            usort($cssRules['inlineable'], [$this, 'sortBySelectorPrecedence']);
 
             $this->caches[static::CACHE_KEY_CSS][$cssKey] = $cssRules;
         }
 
         return $this->caches[static::CACHE_KEY_CSS][$cssKey];
     }
 
+    /**
+     * Parse a string of CSS into the media query, selectors and declarations for each ruleset in order
```
oliverklee: Third person, please ("Parses") and a period at the end of sentence.

src/Emogrifier.php
```diff
@@ -694,23 +701,62 @@ private function parseCssRules($css)
...
+                        'media' => $cssRule['media'],
                         'selector' => trim($selector),
                         'declarationsBlock' => $cssDeclaration,
                         // keep track of where it appears in the file, since order is important
                         'line' => $key,
                     ];
+                    if ($cssRule['media'] === '') {
+                        $cssRules['inlineable'][] = $parsedCssRule;
+                    } else {
+                        $cssRules['uninlineable'][] = $parsedCssRule;
```
oliverklee: I think we can shorten this to something like this:

```php
$ruleType = ($cssRule['media'] === '') ? 'inlineable' : 'uninlineable';
$cssRules[$ruleType] = $parsedCssRule;
```

PR #530, opened 2018-02-23
[CLEANUP] Add a CssConcatenator class

JakeQZ:
This abstracts the (re-)combining of CSS rules for various media queries and
with various selectors and declaration blocks, merging adjacent rule blocks
where possible (i.e. for the same media query, with the same selectors, or with
the same declarations block).

Although not yet utilized, it will be required for #280 to simplify the code.

### DIFF
--- a/src/Emogrifier/CssConcatenator.php
+++ b/src/Emogrifier/CssConcatenator.php
@@ -1,0 +1,137 @@
+<?php
+
+namespace Pelago\Emogrifier;
+
+/**
+ * Facilitates building a CSS string by appending rule blocks one at a time, checking whether the media query,
+ * selectors, or declarations block are the same as those from the preceding block and combining blocks in such cases.
+ *
+ * @author Jake Hotson <jake.github@qzdesign.co.uk>
+ */
+class CssConcatenator
+{
+    /**
+     * CSS under construction.
+     *
+     * @var string
+     */
+    private $css = '';
+
+    /**
+     * Current media query string, e.g. "@media screen and (max-width:639px)" in the currently open media query block,
+     * or an empty string if not currently within a media query block.
+     *
+     * @var string
+     */
+    private $currentMedia = '';
+
+    /**
+     * Array whose keys are selectors for the rule block currently under construction (values are of no significance),
+     * or an empty array if no rule block under construction.
+     *
+     * @var int[]
+     */
+    private $currentSelectorsAsKeys = [];
+
+    /**
+     * Declarations for the rule block currently under construction,
+     * or an empty string if no rule block under construction.
+     *
+     * @var string
+     */
+    private $currentDeclarationsBlock = '';
+
+    /**
+     * Allow extending classes to call `parent::__construct()`.
+     */
+    public function __construct()
+    {
+    }
+
+    /**
+     * Append a declaration block to the CSS.
+     *
+     * @param string[]|string $selectors Array of selectors for the rule, e.g. ["ul", "ol", "p:first-child"],
+     *                                   or a single selector, e.g. "ul".
+     * @param string $declarationsBlock The property declarations, e.g. "margin-top: 0.5em; padding: 0".
+     * @param string $media The media query for the rule, e.g. "@media screen and (max-width:639px)",
+     *                      or an empty string if none.
+     */
+    public function append($selectors, $declarationsBlock, $media = '')
+    {
+        $selectorsAsKeys = array_flip((array)$selectors);
+
+        if ($media !== $this->currentMedia) {
+            $this->closeBlocks();
+            if ($media !== '') {
+                $this->css .= $media . '{';
+                $this->currentMedia = $media;
+            }
+        }
+
+        if ($declarationsBlock === $this->currentDeclarationsBlock) {
+            $this->currentSelectorsAsKeys += $selectorsAsKeys;
+        } elseif ($this->hasEquivalentCurrentSelectors($selectorsAsKeys)) {
+            $this->currentDeclarationsBlock
+                = rtrim(rtrim($this->currentDeclarationsBlock), ';') . ';' . $declarationsBlock;
+        } else {
+            $this->closeRuleBlock();
+            $this->currentSelectorsAsKeys = $selectorsAsKeys;
+            $this->currentDeclarationsBlock = $declarationsBlock;
+        }
+    }
+
+    /**
+     * Close any open rule or media blocks and return the CSS.
+     *
+     * @return string
+     */
+    public function getCss()
+    {
+        $this->closeBlocks();
+        return $this->css;
+    }
+
+    /**
+     * Close any open rule or media blocks.
+     *
+     * @return void
+     */
+    private function closeBlocks()
+    {
+        $this->closeRuleBlock();
+        if ($this->currentMedia !== '') {
+            $this->css .= '}';
+            $this->currentMedia = '';
+        }
+    }
+
+    /**
+     * Close any rule block under construction, appending its contents to the CSS.
+     *
+     * @return void
+     */
+    private function closeRuleBlock()
+    {
+        if ($this->currentSelectorsAsKeys !== [] && $this->currentDeclarationsBlock !== '') {
+            $this->css .= implode(',', array_keys($this->currentSelectorsAsKeys))
+                . '{' . $this->currentDeclarationsBlock . '}';
+        }
+        $this->currentSelectorsAsKeys = [];
+        $this->currentDeclarationsBlock = '';
+    }
+
+    /**
+     * Test if a set of selectors is equivalent to that for the rule block currently under construction
+     * (i.e. the same selectors, possibly in a different order).
+     *
+     * @param int[] $selectorsAsKeys Array in which the selectors are the keys, and the values are of no significance
+     *
+     * @return bool
+     */
+    private function hasEquivalentCurrentSelectors(array $selectorsAsKeys)
+    {
+        return count($selectorsAsKeys) === count($this->currentSelectorsAsKeys)
+            && count($selectorsAsKeys) === count($this->currentSelectorsAsKeys + $selectorsAsKeys);
+    }
+}
--- a/src/Emogrifier/CssConcatenator.php
+++ b/src/Emogrifier/CssConcatenator.php
@@ -6,6 +6,33 @@ namespace Pelago\Emogrifier;
  * Facilitates building a CSS string by appending rule blocks one at a time, checking whether the media query,
  * selectors, or declarations block are the same as those from the preceding block and combining blocks in such cases.
  *
+ * Example:
+ *  $concatenator = new CssConcatenator();
+ *  $concatenator->append(['body'], 'color: blue;');
+ *  $concatenator->append(['body'], 'font-size: 16px;');
+ *  $concatenator->append(['p'], 'margin: 1em 0;');
+ *  $concatenator->append(['ul', 'ol'], 'margin: 1em 0;');
+ *  $concatenator->append(['body'], 'font-size: 14px;', '@media screen and (max-width: 400px)');
+ *  $concatenator->append(['ul', 'ol'], 'margin: 0.75em 0;', '@media screen and (max-width: 400px)');
+ *  $css = $concatenator->closeBlocksAndGetCss();
+ *
+ * `$css` (if unminified) would contain the following CSS:
+ * ` body {
+ * `   color: blue;
+ * `   font-size: 16px;
+ * ` }
+ * ` p, ul, ol {
+ * `   margin: 1em 0;
+ * ` }
+ * ` @media screen and (max-width: 400px) {
+ * `   body {
+ * `     font-size: 14px;
+ * `   }
+ * `   ul, ol {
+ * `     margin: 0.75em 0;
+ * `   }
+ * ` }
+ *
  * @author Jake Hotson <jake.github@qzdesign.co.uk>
  */
 class CssConcatenator
@@ -42,24 +69,16 @@ class CssConcatenator
     private $currentDeclarationsBlock = '';
 
     /**
-     * Allow extending classes to call `parent::__construct()`.
-     */
-    public function __construct()
-    {
-    }
-
-    /**
-     * Append a declaration block to the CSS.
+     * Appends a declaration block to the CSS.
      *
-     * @param string[]|string $selectors Array of selectors for the rule, e.g. ["ul", "ol", "p:first-child"],
-     *                                   or a single selector, e.g. "ul".
+     * @param string[] $selectors Array of selectors for the rule, e.g. ["ul", "ol", "p:first-child"].
      * @param string $declarationsBlock The property declarations, e.g. "margin-top: 0.5em; padding: 0".
      * @param string $media The media query for the rule, e.g. "@media screen and (max-width:639px)",
      *                      or an empty string if none.
      */
-    public function append($selectors, $declarationsBlock, $media = '')
+    public function append(array $selectors, $declarationsBlock, $media = '')
     {
-        $selectorsAsKeys = array_flip((array)$selectors);
+        $selectorsAsKeys = array_flip($selectors);
 
         if ($media !== $this->currentMedia) {
             $this->closeBlocks();
@@ -82,18 +101,18 @@ class CssConcatenator
     }
 
     /**
-     * Close any open rule or media blocks and return the CSS.
+     * Closes any open rule or media blocks and returns the CSS.
      *
      * @return string
      */
-    public function getCss()
+    public function closeBlocksAndGetCss()
     {
         $this->closeBlocks();
         return $this->css;
     }
 
     /**
-     * Close any open rule or media blocks.
+     * Closes any open rule or media blocks.
      *
      * @return void
      */
@@ -107,7 +126,7 @@ class CssConcatenator
     }
 
     /**
-     * Close any rule block under construction, appending its contents to the CSS.
+     * Closes any rule block under construction, appending its contents to the CSS.
      *
      * @return void
      */
@@ -122,7 +141,7 @@ class CssConcatenator
     }
 
     /**
-     * Test if a set of selectors is equivalent to that for the rule block currently under construction
+     * Tests if a set of selectors is equivalent to that for the rule block currently under construction
      * (i.e. the same selectors, possibly in a different order).
      *
      * @param int[] $selectorsAsKeys Array in which the selectors are the keys, and the values are of no significance
--- a/src/Emogrifier/CssConcatenator.php
+++ b/src/Emogrifier/CssConcatenator.php
@@ -14,7 +14,7 @@ namespace Pelago\Emogrifier;
  *  $concatenator->append(['ul', 'ol'], 'margin: 1em 0;');
  *  $concatenator->append(['body'], 'font-size: 14px;', '@media screen and (max-width: 400px)');
  *  $concatenator->append(['ul', 'ol'], 'margin: 0.75em 0;', '@media screen and (max-width: 400px)');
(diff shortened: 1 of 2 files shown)


oliverklee:
Hi @JakeQZ, don't you ever sleep? ;-)

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Allow extending classes to call `parent::__construct()`.
+     */
+    public function __construct()
+    {
+    }
+
+    /**
+     * Append a declaration block to the CSS.
```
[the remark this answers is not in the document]
oliverklee: Ditto for the other method comments.

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,137 @@
...
+                = rtrim(rtrim($this->currentDeclarationsBlock), ';') . ';' . $declarationsBlock;
+        } else {
+            $this->closeRuleBlock();
+            $this->currentSelectorsAsKeys = $selectorsAsKeys;
+            $this->currentDeclarationsBlock = $declarationsBlock;
+        }
+    }
+
+    /**
+     * Close any open rule or media blocks and return the CSS.
```
oliverklee: We should separate this. The method name makes it very un-obvious that this method has side effects. (Or it should get a different name.)

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,137 @@
+<?php
+
+namespace Pelago\Emogrifier;
+
+/**
+ * Facilitates building a CSS string by appending rule blocks one at a time, checking whether the media query,
+ * selectors, or declarations block are the same as those from the preceding block and combining blocks in such cases.
```
oliverklee: A usage example would be great here in the class doc.

tests/Unit/Emogrifier/CssConcatenatorTest.php
```diff
@@ -0,0 +1,267 @@
...
+    {
+        $result = $this->subject->getCss();
+
+        static::assertSame('', $result);
+    }
+
+    /**
+     * @test
+     */
+    public function getCssReturnsSingleAppendedRule()
```
oliverklee: I think the test names should focus on what `append` does, not what `getCss` does (as `append` is what does the work). In this case, I propose the name `appendSetsFirstRule` (or something similar).

tests/Unit/Emogrifier/CssConcatenatorTest.php
```diff
@@ -0,0 +1,267 @@
...
+
+        $result = $this->subject->getCss();
+
+        static::assertSame('p{color: green;}', $result);
+    }
+
+    /**
+     * @test
+     */
+    public function getCssReturnsSingleAppendedMediaRule()
```
oliverklee: Accordingly, in this case the name then would be `appendWithMediaRuleAppendsToMediaRule` (or something similar).

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,137 @@
...
+     * or an empty string if no rule block under construction.
+     *
+     * @var string
+     */
+    private $currentDeclarationsBlock = '';
+
+    /**
+     * Allow extending classes to call `parent::__construct()`.
+     */
+    public function __construct()
```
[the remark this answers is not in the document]
JakeQZ: Ok, have removed it (there is usually a case for it in general, as it would allow extending classes to call `parent::__construct` so that if in future some action was required in the constructor it could be added without breaking extending classes).

src/Emogrifier/CssConcatenator.php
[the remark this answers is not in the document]
JakeQZ: Changed 5 method descriptions.

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,137 @@
...
+    {
+    }
+
+    /**
+     * Append a declaration block to the CSS.
+     *
+     * @param string[]|string $selectors Array of selectors for the rule, e.g. ["ul", "ol", "p:first-child"],
+     *                                   or a single selector, e.g. "ul".
```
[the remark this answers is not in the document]
JakeQZ: Removed possibility of passing a string instead of an array for a single selector - it's just two extra characters to add `[...]` around it in the calling code and it was converted to an array anyway.

src/Emogrifier/CssConcatenator.php
JakeQZ: Renamed the method - there is no reason to allow access the CSS with an incomplete rule block.

src/Emogrifier/CssConcatenator.php
JakeQZ: Added

tests/Unit/Emogrifier/CssConcatenatorTest.php
JakeQZ: Renamed 2 test methods.

JakeQZ:
> Hi @JakeQZ, don't you ever sleep? ;-)

;) Yes, but usually a couple of hours later than most people.

JakeQZ:
I've made the suggested changes.

oliverklee:
I'm currently sick. I'll do the reviews once my brain is not to mushy anymore.

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,156 @@
...
+ * `   margin: 1em 0;
+ * ` }
+ * ` @media screen and (max-width: 400px) {
+ * `   body {
+ * `     font-size: 14px;
+ * `   }
+ * `   ul, ol {
+ * `     margin: 0.75em 0;
+ * `   }
+ * ` }
```
oliverklee: I think a plain string as data structure for this is not optimal. I propose putting this in an array (or object) structure that then is serialized to a string at the end. Like this:
[string media, string[] selectors, string[] rules]

This would also make the code more straightforward.

@JakeQZ What do you think?

JakeQZ: I don't see how this would help with the data structure given.  Wouldn't it just be in effect buffering the calls to `append()` to be replayed later, when the same processing would still be required?

If there was to be some kind of intermediate data structure, I think to be of benefit it would need to reflect the nested nature of the CSS, i.e. an array of media rules at the top level (including `""` for rules not within a media rule), and within each an array of rule blocks for selectors.  So in the example, after combining adjacent rules, we'd have something like
```
[
  [
    'media' => '',
    'rules' => [
      [
        'selectors' => ['body'],
        'declarations' => 'color: blue; font-size: 16px;',
      ],
      [
        'selectors' => ['p', 'ul', 'ol']
        'declarations' => 'margin: 1em 0;',
      ], 
    ],
  ],
  [
    'media' => '@media screen and (max-width: 400px)',
    'rules' => [
      [
        'selectors' => ['body'],
        'declarations' => 'font-size: 14px;'
      ],
      [
        'selectors' => ['ul', 'ol']
        'declarations' => 'margin: 0.75em 0;'
      ],
    ],
  ],
];
```

`append()` could then combine the rule being added with the last entry(ies) of the existing data where media/selectors/declarations match, and create a new entry where they don't.  I don't see any particular point deferring the combining of rules.

Does that sound okay?

oliverklee: > I don't see how this would help with the data structure given. Wouldn't it just be in effect buffering the calls to append() to be replayed later, when the same processing would still be required?

The idea is that this will make the code straightforward: "adding CSS" means "sort it into the existing structure that is eminent in the code", and "getting the CSS" means "serializing the existing structure". So there is no need for the structure to have state (like "currently within a media query").

oliverklee: > append() could then combine the rule being added with the last entry(ies) of the existing data where media/selectors/declarations match, and create a new entry where they don't. I don't see any particular point deferring the combining of rules.
>
> Does that sound okay?

Yes, that sounds perfect!

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,122 @@
...
+     * - string `media` - The media query string, e.g. "@media screen and (max-width:639px)", or an empty string for
+     *   rules not within a media query block;
+     * - stdClass[] `ruleBlocks` - Array of rule blocks in order, where each element is an object with the following
+     *   properties:
+     *   - int[] `selectorsAsKeys` - Array whose keys are selectors for the rule block (values are of no significance);
+     *   - string `declarationsBlock` - The property declarations, e.g. "margin-top: 0.5em; padding: 0".
+     *
+     * @var stdClass[]
+     */
+    private $mediaRules = [];
```
oliverklee: I've noticed another way we could simplify this: We could use the media as array key for `$mediaRules`, effectively removing another dimension from the array, and allowing us to simplify the lookups. @JakeQZ What do you think?

JakeQZ: This wouldn't work if there were multiple media rules for the same media query, interspersed with other rules, e.g.
```
blockquote::before {
  font-size: 16px;
}
@media (max-width: 479px) {
  blockquote::before {
    font-size: 14px;
  }
}
.smallprint blockquote::before {
  font-size: 12px;
}
@media (max-width: 479px) {
  .smallprint blockquote::before {
    font-size: 10px;
  }
}
```
The final `10px` declaration would end up in the first media rule and be overridden by the `12px` declaration when it shouldn't.

oliverklee: Ah, makes sense. Is this case already covered by the tests?

JakeQZ: No, currently they just test combining or not combining pairs of rules.  I'll add a test for not combining rules (which otherwise would be) when there is some other rule between.

JakeQZ:
I've now added tests that nonadjacent rules are not combined (which would result in out-of-order CSS).

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,122 @@
...
+        $css = '';
+        foreach ($this->mediaRules as $mediaRule) {
+            if ($mediaRule->media !== '') {
+                $css .= $mediaRule->media . '{';
+            }
+            foreach ($mediaRule->ruleBlocks as $ruleBlock) {
+                $css .= implode(',', array_keys($ruleBlock->selectorsAsKeys))
+                    . '{' . $ruleBlock->declarationsBlock . '}';
+            }
+            if ($mediaRule->media !== '') {
```
oliverklee: We can simplify this by building the `$currentCssRule` first and then wrapping it if `$mediaRule->media` isn't an empty string.

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+        return $css;
+    }
+
+    /**
+     * Tests if two sets of selectors are equivalent (i.e. the same selectors, possibly in a different order).
+     *
+     * @param int[] $selectorsAsKeys1 Array in which the selectors are the keys, and the values are of no significance.
+     * @param int[] $selectorsAsKeys2 Another such array.
```
oliverklee: `int[]` for selectors does not seem to make sense here. Should be if `string[]` instead?

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @return bool
+     */
+    private static function hasEquivalentSelectors(array $selectorsAsKeys1, array $selectorsAsKeys2)
+    {
+        return count($selectorsAsKeys1) === count($selectorsAsKeys2)
+            && count($selectorsAsKeys1) === count($selectorsAsKeys1 + $selectorsAsKeys2);
```
oliverklee: Will `count($selectorsAsKeys1 + $selectorsAsKeys2)` be the same as `count($selectorsAsKeys1) + count($selectorsAsKeys2)`? If so, we can put the two latter counts into local variables and then work on those, allowing us to have only two count calls.

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    /**
+     * @return string
+     */
+    public function getCss()
+    {
+        $css = '';
+        foreach ($this->mediaRules as $mediaRule) {
```
oliverklee: Please let's simplify this by moving parts of the code to a separate, well-named method, e.g. `buildCssForMediaRule($mediaRule)`. This will make the intention of the code more clear to the reader.

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,122 @@
...
+ */
+class CssConcatenator
+{
+    /**
+     * Array of media rules in order.  Each element is an object with the following properties:
+     * - string `media` - The media query string, e.g. "@media screen and (max-width:639px)", or an empty string for
+     *   rules not within a media query block;
+     * - stdClass[] `ruleBlocks` - Array of rule blocks in order, where each element is an object with the following
+     *   properties:
+     *   - int[] `selectorsAsKeys` - Array whose keys are selectors for the rule block (values are of no significance);
```
oliverklee: Can't we use the selectors as array values?

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,122 @@
...
+     */
+    public function getCss()
+    {
+        $css = '';
+        foreach ($this->mediaRules as $mediaRule) {
+            if ($mediaRule->media !== '') {
+                $css .= $mediaRule->media . '{';
+            }
+            foreach ($mediaRule->ruleBlocks as $ruleBlock) {
+                $css .= implode(',', array_keys($ruleBlock->selectorsAsKeys))
```
oliverklee: Please let's put `array_keys($ruleBlock->selectorsAsKeys` into a well-named variable to make the intention more clear, e.g. `$selectors`.

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,122 @@
...
+        $lastMediaRule = end($this->mediaRules);
+        if ($lastMediaRule !== false && $media === $lastMediaRule->media) {
+            $mediaRule = $lastMediaRule;
+        } else {
+            $mediaRule = (object)[
+                'media' => $media,
+                'ruleBlocks' => [],
+            ];
+            $this->mediaRules[] = $mediaRule;
+        }
```
oliverklee: Please let's move the code above this to a separate method with a "speaking" name, e.g. `getOrCreateLastMediaRuleForMedia`.

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,122 @@
...
+        if ($lastRuleBlock !== false && $declarationsBlock === $lastRuleBlock->declarationsBlock) {
+            $lastRuleBlock->selectorsAsKeys += $selectorsAsKeys;
+        } elseif ($lastRuleBlock !== false
+            && $this->hasEquivalentSelectors($selectorsAsKeys, $lastRuleBlock->selectorsAsKeys)
+        ) {
+            $lastRuleBlock->declarationsBlock
+                = rtrim(rtrim($lastRuleBlock->declarationsBlock), ';') . ';' . $declarationsBlock;
+        } else {
+            $mediaRule->ruleBlocks[] = (object)compact('selectorsAsKeys', 'declarationsBlock');
+        }
```
oliverklee: I don't understand the code above this. :crying_cat_face: Could you please refactor it with "speaking" local variables or separate methods?

src/Emogrifier/CssConcatenator.php
JakeQZ: The array with the actual values in the array keys is a set as described in the question and accepted answer at https://stackoverflow.com/questions/1166953.  Putting the actual values in the array values would make lookup, set comparison, elimination of duplicates, etc., very inefficient, as indicated in some of the comments there, as well as at https://secure.php.net/manual/en/function.array-unique.php#98453.

src/Emogrifier/CssConcatenator.php
JakeQZ: It should probably be `mixed`.  It's whatever the keys happened to be in the original array of selectors passed, before `array_flip`, so it could be int or string.

src/Emogrifier/CssConcatenator.php
JakeQZ: `count(A+B)` gives the number of unique keys in the union of A and B, and will be lower than `count(A) + count(B)` whenever there are common keys.

src/Emogrifier/CssConcatenator.php
JakeQZ: Would a local variable for each of the conditions help here - e.g. `$sameDeclarationsAsLastRule` and `$sameSelectorsAsLastRule`?  Or methods for the conditional blocks like `addSelectorsToRuleBlock`, `addDeclarationsToRuleBlock` and `addRuleBlockToMediaRule`?  Or both?  And which, specifically (to me it seems redundant to add a method for something like `$lastRuleBlock->selectorsAsKeys += $selectorsAsKeys`, but somewhere a line must be drawn for what justifies a separate method)?

JakeQZ:
I've replied to some of the comments, where I'd like feedback before proceeding.  If I haven't replied to a particular comment, the response is "okay" (or I missed it).

src/Emogrifier/CssConcatenator.php
oliverklee: Ah, I didn't know that. Thanks for the link!

src/Emogrifier/CssConcatenator.php
JakeQZ: Done.

src/Emogrifier/CssConcatenator.php
JakeQZ: Moved the body if each `for` loop into separate methods - which are now be passed to `array_map` avoiding the loop constructs altogether.

src/Emogrifier/CssConcatenator.php
JakeQZ: Done.

src/Emogrifier/CssConcatenator.php
JakeQZ: Done.

JakeQZ:
I've made the changes suggested.  I wasn't totally sure how to make the code in `append()` easier to understand, but have added local variables for the conditions as well as the declarations block with the optional final semicolon removed.

src/Emogrifier/CssConcatenator.php
```diff
@@ -0,0 +1,154 @@
...
+
+        $hasSameDeclarationsAsLastRule = $lastRuleBlock !== false
+            && $declarationsBlock === $lastRuleBlock->declarationsBlock;
+        if ($hasSameDeclarationsAsLastRule) {
+            $lastRuleBlock->selectorsAsKeys += $selectorsAsKeys;
+        } else {
+            $hasSameSelectorsAsLastRule = $lastRuleBlock !== false
+                && $this->hasEquivalentSelectors($selectorsAsKeys, $lastRuleBlock->selectorsAsKeys);
+            if ($hasSameSelectorsAsLastRule) {
+                $lastDeclarationsBlockWithoutSemicolon = rtrim(rtrim($lastRuleBlock->declarationsBlock), ';');
```
oliverklee: Can't we do this with a single rtrim that has the semicolon and space as second parameter?

JakeQZ: Yes, good point.  However, we'd then have to (re-)specify all of the whitespace characters in the call to `rtrim`, which would be a slight drawback, so I'm not so sure.

oliverklee: Okay, then let's leave it as is.

PR #532, opened 2018-02-27
[BUGFIX] @media query with extra whitespace

JakeQZ:
Corrected regular expression for splitting out @media rules from the CSS not to
include disallowed media types with extra whitespace before the type.


JakeQZ:
I'll merge the changes into `CssInliner` when they are approved...

PR #534, opened 2018-02-28
[BUGFIX] @media rule after empty @media rule

JakeQZ:
Corrected regular expression for splitting out `@media` rules from the CSS not
to consume a subsequent `@media` rule after an empty one.

Similarly corrected the regular expression cleaning the unneeded `@media` rules,
using a common subexpression for matching a `@media` rule body.

(Found while working on #280, as the regular expression would match `"@media
screen {} @media tv { h1 { color: red; } }"` as a single `@media` rule, causing an
issue parsing its presumed rule body later.)

### DIFF
--- a/tests/Unit/EmogrifierTest.php
+++ b/tests/Unit/EmogrifierTest.php
@@ -1726,6 +1726,85 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @return mixed[][]
+     */
+    public function mediaTypesDataProvider()
+    {
+        return [
+            'disallowed type after disallowed type' => ['tv', 'speech', false],
+            'allowed type after disallowed type' => ['tv', 'all', true],
+            'disallowed type after allowed type' => ['screen', 'tv', false],
+            'allowed type after allowed type' => ['screen', 'all', true],
+        ];
+    }
+
+    /**
+     * @test
+     *
+     * @param string $emptyRuleMediaType
+     * @param string $mediaType
+     * @param bool $isAllowedType
+     *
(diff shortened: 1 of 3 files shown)


tests/Unit/EmogrifierTest.php
```diff
@@ -1725,6 +1725,85 @@ public function emogrifyIgnoresMediaQueryWithWhitespaceOnly()
...
+    /**
+     * @test
+     *
+     * @param string $emptyRuleMediaType
+     * @param string $mediaType
+     * @param bool $isAllowedType
+     *
+     * @dataProvider mediaTypesDataProvider
+     */
+    public function emogrifyKeepsMediaRuleAfterEmptyMediaRuleIffAllowed($emptyRuleMediaType, $mediaType, $isAllowedType)
```
oliverklee: Iff -> If

tests/Unit/EmogrifierTest.php
```diff
@@ -1725,6 +1725,85 @@ public function emogrifyIgnoresMediaQueryWithWhitespaceOnly()
...
+    {
+        $this->subject->setHtml('<html><h1></h1></html>');
+        $mediaCss = '@media ' . $mediaType . ' { h1 { color: red; } }';
+        $this->subject->setCss('@media ' . $emptyRuleMediaType . ' {} ' . $mediaCss);
+
+        $result = $this->subject->emogrify();
+
+        if ($isAllowedType) {
```
oliverklee: Please let's have two separate tests with separate data providers each for this.

JakeQZ:
I've separated the test method for keeping and not keeping media rules into two methods.

PR #538, opened 2018-04-01
[CLEANUP] Merge #528 into CssInliner

JakeQZ:
The only difference from #528 changes to Emogrifier.php are that the `require_once` fallback for non-autoload is omitted, as is the namespace path to `CssConcatenator` which is in the same namespace as `CssInliner`.


oliverklee:
Merged. Thanks, @JakeQZ! :coffee: :heart:

PR #539, opened 2018-04-02
[CLEANUP] Refactor `copyCssWithMediaToStyleNode`

JakeQZ:
Renamed `copyCssWithMediaToStyleNode` to `copyUninlineableCssToStyleNode` and
changed the final parameter from a string of CSS to the array of "uninlineable"
parsed CSS rule data returned by `parseCssRules`.

The method can now potentially also include rules other than @media rules in the
style element added to the document, and will do when the final changes for #280
are implemented.

### DIFF
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -426,11 +426,7 @@ class Emogrifier
 
         $this->removeImportantAnnotationFromAllInlineStyles($xPath);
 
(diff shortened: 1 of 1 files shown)


oliverklee:
Merged. Thanks for the beautiful refactoring, @JakeQZ!

PR #540, opened 2018-04-02
[CLEANUP] Use the Symfony CSS selector component

### LINKED ISSUE #287: Use Symfony CSS selector component
I came across this Symfony component for converting css selectors to xpath https://github.com/symfony/css-selector and wondered whether it has previously been considered?

I would expect it to be much more complete than the current Emogrifier and probably more robust too.

I have made  a simple change to use it instead of the code in method translateCssToXpath() and almost all unit tests pass. Those that fail seem to highlight bugs in the current Emogrifier code, such as https://github.com/jjriv/emogrifier/issues/286

Would all unit tests passing be sufficient assurance for the component to be used?

oliverklee:
Fixes #287

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#476](https://github.com/jjriv/emogrifier/pull/476))
 
 ### Changed
+- Use the Symfony CSS selector component
+  ([#540](https://github.com/jjriv/emogrifier/pull/540))
 
 ### Deprecated
 - Support for PHP 5.5 will be removed in Emogrifier 3.0.
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,8 @@
         "source": "https://github.com/MyIntervals/emogrifier"
     },
     "require": {
-        "php": "^5.5.0 || ~7.0.0 || ~7.1.0 || ~7.2.0"
+        "php": "^5.5.0 || ~7.0.0 || ~7.1.0 || ~7.2.0",
+        "symfony/css-selector": "^3.4.0 || ^4.0.0"
     },
     "require-dev": {
         "squizlabs/php_codesniffer": "^3.2.0",
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -855,37 +855,13 @@ class Emogrifier
      */
     private function clearAllCaches()
     {
-        $this->clearCache(static::CACHE_KEY_CSS);
-        $this->clearCache(static::CACHE_KEY_SELECTOR);
-        $this->clearCache(static::CACHE_KEY_XPATH);
-        $this->clearCache(static::CACHE_KEY_CSS_DECLARATIONS_BLOCK);
-        $this->clearCache(static::CACHE_KEY_COMBINED_STYLES);
-    }
-
-    /**
-     * Clears a single cache by key.
-     *
-     * @param int $key the cache key, must be CACHE_KEY_CSS, CACHE_KEY_SELECTOR, CACHE_KEY_XPATH
-     *                 or CACHE_KEY_CSS_DECLARATION_BLOCK
-     *
-     * @return void
-     *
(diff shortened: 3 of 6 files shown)


oliverklee:
Currently, with this change the tests for specificity fail for these cases where we have something like `:not(* + *)`:

```
// :not
':not alone as specific as universal' => ['<p class="p-1"', '*:not(* + *)', '*'],
'type & :not alone as specific as type' => ['<p class="p-1"', 'p:not(* + *)', 'p'],
'class & :not alone as specific as class' => ['<p class="p-1"', '.p-1:not(* + *)', '.p-1'],
'ID & :not alone as specific as ID' => ['<p class="p-4" id="p4"', '#p4:not(* + * + * + * + *)', '#p4'],
'2 types & 2 classes & 2 IDs & :not alone as specific as 2 types & 2 classes & 2 IDs' => [
'<span id="text"',
    'p.p-4.p-4#p4 span#text:not(* + *)',
     'p.p-4.p-4#p4 span#text'
],
```

@JakeQZ Do you think this is a bug in the CSS selector library, or a bug in the test cases?

(Now that I see it, the test cases test `:not` for a universal selector with adjacency, which semantically does not make very much sense to me. But then I'm not very deep into these edge cases.)

JakeQZ:
The `:not` tests are to confirm the presence of `:not` does not in itself increase specificity, by using a selector expression as the argument for `:not` which also does not increase specificity (i.e. combinations of the universal selector).

The Symfony CssSelector component has only limited support for `:not` - it supports only single-element selector expressions in the argument (i.e. descendent, adjacent, etc., are not supported).

`:not(*)` would not match anything, and `:not(* *)` and `:not(:not(*))` are not supported either, so I cannot think of a selector expression that is supported by Symfony that could be used in these tests without changing them slightly.

The erring tests could be rewritten to use a type that won't match in the argument of `:not` and add an additional type to the other selector so their specificities will still be the same, e.g. the following would work (though the descriptions should probably also be changed):

```
            // :not
            ':not alone as specific as universal' => ['<p class="p-1"', '*:not(html)', 'html *'],
            'type & :not alone as specific as type' => ['<p class="p-1"', 'p:not(html)', 'html p'],
            'class & :not alone as specific as class' => ['<p class="p-1"', '.p-1:not(html)', 'html .p-1'],
            'ID & :not alone as specific as ID' => ['<p class="p-4" id="p4"', '#p4:not(html)', 'html #p4'],
            '2 types & 2 classes & 2 IDs & :not alone as specific as 2 types & 2 classes & 2 IDs' => [
                '<span id="text"',
                'p.p-4.p-4#p4 span#text:not(html)',
                'html p.p-4.p-4#p4 span#text'
            ],
```

oliverklee:
@JakeQZ Thanks for the quick reply! Would you be willing to provide a small PR that changes the test cases of `Emogrifier` and `CssInliner` accordingly?

JakeQZ:
Will do...

oliverklee:
Thanks! :heart:

oliverklee:
> Was the flattening of the cache array required here?

@zoliszabo Yes, because all caches but one were used only for the CSS-selector-to-Xpath mapping. So with only one cache being left, there is no need anymore to allow several caches.

zoliszabo:
> @zoliszabo Yes, because all caches but one were used only for the CSS-selector-to-Xpath mapping. So with only one cache being left, there is no need anymore to allow several caches.

From what I see, $cssCache is still being used to store different "types" of data (e.g. precedence for selectors, combined CSS for 2 style arrays, etc.). Nevertheless, there is practically 0 chance for cache key conflicts.

oliverklee:
> From what I see, $cssCache is still being used to store different "types" of data (e.g. precedence for selectors, combined CSS for 2 style arrays, etc.). Nevertheless, there is practically 0 chance for cache key conflicts.

Oh, now I see it. I'll rework the PR.

oliverklee:
@zoliszabo Really excellent catch! I've fixed it and repushed.

PR #542, opened 2018-04-02
[CLEANUP] Extract method: copyInlineableCssToStyleNode

oliverklee:
This refactoring reduces the complexity of the process method.

### DIFF
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -383,7 +383,7 @@ class Emogrifier
             // executions, which can still flood logs/output unnecessarily. Instead, Emogrifier's error handler should
             // always throw an exception and it must be caught here and only rethrown if in debug mode.
             try {
-                // \DOMXPath::query will always return a DOMNodeList or an exception when errors are caught.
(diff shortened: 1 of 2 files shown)


src/Emogrifier.php
```diff
@@ -1106,6 +1093,32 @@ private function attributeValueIsImportant($attributeValue)
...
+     * Copies $cssRule into the style attribute of $node.
+     *
+     * Note: This method does not check whether $cssRule matches $node.
+     *
+     * @param \DOMElement $node
+     * @param string[][] $cssRule
+     *
+     * @return void
+     */
+    private function copyInlineableCssToStyleNode(\DOMElement $node, array $cssRule)
```
zoliszabo: I am not that sure of the name of the method. Maybe `copyInlineableCssToStyleAttribute` or `copyInlineableCssToNodeStyleAttribute` would be better? What do you think?

oliverklee: Oh, of course. Good catch! The "style node" does not make sense here.

I'll change it to `copyInlineableCssToStyleAttribute`.

oliverklee:
Done and repushed.

Issue #403, opened 2017-09-25
Make the tests easier to read and understand


oliverklee:

oliverklee:
Nothing more to do here.

PR #555, opened 2018-04-03
[CLEANUP] Simplify and clean up the test case (part 8)

oliverklee:
Part of #403

### DIFF
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -120,7 +120,9 @@ class CssInlinerTest extends \PHPUnit_Framework_TestCase
 
     /**
      * @test
+     *
      * @param string $html
+     *
      * @dataProvider contentWithoutHtmlTagDataProvider
      */
     public function emogrifyAddsMissingHtmlTag($html)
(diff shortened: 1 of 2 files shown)


tests/Unit/CssInlinerTest.php
```diff
@@ -1462,20 +1459,7 @@ public function addAllowedMediaTypeKeepsStylesForTheGivenMediaType()
...
      */
-    public function emogrifyAddsMissingHeadElement()
-    {
-        $this->subject->setHtml('<html></html>');
-        $this->subject->setCss('@media all { html { some-property: value; } }');
-
-        $result = $this->subject->emogrify();
-
-        static::assertContains('<head>', $result);
-    }
```
JakeQZ: Shouldn't this be retained until `CssInliner` extends `AbstractHtmlProcessor`?

tests/Unit/EmogrifierTest.php
```diff
@@ -1461,20 +1458,7 @@ public function addAllowedMediaTypeKeepsStylesForTheGivenMediaType()
...
      */
-    public function emogrifyAddsMissingHeadElement()
-    {
-        $this->subject->setHtml('<html></html>');
-        $this->subject->setCss('@media all { html { some-property: value; } }');
-
-        $result = $this->subject->emogrify();
-
-        static::assertContains('<head>', $result);
-    }
```
JakeQZ: Shouldn't this be retained as `Emogrifier` doesn't extend `AbstractHtmlProcessor`?

tests/Unit/CssInlinerTest.php
JakeQZ: Ah yes, I see now - sorry.

PR #557, opened 2018-04-03
[CLEANUP] AssertCss - whitespace around comma

JakeQZ:
Allow varying whitespace around commas (in selector list separators) in
`AssertContainsCss`, etc.

This will be required for testing that CSS rules with state-based pseudo-classes
are copied to the style element for #280.


oliverklee:
Merged. Thanks, @JakeQZ! :coffee:

PR #570, opened 2018-04-08
[FEATURE] Add CssInliner::render

oliverklee:
This is the next step toward making `CssInliner` inherit from
`AbstractHtmlProcessor`.

Part of #554


oliverklee:
> I presume at some stage `$xmlDocument` will become a class property...

Yes, it will indeed.

PR #572, opened 2018-04-08
[CLEANUP] Fix PhpStorm code inspection warnings

oliverklee:


oliverklee:
> I don't suppose PhpStorm code inspector is available via Composer/Travis,

That's correct.

PR #573, opened 2018-04-08
[TASK] Use explicitly namespaced PHP function calls

oliverklee:
This slightly improves performance.


oliverklee:
@JakeQZ I don't know of any sniff for this (which does not mean much). Please let's add a check for this.

Issue #574, opened 2018-04-08
Add additional PHPCS sniffs


JakeQZ:
Following on from #563, there are currently 6 sniffs for PHPDoc from Squiz.Commenting.FunctionComment that are currently disabled, which require 'fixes' to the PHPDoc before they can be enabled.

Also noted there, [Slevomat Coding Standard](https://github.com/slevomat/coding-standard) provides further sniffs that may be desirable, notably [SlevomatCodingStandard.Arrays.TrailingArrayComma](https://github.com/slevomat/coding-standard#slevomatcodingstandardarraystrailingarraycomma-), but there may be others that a closer review of what is available would reveal.

JakeQZ:
Also, possible additional sniffs suggested by recent PRs #572 and #573.

oliverklee:
Yes, please! (Except for the parameter alignment, which would create busywork on potentially every change to the parameters.) And we should look at the necessary changes for each additional sniff.

JakeQZ:
For the capital letter & full stop PHPDoc sniffs, we have a large number of @param and some @throws which don't have capitals or full stops (most of them it seems), so I would be inclined to continue to allow these for now rather than making a large number of changes (particularly as doing so may make maintaining Emogrifier and CssInliner in tandem more difficult).

What I'm specifically looking for is a sniff for this in the main PHPDoc comment, which Squiz.Commenting.FunctionComment doesn't seem to do...

JakeQZ:
The only sniffs I could find for capitals and full stops in the PHPDoc short description don't distinguish between methods and properties, the latter for which we want to allow incomplete sentences.  (`Generic.Commenting.DocComment.ShortNotCapital` and, via `drupal/coder` which is out-of-date and incompatible with `php_codesniffer` 3.x, `Drupal.Commenting.DocComment.ShortFullStop`.)

JakeQZ:
All of the potential sniffs alluded to here have either been added, or I was unable to find an off-the-peg sniff that would meet our requirements.  Whilst it is possible to write custom sniffs for PHPCS, the effort would probably outweigh the benefit for a relatively small project.

So I think this can be closed now.  We can always open a new issue if we come across specific sniffs that should be included.

PR #576, opened 2018-04-11
[TASK] Sniff for FQN in global function calls

JakeQZ:
Also fully-qualified the global namespace in a few function call instances that
were missed in #573.

This change makes available all Slevomat PHPCS sniffs
(https://github.com/slevomat/coding-standard), via Composer and the config XML,
should any others of them also be desired in future.

However, as Slevomat requires PHP 7, PHPCS is now skipped for the CI build with
earlier PHP versions (it only really needs to run against one PHP version).

Part of #574.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,10 +57,10 @@ script:
 
 - >
   echo;
-  if php -v | grep -Pq 'PHP\s++(?:[789]|\d{2})'; then
+  function version_gte() { test "$(printf '%s\n' "$@" | sort -n -t. -r | head -n 1)" = "$1"; };
+  if version_gte $(composer php:version) 7; then
     echo "Running PHP_CodeSniffer";
     composer ci:php:sniff;
   else
-    echo "Skipped PHP_CodeSniffer due to insufficient PHP version";
-    php -v | grep -Po 'PHP\s++[56]\.(\d++\.)*+\d++(?:-\w++)?+';
+    echo "Skipped PHP_CodeSniffer due to insufficient PHP version: $(composer php:version)";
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -57,5 +57,10 @@ script:
...
 - >
   echo;
-  echo "Running PHP_CodeSniffer";
-  composer ci:php:sniff;
+  if php -v | grep -Pq 'PHP\s++(?:[789]|\d{2})'; then
+    echo "Running PHP_CodeSniffer";
+    composer ci:php:sniff;
+  else
+    echo "Skipped PHP_CodeSniffer due to insufficient PHP version";
+    php -v | grep -Po 'PHP\s++[56]\.(\d++\.)*+\d++(?:-\w++)?+';
```
oliverklee: I propose moving this to a composer script, e.g., `php:version`. This will make the purpose of this line more obvious to the reader.

JakeQZ: Have done so.  It was not also possible to add a version comparison function as a Composer script, as Composer just tacks parameters on the end of the command, so this is in the travis.yml script.

.travis.yml
```diff
@@ -57,5 +57,10 @@ script:
 
 - >
   echo;
-  echo "Running PHP_CodeSniffer";
-  composer ci:php:sniff;
+  function version_gte() { test "$(printf '%s\n' "$@" | sort -n -t. -r | head -n 1)" = "$1"; };
+  if version_gte $(composer php:version) 7; then
```
oliverklee: :heart:

PR #582, opened 2018-05-17
Change version requirements to support PHP 7.2 and add 7.2 to the tra…

### LINKED ISSUE #580: Support PHP 7.2 in version 1.2.x
My dev environment typically uses the most up to date PHP version available, though the application we develop actually requires a minimum of PHP 5.4 currently so we have to use v1.2.0. To avoid introducing code that is not compatible, I develop with the language level set to PHP 5.4 in my IDE.

However, because the composer.json file for v.1.2.0 requires `"php": ">=5.4.0,<=7.1.99"` every time I add or update our composer dependencies, the emogrifier dependency prevents this, and I have to downgrade to PHP 7.1 in order to perform this task and revert back to PHP 7.2.

So a few of questions:

1) Is there confirmed issues with emogrifier v1.2.0 and PHP 7.2?
2) If not, can we update composer.json to remove this restriction (or at least change it to `"php": ">=5.4.0,<=7.2.99"` until compatibility with PHP 7.3 is confirmed)
3) If there are confirmed issues, would you be open to me submitting a pull request to resolve them, and releasing that as v1.2.1?

FWIW, the only reason I haven't dived into the code, yet, is because I believe there might not actually be any compatibility issues, so it would just involve updating the composer.json file accordingly. The reason I say this is because we actually have nearly 12% of our customer base (who are using a version of our software that contains emogrifier) who are using PHP 7.2 (with emails being a fairly core component in our application) and there have been no reported issues so far.

So, ultimately, if this is just a case of being cautious or unsure about compatibility and there are no known issues (I will be happy to do some more detailed testing to 100% confirm that, though) then I think we're safe to remove that requirement.

chrisdeeming:
…vis matrix. Closes #580.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ matrix:
     env: CODE_SNIFFER=yes
   - php: 7.1
     env: CODE_SNIFFER=yes
+  - php: 7.2
+    env: CODE_SNIFFER=yes
   - php: hhvm
     env: CODE_SNIFFER=yes
 
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0,<=7.1.99"
+        "php": ">=5.4.0,<=7.2.99"
     },
     "require-dev": {
         "squizlabs/php_codesniffer": "2.6.0",


oliverklee:
Merged. Thanks, @chrisdeeming! :heart: 

I'll do a new minor release 1.2.1 soon.

chrisdeeming:
Hey @oliverklee 

Just a reminder regarding a 1.2.1 release :)

Thanks,

Chris

oliverklee:
@chrisdeeming Thanks for the reminder!

I've just released version 1.2.1: https://github.com/MyIntervals/emogrifier/releases/tag/v1.2.1

PR #584, opened 2018-06-05
Make LICENSE GitHub-parseable

zoliszabo:
Generated using the "Choose a license template" button.
https://help.github.com/articles/licensing-a-repository/#applying-a-license-to-a-repository-with-an-existing-license
https://help.github.com/articles/adding-a-license-to-a-repository/


JakeQZ:
The PHP 5 build errors appear to be due to Composer 1.6.5 aborting after encountering a package that requires PHP 7, whereas Composer 1.6.3 would install the remaining packages...

zoliszabo:
Right. I also added issue https://github.com/MyIntervals/emogrifier/issues/585.

Strange is that on my localhost I have the same problem with Composer 1.6.3...

oliverklee:
@zoliszabo Could you please rebase now that the build is fixed?

Issue #585, opened 2018-06-05
composer update fails for PHP version < 7


zoliszabo:
It looks like Travis CI builds fail now, because `composer update` won't install packages anymore for PHP versions smaller than 7.0.

Log of the latest successful run of the Travis build:
https://travis-ci.org/MyIntervals/emogrifier/jobs/365757072#L516

Currently failing build log:
https://travis-ci.org/MyIntervals/emogrifier/jobs/388365600#L506

Please observe the difference how now it says `No dependencies installed. Try running composer install or update.`, even if `slevomat/coding-standard` could not be installed in both cases.

Not sure what the issue might be, the only difference seems to be the version of composer (1.6.3 vs 1.6.5), but I have the same problem on my localhost even with 1.6.3.

JakeQZ:
I'm also seeing it fail with 1.6.3 locally.  I think it had succeeded before because it was not starting with a clean system, and the dependencies and autoload files were already present from a previous build, e.g. in https://travis-ci.org/MyIntervals/emogrifier/jobs/364877259 before #576 we see

> Nothing to install or update
> Writing lock file
> Generating autoload files

whereas in https://travis-ci.org/MyIntervals/emogrifier/jobs/365752325 after that change we don't see this in the composer output, but instead the error message.  In both cases the list of packages follows, but in the latter I think it is only because they were _already_ installed before Composer ran.

I'm trying to find a way to tell composer that a package is only required for (say) PHP>=7, but to no avail so far.  There is `--ignore-platform-reqs` (https://github.com/composer/composer/pull/3320) but that is no good as it causes the PHP version to be completely ignored, and for every single package.  Maybe we would need to provide a separate composer.json for PHP 5 in Travis but that would be an ugly solution.

zoliszabo:
> I'm also seeing it fail with 1.6.3 locally. I think it had succeeded before because it was not starting with a clean system, and the dependencies and autoload files were already present from a previous build, e.g. in https://travis-ci.org/MyIntervals/emogrifier/jobs/364877259 before #576 we see
>
>   > Nothing to install or update
>   > Writing lock file
>   > Generating autoload files
>
>whereas in https://travis-ci.org/MyIntervals/emogrifier/jobs/365752325 after that change we don't see this in the composer output, but instead the error message. In both cases the list of packages follows, but in the latter I think it is only because they were already installed before Composer ran.

You're right.

[...]

PR #588, opened 2018-06-26
Escape hyphens in regular expressions

kozaktomas:
Hyphens need to be escaped in regular expressions. Because of that tests will be falling in PHP 7.3

Error from tests:
```
Warning: preg_replace(): Compilation failed: invalid range in character class at offset 39 in /builds/php/mailing/vendor/pelago/emogrifier/Classes/Emogrifier.php on line 1507
```

See:
https://stackoverflow.com/questions/26208622/preg-match-compilation-failed-invalid-range-in-character-class-at-offset-15?rq=1

### DIFF
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -179,7 +179,7 @@ class Emogrifier
         // classes, attributes, pseudo-classes (not pseudo-elements) except `:not`: worth 100
         '(?:\\.|\\[|(?<!:):(?!not\\())' => 100,
         // elements (not attribute values or `:not`), pseudo-elements: worth 1
-        '(?:(?<![="\':\\w-])|::)' => 1
+        '(?:(?<![="\':\\w\\-])|::)' => 1
     ];
 
     /**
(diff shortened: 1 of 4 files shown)


src/Emogrifier.php
```diff
@@ -179,7 +179,7 @@ class Emogrifier
         // classes, attributes, pseudo-classes (not pseudo-elements) except `:not`: worth 100
         '(?:\\.|\\[|(?<!:):(?!not\\())' => 100,
         // elements (not attribute values or `:not`), pseudo-elements: worth 1
-        '(?:(?<![="\':\\w-])|::)' => 1
+        '(?:(?<![="\':\\w\-])|::)' => 1
```
oliverklee: Could you please use two backslashes for escaping the hyphen (also in the other places) so it's clear to the reader that this is escaping for the regexp, not for the PHP string?

oliverklee:
We'll need to fix #585 first so the build will be green again.

kozaktomas:
Hi, my source branch is now #589.

- backslashes added 
- changelog updated (#589 also added)
- amend test getCssNeedleRegExpNotEscapesNonSpecialCharacters()
    - https://wiki.php.net/rfc/pcre2-migration 
    - The 'X' modifier is the default behavior in PCRE2
    - `#` - `\preg_quote` now escapes as `\#` - no issue - just failing test

oliverklee:
@kozaktomas Could you please rebase now that the build is fixed?

kozaktomas:
done

PR #589, opened 2018-06-27
Fix Travis for PHP 5.x

kozaktomas:
- Install `slevomat/coding-standard` only for PHP 7.x in Travis.
- Delete `Squiz.PHP.ForbiddenFunctions` (`Generic.PHP.ForbiddenFunctions`)
- Enable installation on `~7.3.0`

Releated issue: #585

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,8 @@ script:
   echo;
   function version_gte() { test "$(printf '%s\n' "$@" | sort -n -t. -r | head -n 1)" = "$1"; };
   if version_gte $(composer php:version) 7; then
+    echo "Installing slevomat/coding-standard only for PHP 7.x";
+    composer require --dev slevomat/coding-standard ^4.0
     echo "Running PHP_CodeSniffer";
     composer ci:php:sniff;
   else
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,6 @@
     },
     "require-dev": {
         "squizlabs/php_codesniffer": "^3.2.0",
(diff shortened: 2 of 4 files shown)


kozaktomas:
wow. Dependency hell!

composer.json
```diff
@@ -37,12 +37,11 @@
         "source": "https://github.com/MyIntervals/emogrifier"
     },
     "require": {
-        "php": "^5.5.0 || ~7.0.0 || ~7.1.0 || ~7.2.0",
+        "php": "^5.5.0 || ~7.0.0 || ~7.1.0 || ~7.2.0 || ~7.3.0",
```
oliverklee: Could you please undo the change? I'd like to keep a "only one issue per PR" policy so that git blame is more meaningful, and so that we can revert individual changes more easily.

.travis.yml
```diff
@@ -59,6 +59,8 @@ script:
   echo;
   function version_gte() { test "$(printf '%s\n' "$@" | sort -n -t. -r | head -n 1)" = "$1"; };
   if version_gte $(composer php:version) 7; then
+    echo "Installing slevomat/coding-standard only for PHP 7.x";
+    composer require --dev slevomat/coding-standard v4.0.0
```
oliverklee: Why `v4.0.0` instead of `^4.0` as in the `composer.json`?

kozaktomas:
Should be fixed.

JakeQZ:
Thanks @kozaktomas for taking this on - it was on my 'to do' list but I haven't had time.

PR #596, opened 2018-10-14
[CLEANUP] Simplify the dependencies installation

oliverklee:
`composer update` by default installs the latest dependencies. So
we only need a parameter if we want to install the oldest versions of
the dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ cache:
 
 env:
   matrix:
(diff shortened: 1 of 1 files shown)


oliverklee:
> Looks neat, however the echo command would now give the wrong information when installing the lowest dependencies.

Good catch! Fixed and repushed.

PR #602, opened 2018-10-20
[BUGFIX] Increase PHP_CodeSniffer version for PHP 7.3 compatibility

chrisdeeming:


chrisdeeming:
@oliverklee GitHub is having a "case of the Mondays" today.

https://status.github.com/messages

PR #607, opened 2018-10-23
[FEATURE] Add a Composer script for PHP CS Fixer

oliverklee:
As the fixer should fix the files in multiple directories, the
`--config` argument is required, which removes the benefits of
storing the configuration in the default `.php_cs.dist`.

Also autoformat the code with the fixer.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -98,6 +98,12 @@ If you add new methods or fields, please add proper PHPDoc for the new
 methods/fields. Please use grammatically correct, complete sentences in the
 code documentation.
 
+You can autoformat your code using the following command:
+
+```shell
+composer php:fix
+```
+
 
 ## Git commits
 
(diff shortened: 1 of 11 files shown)


oliverklee:
> I see that the CodeSniffer rule Squiz.Arrays.ArrayDeclaration.NoCommaAfterLast seems to have missed a few instances (where elements have keys declared or there is only one element).

AFAIK, arrays that have the square brackets on the same line should not be affected by this rule. Are those the instances you've seen?

JakeQZ:
> AFAIK, arrays that have the square brackets on the same line should not be affected by this rule. Are those the instances you've seen?

I was referring to the instances changed by autoformatting in this PR, which are all multiline arrays.  But see also https://github.com/MyIntervals/emogrifier/pull/577 - this picked up most instances but should also have picked up those changed here, so it looks like that PHP_CodeSniffer rule doesn't detect a multiline array in some cases.

PR #608, opened 2018-10-23
[BUGFIX] Consistently use the Composer dependencies preference

oliverklee:
The switch for installing the oldest/newest dependencies should also
be applied when installing slevomat/coding-standard.

PR #609, opened 2018-10-23
[CLEANUP] Move the prefer-stable/dist to `composer.json`

oliverklee:
This reduces the amount of typing required for Composer updates.

This does not affect projects that use Emogrifier as a library,
but only stand-alone installations of Emogrifier (usually for
development).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
 - >
   echo;
   echo "Updating the dependencies";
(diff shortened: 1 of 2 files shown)


oliverklee:
This PR will (and should) conflict with #608. No matter which PR gets merged first, I will need to update the other one.

oliverklee:
Rebased, resolved the conflicts and repushed. Should be ready for review now.

PR #613, opened 2018-10-24
[TASK] Make `CssInliner::setHtml` private

oliverklee:
This is another step towards making `CssInline` an `HtmlProcessor`.

### DIFF
--- a/src/Emogrifier/CssInliner.php
+++ b/src/Emogrifier/CssInliner.php
@@ -201,7 +201,7 @@ class CssInliner
      *
      * @return void
      */
-    public function setHtml($html)
+    private function setHtml($html)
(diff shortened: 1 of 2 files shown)


oliverklee:
> In the abstract class it is set by the constructor, so this method will ultimately be removed, I presume.

Yes, after some more refactoring.

JakeQZ:
> Waiting to be queued

Travis seems to be stuck...

oliverklee:
> Travis seems to be stuck...

No, it's still building. (There are quite a few builds in the queue.) You can click on the "Details" link in the checks list to see the build status.

oliverklee:
Now everything is green.

PR #615, opened 2018-10-24
[CLEANUP] Use double quotes for strings where appropriate

oliverklee:
Thus, we can avoid quoting single quotes and using the LF constant.


oliverklee:
Wait, I've spotted a few more places to change …

oliverklee:
Done and repushed.

PR #616, opened 2018-10-24
[CLEANUP] Rename $this->xmlDocument to $this->domDocument

### LINKED ISSUE #610: Rename `$this->xmlDocument`?
`$this->xmlDocument` is an instance of `DOMDocument`, and semantically, it's an HTML document, not an XML document. Should we rename it to `htmlDocument` or `domDocument`?

@zoliszabo, @JakeQZ, what do you think?

I myself would opt for renaming to `domDocument`, as I think semantic wording is a good thing in code.

oliverklee:
Closes #610


src/Emogrifier.php
```diff
@@ -339,47 +339,47 @@ public function emogrify()
...
-        $bodyNodeHtml = $processedXmlDocument->saveHTML($this->getBodyElement($processedXmlDocument));
+        $processedDomDocument = $this->process($this->domDocument);
+        $bodyNodeHtml = $processedDomDocument->saveHTML($this->getBodyElement($processedDomDocument));
 
         return \str_replace(['<body>', '</body>'], '', $bodyNodeHtml);
     }
 
     /**
-     * Creates an XML document from $this->html.
+     * Creates an DOM document from $this->html.
```
zoliszabo: How is it correct: "a**n** DOM ..." or "a DOM document"?

oliverklee: Oh, this needs to be "a DOM …". Good catch! I'll repush in a minute.

oliverklee: Fixed and repushed.

PR #626, opened 2018-10-25
[TASK] Pass the HTML around for processing

oliverklee:
This is another step towards making the `CssInliner`
an `HtmlProcessor`, preparing for the DOM document
getting created directly when the HTML is set.

Also make the corresponding methods in `AbstractHtmlProcessor`,
`CssInliner` and `Emogrifier` more similar to each other.

### DIFF
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -323,7 +323,9 @@ class Emogrifier
      */
     public function emogrify()
     {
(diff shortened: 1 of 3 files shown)


oliverklee:
This PR will conflict with #627. After any one of those two has been merged, I'll take care of rebasing the other one.

PR #627, opened 2018-10-25
[TASK] Remove unprocessable tags from the DOM, not from the raw HTML

oliverklee:
Required for #554.

### DIFF
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -394,7 +394,7 @@ class Emogrifier
         $domDocument->strictErrorChecking = false;
         $domDocument->formatOutput = true;
         $libXmlState = \libxml_use_internal_errors(true);
-        $domDocument->loadHTML($this->unifyHtml($html));
+        $domDocument->loadHTML($this->prepareHtmlForDomConversion($html));
         \libxml_clear_errors();
         \libxml_use_internal_errors($libXmlState);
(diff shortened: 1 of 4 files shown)


oliverklee:
This PR will conflict with #626. After any one of those two has been merged, I'll take care of rebasing the other one.

oliverklee:
Rebased and repushed.

src/Emogrifier.php
```diff
@@ -412,8 +412,7 @@ private function createRawDomDocument($html)
      */
```
zoliszabo: The description of this method is not accurate anymore. It should be something like:
```
     * Returns the HTML with added document type and Content-Type meta tag if needed.
```
Given this, I am not sure that `unifyHtml` is the best name anymore for this method. What do you think?

oliverklee: Good catch! I've updated the descriptions and will repush after we've come to a conclusion about the name.

oliverklee: The name should communicate that the method prepares the HTML in a way that it is safe for creating a DOM document from.

What do you think about the name `prepareHtmlForDomConversion`?

zoliszabo: I like it!

oliverklee: Great! Renamed and repushed.

PR #631, opened 2018-11-07
[TASK] Consistently use the render() method

oliverklee:
- Add a (protected) `render()` method to the `Emogrifier` class.
- Use the `render()` method where it can be used.
- Change the tests in `CssInlinerTest` to use `render()` instead of
  `emogrify` if they actually are about the rendering, not the
  emogrification.
- Streamline some PHPDoc annotations.

This change work towards splitting `emogrify()` into `inlineCss()` and
`render()`in `CssInliner`.

Part of #554.

### DIFF
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -317,6 +317,16 @@ class Emogrifier
     }
 
     /**
+     * Renders the normalized and processed HTML.
+     *
+     * @return string
+     */
+    protected function render()
+    {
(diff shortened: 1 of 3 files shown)


oliverklee:
> Should the corresponding tests in EmogrifierTest also be changed to test render rather than emogrify or will this be covered by a separate PR?

No, as I don't want to change the public API of the `Emogrifier` class now that it will be replaced with `CssInliner`, and `Emogrifier::render()` is internal.

PR #640, opened 2018-12-07
[TASK] Prepare the 2.1.0 release

### LINKED ISSUE #639: Release version 2.1.0?
In https://github.com/MyIntervals/emogrifier/pull/638#issuecomment-444890570, @TimWolla asked whether a 2.1.0 release really soon would be possible.

Originally, I had planned to finished the `CssInliner` class first and make the final API in place before people start using it first.

Instead, I could mark the `CssInliner` class as `@internal` for the time being and add a comment that the API is still apt to change and that people should not use the class yet. Then we could release version 2.1.0 really soon.

@JakeQZ @zoliszabo @jjriv What do you think?

oliverklee:
Closes #639

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,19 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
-
 ## x.y.z
+
+### Added
+
+### Changed
+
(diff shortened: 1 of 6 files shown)


oliverklee:
@JakeQZ Thanks for the review comments! I've updated the PR accordingly and repushed.

PR #641, opened 2018-12-09
[TASK] Add test for automatically adding a missing DOCTYPE

oliverklee:
This bug fix slipped in with one of the latest refactorings. Now we
also have a unit test for it.

### DIFF
--- a/tests/Unit/EmogrifierTest.php
+++ b/tests/Unit/EmogrifierTest.php
@@ -268,6 +268,18 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @test
+     */
+    public function addsMissingHtml5DocumentType()
+    {
+        $this->subject->setHtml('<html></html>');
+
+        $result = $this->subject->emogrify();
+
+        static::assertContains('<!DOCTYPE html>', $result);
+    }
+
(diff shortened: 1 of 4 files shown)


oliverklee:
I've had this branch on my computer for some time now and would like to have it out of the way.

oliverklee:
It looks like something still is missing. I'll have a look.

tests/Unit/EmogrifierTest.php
```diff
@@ -267,6 +267,18 @@ public function emogrifyBodyContentKeepsSpecialCharacters($codeNotToBeChanged)
...
     }
 
+    /**
+     * @test
+     */
+    public function addsMissingHtml5DocumentType()
+    {
+        $this->subject->setHtml('<html></html>');
+
+        $result = $this->subject->emogrifyBodyContent();
```
zoliszabo: As far as I can see, the only issue (the cause of failed test) is that this line should read `$result = $this->subject->emogrify()`.

oliverklee: Oh, of course. D'oh!

oliverklee:
Fixed and repushed.

zoliszabo:
👍

PR #642, opened 2018-12-09
[BUGFIX] Keep the `style` element the `head`

### LINKED ISSUE #634: Stylesheet in body seems to be ignored by gmail
I'm currently working on a responsive email design. I don't get the media queries to work correct in gmail while using v2.0 of emogrifier. If I use the older version (1.2.1), it seems to work fine. 

I think it's related to (https://github.com/MyIntervals/emogrifier/pull/396), since all style blocks are automatically moved to the body. I think gmail automatically ignores those stylesheets. 

Is anyone aware of this issue? Or am I missing something here. I would like to use the latest version of this lib if possible.

Thanks for you help!

oliverklee:
Fixes #634

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
+- Keep the `style` element the `head`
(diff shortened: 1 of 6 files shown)


oliverklee:
After this PR has been merged, I'll create a follow-up PR that moves the "let's ensure that there is a HEAD element" to a separate, well-named method.

PR #643, opened 2018-12-09
[FEATURE] Add PHP 7.3 support (v1.2.x)

TimWolla:
see https://github.com/MyIntervals/emogrifier/issues/639#issuecomment-445455840

I believe this needs tagging only and is good to go afterwards.


oliverklee:
Merged. Thanks, @TimWolla! :heart:

Issue #646, opened 2018-12-30
Emogrifier turns {} into %7B and %7D


Johnny99211:
I'm currently setting up an email template which has two parts: The HTML and the CSS. To put everything together and merge the HTML with the CSS I'm using the PHP Emogrifier.

Now I've found out that when I emogrify the two elements that there is a problem with the uft-8 encoding.

All works like Ä, Ü, Ö are correctly displayed but when I put a link like this here into the content, the utf-8 encoding don't works for this element:

Before emogrify:

> `<a href="{password_reset_link}" id="button">Passwort zurücksetzten</a>`

After emogrify:

>     <a href="%7Bpassword_reset_link%7D" id="button" style="color: #69f0ae; text-decoration: none; font-size: 20px; text-align: center; display: block; line-height: 1.7em !important; padding: .3em 1em;
>     > border: 2px solid #69f0ae; border-radius: 3px; -webkit-border-radius:
>     > 3px; -moz-border-radius: 3px; width: 40%; margin: auto; cursor:
>     > pointer !important;">Passwort zurücksetzten</a>

So as you can see the href is broken which is set with a placeholder which gets replaced with the correct link after the emogrify process:

    %7Bpassword_reset_link%7D

zoliszabo:
_This is not a UTF-8 encoding issue, but the correct representation (encoding) of a href attribute._

Depending on how you replace your template variables (placeholders), one quick solution would be to also encode the placeholder before the replace:

```php
str_replace(rawurlencode('{placeholder}'), 'https://any.url.here');
```

Johnny99211:
This is a bug in the emogrifier. I'm replacing the placeholder after the emogrify process but the placeholder is broken after merge css to the html with the help of emogrifier: https://github.com/MyIntervals/emogrifier/issues/98

Johnny99211:
I've temporarily fixed it with this workaround after the emogrify process but in my eyes this is a bug which needs to be fixed:

```
//Apply CSS styles inline for picky email clients.
try {
	$emogrifier = new Emogrifier( $message, $css );
	$message    = $emogrifier->emogrify();
} catch ( Exception $e ) {
	$logger = wc_get_logger();
	$logger->error( $e->getMessage(), array( 'source' => 'emogrifier' ) );
}
$message = str_replace( array( '%7B', '%7D' ), array( '{', '}' ), $message );
```

zoliszabo:
[...]

PR #647, opened 2019-01-02
[TASK] Add test for keeping curly braces in text nodes

oliverklee:
The new test shows that issue #646 only affects attribute values, not text
nodes.

### DIFF
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -258,6 +258,7 @@ class CssInlinerTest extends \PHPUnit_Framework_TestCase
             'template markers with dollar signs & square brackets' => ['$[USER:NAME]$'],
             'UTF-8 umlauts' => ['Küss die Hand, schöne Frau.'],
             'HTML entities' => ['a &amp; b &gt; c'],
+            'curly braces' => ['{Happy new year!}'],
         ];
     }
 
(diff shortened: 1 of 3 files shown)


oliverklee:
> I wonder if a similar test (with the same data provider) should also be added for text-like attributes, such as alt attribute of img or title attribute?

Yes, but for that to work, we'll need a fix for #646 (i.e., the tests for this should be part of the fix for #646).

JakeQZ:
> Yes, but for that to work, we'll need a fix for #646

I was presuming that only certain attributes such as `href` are affected by #646, and others such as `alt` and `title` would not be.

oliverklee:
> I was presuming that only certain attributes such as href are affected by #646, and others such as alt and title would not be.

Actually, I'm not quite sure. Let's test this!

oliverklee:
(after thinking about my assumptions)

zoliszabo:
As far as I have seen and tested only URI-specific attribute values are affected: `a`'s `href`, `link`'s `href`, `base`'s `href`, `script`'s `src`, `form`'s `action` and `iframe`'s `src`. There are a couple more such attributes, but those do not seem to be encoded by `DOMDocument`. Of course, some of the above are rarely or never used in emails, but still...
URIs from inline style declarations, such as `style="background-image:url('{url}')` seem to be _not_ affected.

Put together a test:
https://github.com/MyIntervals/emogrifier/blob/ba33460c5ae39e18ad39f6d9a09124c4ad4b1b7c/tests/Unit/EmogrifierTest.php#L287
The first 6 attributes are failing, the others are not (normally they should, but apparently DOMDocument doesn't handle them well).

PR #650, opened 2019-02-03
[BUGFIX] Remove all ‘unprocessable’ tags.

JakeQZ:
Not all ‘unprocessable’ tags (`<wbr>` by default) were being removed, because a
‘live’ NodeList was being iterated.  This was a regression as a result of #627.

Also modified related tests to include a `<body>` element in the input HTML.
Prior to #627, `removeUnprocessableTags()` did not behave as documented – if
elements had content it would still remove the tags (though not the content).
The tests did not pick this up because `DOMDocument::loadHTML()` creates a
`<body>` element if there isn’t one, and if there is text content (at the start)
not inside any element, wraps it in a `<p>` element
(https://secure.php.net/manual/en/domdocument.loadhtml.php#88864).  (It does not
create a `<p>` element if there is already a `<body>` element.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
+- Remove all ‘unprocessable’ (e.g. `<wbr>`) tags
+  ([#650](https://github.com/MyIntervals/emogrifier/pull/650))
 - Correct translated xpath of `:nth-child` selector
   ([#648](https://github.com/MyIntervals/emogrifier/pull/648))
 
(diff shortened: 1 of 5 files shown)


JakeQZ:
An observation arises.  The `<wbr>` tag causes issues because DOMDocument does not know it is self-closing (void).  Several other elements also suffer the same fate: `<command>`, `<embed>`, `<keygen>`, `<source>`, `<track>` (from https://bugs.php.net/bug.php?id=73175).  Should these all by default be removed from the input?  Or should we perhaps support them by simply stripping the erroneous closing tags from the output (which I believe is all that is required)?

@oliverklee, @zoliszabo, what do you think?

oliverklee:
> Or should we perhaps support them by simply stripping the erroneous closing tags from the output (which I believe is all that is required)?

I'm for this approach.

JakeQZ:
> > Or should we perhaps support them by simply stripping the erroneous closing tags from the output (which I believe is all that is required)?
> 
> I'm for this approach.

Also required is those tags being in XML self-closing format in the input (i.e. `<wbr/>` rather than `<wbr>`).  See PR #651…

PR #651, opened 2019-02-13
[BUGFIX] `AbstractHtmlProcessor`: HTML5 void tags

JakeQZ:
Added support to `AbstractHtmlProcessor` for HTML5 self-closing tags not
recognized as such by PHP’s `DOMDocument` implementation.  In effect this is a
workaround for the issue reported in https://bugs.php.net/bug.php?id=73175.

Affected tags require a self-closing slash in the HTML input to `DOMDocument`
(e.g. `<wbr/>` rather than `<wbr>`), and their invalid corresponding closing tag
(e.g. `</wbr>`) removing from its HTML output.

Follows from discussion in #650.

### DIFF
--- a/src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
+++ b/src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
@@ -26,7 +26,9 @@ abstract class AbstractHtmlProcessor
     /**
      * @var string Regular expression part to match tag names that PHP's DOMDocument implementation is not aware are
      *      self-closing.  These are mostly HTML5 elements, but for completeness <command> (obsolete) and <keygen>
-     *      (deprecated) from the list at https://bugs.php.net/bug.php?id=73175 are also included.
+     *      (deprecated) are also included.
+     *
+     * @see https://bugs.php.net/bug.php?id=73175
      */
     const PHP_UNRECOGNIZED_VOID_TAGNAME_MATCHER = '(?:command|embed|keygen|source|track|wbr)';
 
@@ -81,7 +83,9 @@ abstract class AbstractHtmlProcessor
      */
     public function render()
     {
-        return $this->renderDomDocument();
+        $htmlWithPossibleErroneousClosingTags = $this->domDocument->saveHTML();
+
+        return $this->removeSelfClosingTagsClosingTags($htmlWithPossibleErroneousClosingTags);
     }
 
     /**
@@ -91,25 +95,22 @@ abstract class AbstractHtmlProcessor
      */
     public function renderBodyContent()
     {
-        $bodyNodeHtml = $this->renderDomDocument($this->getBodyElement());
+        $htmlWithPossibleErroneousClosingTags = $this->domDocument->saveHTML($this->getBodyElement());
+        $bodyNodeHtml = $this->removeSelfClosingTagsClosingTags($htmlWithPossibleErroneousClosingTags);
 
         return \str_replace(['<body>', '</body>'], '', $bodyNodeHtml);
     }
 
     /**
-     * Renders the DOMDocument as HTML, eliminating any invalid closing tags for void elements.
+     * Eliminates any invalid closing tags for void elements from the given HTML.
      *
-     * @param \DOMNode|null $node Optional parameter to output a subset of the document.
+     * @param string $html
      *
      * @return string
      */
-    protected function renderDomDocument(\DOMNode $node = null)
+    private function removeSelfClosingTagsClosingTags($html)
     {
-        return \preg_replace(
-            '%</' . static::PHP_UNRECOGNIZED_VOID_TAGNAME_MATCHER . '>%',
-            '',
-            $this->domDocument->saveHTML($node)
-        );
+        return \preg_replace('%</' . static::PHP_UNRECOGNIZED_VOID_TAGNAME_MATCHER . '>%', '', $html);
     }
 
     /**
@@ -170,7 +171,6 @@ abstract class AbstractHtmlProcessor
     private function prepareHtmlForDomConversion($html)
     {
         $htmlWithSelfClosingSlashes = $this->ensurePhpUnrecognizedSelfClosingTagsAreXml($html);
-
         $htmlWithDocumentType = $this->ensureDocumentType($htmlWithSelfClosingSlashes);
 
         return $this->addContentTypeMetaTag($htmlWithDocumentType);
--- a/src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
+++ b/src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
@@ -24,6 +24,13 @@ abstract class AbstractHtmlProcessor
     const CONTENT_TYPE_META_TAG = '<meta http-equiv="Content-Type" content="text/html; charset=utf-8">';
 
     /**
+     * @var string Regular expression part to match tag names that PHP's DOMDocument implementation is not aware are
+     *      self-closing.  These are mostly HTML5 elements, but for completeness <command> (obsolete) and <keygen>
+     *      (deprecated) from the list at https://bugs.php.net/bug.php?id=73175 are also included.
+     */
+    const PHP_UNRECOGNIZED_VOID_TAGNAME_MATCHER = '(?:command|embed|keygen|source|track|wbr)';
+
+    /**
      * @var \DOMDocument
      */
     protected $domDocument = null;
@@ -74,7 +81,7 @@ abstract class AbstractHtmlProcessor
      */
     public function render()
     {
-        return $this->domDocument->saveHTML();
+        return $this->renderDomDocument();
     }
 
     /**
@@ -84,9 +91,25 @@ abstract class AbstractHtmlProcessor
      */
     public function renderBodyContent()
     {
-        $bodyNodeHtml = $this->domDocument->saveHTML($this->getBodyElement());
+        $bodyNodeHtml = $this->renderDomDocument($this->getBodyElement());
 
         return \str_replace(['<body>', '</body>'], '', $bodyNodeHtml);
+    }
+
+    /**
+     * Renders the DOMDocument as HTML, eliminating any invalid closing tags for void elements.
+     *
+     * @param \DOMNode|null $node Optional parameter to output a subset of the document.
+     *
+     * @return string
+     */
+    protected function renderDomDocument(\DOMNode $node = null)
+    {
+        return \preg_replace(
+            '%</' . static::PHP_UNRECOGNIZED_VOID_TAGNAME_MATCHER . '>%',
+            '',
+            $this->domDocument->saveHTML($node)
+        );
     }
 
     /**
@@ -137,7 +160,7 @@ abstract class AbstractHtmlProcessor
     }
 
     /**
-     * Returns the HTML with added document type and Content-Type meta tag if needed,
+     * Returns the HTML with added document type, Content-Type meta tag, and self-closing slashes, if needed,
      * ensuring that the HTML will be good for creating a DOM document from it.
      *
      * @param string $html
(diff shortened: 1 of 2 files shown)


JakeQZ:
I'll make corresponding changes to `Emogrifier` and `CssInliner` once the changes here are approved.

JakeQZ:
A quick file compare shows there are only two significant differences between the code in `AbstractHtmlProcessor` and that currently duplicated in `CssInliner`, so it should be relatively straightforward now to make the latter inherit the former.  As such, it may be better to do so before committing this change, while the file diffs don't have any 'noise', and avoiding duplicating the changes here in `CssInliner`.

The differences are both in `createRawDomDocument`:
```
$domDocument->encoding = 'UTF-8';
// ...
$domDocument->normalizeDocument();
```
@oliverklee Is there any particular reason these lines are not currently in `AbstractHtmlProcessor`?

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -23,6 +23,13 @@ abstract class AbstractHtmlProcessor
      */
     const CONTENT_TYPE_META_TAG = '<meta http-equiv="Content-Type" content="text/html; charset=utf-8">';
 
+    /**
+     * @var string Regular expression part to match tag names that PHP's DOMDocument implementation is not aware are
+     *      self-closing.  These are mostly HTML5 elements, but for completeness <command> (obsolete) and <keygen>
+     *      (deprecated) from the list at https://bugs.php.net/bug.php?id=73175 are also included.
```
oliverklee: I propose we move the URL to an `@see` annotation.

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -84,11 +91,27 @@ public function render()
...
     }
 
+    /**
+     * Renders the DOMDocument as HTML, eliminating any invalid closing tags for void elements.
+     *
+     * @param \DOMNode|null $node Optional parameter to output a subset of the document.
+     *
+     * @return string
+     */
+    protected function renderDomDocument(\DOMNode $node = null)
```
oliverklee: Please let's split this into two functions instead of having kind of a feature switch with the parameter (which is a pattern to be avoided). Or maybe we can move the `preg_replace` into a private method that is called both from `renderBodyContent` and `render`.

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -146,7 +169,9 @@ private function createRawDomDocument($html)
      */
     private function prepareHtmlForDomConversion($html)
     {
-        $htmlWithDocumentType = $this->ensureDocumentType($html);
+        $htmlWithSelfClosingSlashes = $this->ensurePhpUnrecognizedSelfClosingTagsAreXml($html);
+
```
oliverklee: Why the blank line?

tests/Unit/Emogrifier/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -302,36 +302,138 @@ public function notAddsSecondContentTypeMetaTag()
...
+    {
+        $documentTypeDatasets = $this->documentTypeDataProvider();
+        $selfClosingTagDatasets = $this->selfClosingTagDataProvider();
+        $datasets = [];
+        foreach ($documentTypeDatasets as $documentTypeDatasetName => $documentTypeDataset) {
+            foreach ($selfClosingTagDatasets as $selfClosingTagDatasetName => $selfClosingTagDataset) {
+                $datasets[$documentTypeDatasetName . ' & ' . $selfClosingTagDatasetName]
+                    = \array_merge($documentTypeDataset, $selfClosingTagDataset);
+            }
+        }
```
oliverklee: That is quite a lot of logic for unit tests. Can we simplify this?

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
JakeQZ: This would just be changing the comment to `... the list at {@see https://bugs.php.net/bug.php?id=73175} are also included.`, right?

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
JakeQZ: The method as proposed is a 'fixed' version of [`DOMDocument::saveHTML()`](https://secure.php.net/manual/en/domdocument.savehtml.php) which has the same method signature and feature switch parameter.  The regex replacement could be moved to a separate method such as `removeSelfClosingTagsClosingTags()`, but without `renderDomDocument()`, both `render` and `renderBodyContent` would have to call `removeSelfClosingTagsClosingTags()` on the result of `DOMDocument::saveHTML()` which to me seems like a duplication, albeit a very small one.  What do you think?

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
JakeQZ: Because there was previously a blank line between the other two statements.  But I guess that is because the latter is the final `return` statement?  I'll remove the blank line.

tests/Unit/Emogrifier/HtmlProcessor/AbstractHtmlProcessorTest.php
JakeQZ: Simplest would be with the data copied and pasted, but that would be least maintainable and most error-prone, which is what I'm trying to avoid.

Would adding an intermediate helper method such as `joinDatasets()` make this clearer?  Then we would have
```
    /**
     * @return string[][]
     */
    public function documentTypeAndSelfClosingTagDataProvider()
    {
        return static::joinDatasets($this->documentTypeDataProvider(), $this->selfClosingTagDataProvider());
    }
```
It also occurs to me that the tests should cover each document type with at least one self-closing tag, and each self-closing tag with at least one document type, but don't necessarily need to cover all possible combinations...

JakeQZ:
I've replied with some comments for clarification.

> Also, to me this looks more like a bug fix than a feature. What do you think?

I guess it is a bug fix even if it's a workaround for a bug in a 3rd party library.  Though it's a 'feature' in the sense that we're adding support for some HTML5 elements.  So I'm not sure.

If changing "[FEATURE]" to "[BUGFIX]" in the title, I suspect you would need to ensure you repeat this change when doing the "squash and merge" as I suspect the default change description comes from the first commit rather than the PR (and doing `git commit --amend` and force-pushing would probably orphan the review comments).

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
oliverklee: Almost. ;-) The `@see some-url` annotation should be in its own line and not be embedded in a sentence.

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
oliverklee: Yes, that was the reason.

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
oliverklee: Some "structural duplication" is okay, I think. Generally, I prefer this approach:

```
function a1() {
  doSomething;
  sharedParts();
}

function a2() {
  doSomethingElse;
  sharedParts();
}

function sharedParts() {
  …
}
```

over this:

```
function a($someSwitch) {
  if ($someSwitch) {
    doSomething;
  } else {
    doSomethingElse;
  }

  sharedStuff;
}
```

tests/Unit/Emogrifier/HtmlProcessor/AbstractHtmlProcessorTest.php
oliverklee: > Would adding an intermediate helper method such as joinDatasets() make this clearer? 

Yes, I think so.

JakeQZ:
Ok, I've made the changes as agreed.  Still not sure if its a '[BUGFIX]' or '[FEATURE]'...

oliverklee:
Thanks, I'll have a look at the new PR.

> Still not sure if its a '[BUGFIX]' or '[FEATURE]'...

I'd prefer `[BUGFIX]`, but can also live with `[FEATURE]`.

JakeQZ:
> I'd prefer [BUGFIX], but can also live with [FEATURE].

You're welcome to change it when doing the squash-and-merge, if you like.

JakeQZ:
> Still not sure if its a '[BUGFIX]' or '[FEATURE]'...

On another project, I've just marked adding the selector for `input[type="tel"]` to a CSS rule for text type inputs as a bug-fix (in that particular context, there's no way it's anything else: the telephone number input is simply not styled correctly), so I am now persuaded that this a bug-fix too, as it's a very similar sort of thing...

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
oliverklee: We still have the feature switch here (kind of). I propose we have a `removeUnrecognizedClosingTags(string $html): string` (or with a different method name) here. What do you think?

JakeQZ: This is outdated.  Did you refresh the page?

JakeQZ:
> Oh, and please also let's have a line for this in the CHANGELOG.

I was going to add that when making the equivalent changes to the `Emogrifier` class - until then the change is not available to those using the documented interface.

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
oliverklee: Oh, it indeed is outdated. Strange. I don't know how this happened. Maybe some hiccup with caches and a flaky mobile internet connection. I'll re-review right away.

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -23,6 +23,15 @@ abstract class AbstractHtmlProcessor
...
     const CONTENT_TYPE_META_TAG = '<meta http-equiv="Content-Type" content="text/html; charset=utf-8">';
 
+    /**
+     * @var string Regular expression part to match tag names that PHP's DOMDocument implementation is not aware are
+     *      self-closing. These are mostly HTML5 elements, but for completeness <command> (obsolete) and <keygen>
+     *      (deprecated) are also included.
+     *
+     * @see https://bugs.php.net/bug.php?id=73175
+     */
+    const PHP_UNRECOGNIZED_VOID_TAGNAME_MATCHER = '(?:command|embed|keygen|source|track|wbr)';
```
oliverklee: My regexp knowledge is failing me here … why the question mark?

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -74,7 +83,9 @@ public function getDomDocument()
      */
     public function render()
     {
-        return $this->domDocument->saveHTML();
+        $htmlWithPossibleErroneousClosingTags = $this->domDocument->saveHTML();
+
+        return $this->removeSelfClosingTagsClosingTags($htmlWithPossibleErroneousClosingTags);
```
oliverklee: Beautiful!

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
JakeQZ: The `?:` makes the group [non-capturing](https://www.regular-expressions.info/brackets.html), as an optimization, as we don't actually need its result stored.

oliverklee: Ah, good to know. I've just learned something. Thanks for the explanation! :heart:

PR #652, opened 2019-03-13
[BUGFIX] `CssInliner`: Support HTML5 void tags

JakeQZ:
Added support to `CssInliner` for HTML5 self-closing tags not recognized as such
by PHP’s `DOMDocument` implementation.  This is a direct port of the equivalent
changes made for `AbstractHtmlProcessor` in #651.

### DIFF
--- a/src/Emogrifier/CssInliner.php
+++ b/src/Emogrifier/CssInliner.php
@@ -64,6 +64,15 @@ class CssInliner
     const DEFAULT_DOCUMENT_TYPE = '<!DOCTYPE html>';
 
     /**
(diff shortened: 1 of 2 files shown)


oliverklee:
Looks good. Could you also please add the changelog line?

JakeQZ:
> Could you also please add the changelog line?

I was going to do this when making the equivalent changes for the `Emogrifier` class, as that is still the documented class to use.  Otherwise, if for some reason this change gets included in a realease but not the changes to `Emogrifier`, people will see a bug fix in the changelog which is not actually fixed.

PR #654, opened 2019-03-25
[TASK] Separate CssInliner::inlineCss and the rendering

oliverklee:
This is another step toward making `CssInliner` a `HtmlProcessor`.

### DIFF
--- a/src/Emogrifier/CssInliner.php
+++ b/src/Emogrifier/CssInliner.php
@@ -78,11 +78,6 @@ class CssInliner
     protected $domDocument = null;
 
     /**
-     * @var string
-     */
-    private $css = '';
-
-    /**
      * @var bool[]
      */
     private $excludedSelectors = [];
@@ -220,18 +215,6 @@ class CssInliner
     }
 
     /**
-     * Sets the CSS to merge with the HTML.
-     *
-     * @param string $css the CSS to merge, must be UTF-8-encoded
-     *
-     * @return void
-     */
-    public function setCss($css)
-    {
-        $this->css = $css;
-    }
-
-    /**
      * Renders the normalized and processed HTML.
      *
      * @return string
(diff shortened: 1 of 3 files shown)


src/Emogrifier/CssInliner.php
```diff
@@ -220,15 +215,17 @@ public function getDomDocument()
...
+     * @return CssInliner fluent interface
      */
-    public function setCss($css)
+    public function inlineCss($css = '')
     {
-        $this->css = $css;
+        $this->process($css);
+
+        return $this;
     }
```
JakeQZ: Could we dispense with this method and instead rename `process()` to `inlineCss()` (and have it `return $this` for chaining)?

src/Emogrifier/CssInliner.php
```diff
@@ -399,7 +364,6 @@ protected function process()
 
         // grab any existing style blocks from the html and append them to the existing CSS
         // (these blocks should be appended so as to have precedence over conflicting styles in the existing CSS)
-        $allCss = $this->css;
         if ($this->isStyleBlocksParsingEnabled) {
             $allCss .= $this->getCssFromAllStyleNodes($xPath);
         }
```
JakeQZ: This modifies the parameter passed to the function.  Should the parameter be a separate entity (e.g. `$css`) with the preceding line being changed to `$allCss = $css;` rather than removed?

tests/Unit/CssInlinerTest.php
```diff
@@ -430,11 +440,11 @@ public function wbrTagDataProvider()
      *
      * @dataProvider wbrTagDataProvider
      */
-    public function emogrifyByDefaultRemovesWbrTag($html)
+    public function renderByDefaultRemovesWbrTag($html)
     {
```
JakeQZ: Should the method name be `inlineCss...` rather than `render...`, since it's the `inlineCss` method that removes these elements?

tests/Unit/CssInlinerTest.php
```diff
@@ -773,14 +783,14 @@ public function matchedCssDataProvider()
...
+    public function inlineCssAppliesCssToMatchingElements($css, $expectedHtml)
     {
         $cssDeclaration1 = 'color: red;';
         $cssDeclaration2 = 'text-align: left;';
         $subject = $this->buildDebugSubject(static::COMMON_TEST_HTML);
-        $subject->setCss(\sprintf($css, $cssDeclaration1, $cssDeclaration2));
+        $subject->inlineCss(\sprintf($css, $cssDeclaration1, $cssDeclaration2));
 
-        $result = $subject->emogrify();
+        $result = $subject->render();
```
JakeQZ: Now that `inlineCss()` performs the action being tested, the blank lines separating 'set-up', 'execute' and 'check result' no longer achieve that logical grouping (this also applies to many other tests below).

src/Emogrifier/CssInliner.php
oliverklee: Done.

src/Emogrifier/CssInliner.php
oliverklee: Done.

tests/Unit/CssInlinerTest.php
oliverklee: Done.

tests/Unit/CssInlinerTest.php
oliverklee: Done.

oliverklee:
Thanks for the review! I've updated the PR according to the review comments and repushed.

PR #658, opened 2019-05-14
[CLEANUP] Update repository URLs in annotations

oliverklee:
The repository has been moved in the meantime. The URLs should
reflect this.


zoliszabo:
There still are a couple of old URLs in the CHANGELOG file: https://github.com/MyIntervals/emogrifier/search?q=jjriv&unscoped_q=jjriv. Should we also update those?

oliverklee:
Yes, definitely. I've updated the README and repushed.

PR #660, opened 2019-05-17
[TASK] Make the CssInliner inherit AbstractHtmlProcessor

### LINKED ISSUE #554: Make the CssInliner inherit AbstractHtmlProcessor

oliverklee:
Closes #554

### DIFF
--- a/src/Emogrifier/CssInliner.php
+++ b/src/Emogrifier/CssInliner.php
@@ -2,6 +2,7 @@
 
 namespace Pelago\Emogrifier;
 
+use Pelago\Emogrifier\HtmlProcessor\AbstractHtmlProcessor;
 use Symfony\Component\CssSelector\CssSelectorConverter;
 use Symfony\Component\CssSelector\Exception\SyntaxErrorException;
 
@@ -21,7 +22,7 @@ use Symfony\Component\CssSelector\Exception\SyntaxErrorException;
  * @author Sander Kruger <s.kruger@invessel.com>
  * @author Zoli Szabó <zoli.szabo+github@gmail.com>
  */
-class CssInliner
+class CssInliner extends AbstractHtmlProcessor
 {
     /**
      * @var int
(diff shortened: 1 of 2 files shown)


src/Emogrifier/CssInliner.php
```diff
@@ -275,59 +174,6 @@ private function getHeadElement()
...
-     * Creates a DOMDocument instance from the given HTML and stores it in $this->domDocument.
-     *
-     * @param string $html
-     *
-     * @return void
-     */
-    private function createRawDomDocument($html)
-    {
-        $domDocument = new \DOMDocument();
-        $domDocument->encoding = 'UTF-8';
```
JakeQZ: `AbstractHtmlProcessor` does not set `encoding`.  Should it?

src/Emogrifier/CssInliner.php
```diff
@@ -275,59 +174,6 @@ private function getHeadElement()
...
-        $domDocument->strictErrorChecking = false;
-        $domDocument->formatOutput = true;
-        $libXmlState = \libxml_use_internal_errors(true);
-        $domDocument->loadHTML($this->prepareHtmlForDomConversion($html));
-        \libxml_clear_errors();
-        \libxml_use_internal_errors($libXmlState);
-        $domDocument->normalizeDocument();
```
JakeQZ: `AbstractHtmlProcessor` does not call `normalizeDocument()`.  Should it?

oliverklee:
> We should also remove the `CssInliner` tests that duplicate those for `AbstractHtmlProcessor` but this can be a separate PR.

I agree. In the separate PR, I'll also add a test that `CssInliner` extends `AbstractHtmlProcessor`.

src/Emogrifier/CssInliner.php
oliverklee: Good catch! No, because `loadHTML` will overwrite the encoding anyway. I'll create a PR for removing that line.

src/Emogrifier/CssInliner.php
oliverklee: Probably. It looks like this is not covered with tests yet. I'll create a PR for that.

oliverklee: I haven't been able to come up with a regression test for the missing `normalizeDocument`. As it's recommended to use it, I'll create a PR anyway.

oliverklee: #662

src/Emogrifier/CssInliner.php
oliverklee: #661

oliverklee:
Rebased and repushed.

PR #661, opened 2019-05-19
[CLEANUP] Stop setting DOMDocument->encoding

oliverklee:
The encoding will be reset a few lines later with the
`loadHTML` call anyway.

### DIFF
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -461,7 +461,6 @@ class Emogrifier
     private function createRawDomDocument($html)
     {
         $domDocument = new \DOMDocument();
-        $domDocument->encoding = 'UTF-8';
         $domDocument->strictErrorChecking = false;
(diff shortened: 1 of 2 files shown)


JakeQZ:
I've since found this in a [Stack Overflow](https://stackoverflow.com/questions/8218230/php-domdocument-loadhtml-not-encoding-utf-8-correctly#answer-8218649) answer:

> `DOMDocument::loadHTML` will treat your string as being in ISO-8859-1 unless you tell it otherwise. This results in UTF-8 strings being interpreted incorrectly.

Although the answer dates from 2011, some of the supporting comments ("fixed it for me") are much more recent and/or refer to PHP 7.  We should perhaps consider the example from the Stack Overflow question in a test?

oliverklee:
> We should perhaps consider the example from the Stack Overflow question in a test?

I've just created PR #663 for this.

PR #662, opened 2019-05-19
[BUGFIX] Always use `DOMDocument::normalizeDocument` after loading HTML

oliverklee:


JakeQZ:
> As it's recommended to use it

Do you have a citation reference for this recommendation?

oliverklee:
> Do you have a citation reference for this recommendation?

Only the example on StackOverflow (where the recommendation is not explicit, though): https://stackoverflow.com/questions/1731531/how-to-use-phps-dom-extension-loadhtml

oliverklee:
The more I think about it, the more I tend to remote the existing instances of `normalizeDocument` calls as this does not break anything according to our tests, and the performance argument is very alluring. @JakeQZ @zoliszabo Would that be okay for you?

Then I'd drop this PR and create the opposite PR instead.

JakeQZ:
> @JakeQZ @zoliszabo Would that be okay for you?

I'm happy either way.

JakeQZ:
> I'm happy either way.

Having said that, I am also persuaded by the optimization argument.  I also found [some](https://hotexamples.com/examples/-/DOMDocument/loadHTML/php-domdocument-loadhtml-method-examples.html) [examples](https://hotexamples.com/examples/-/DOMDocument/normalizeDocument/php-domdocument-normalizedocument-method-examples.html) of which about half `normalizeDocument()` after `loadHTML()` and the other half don't.

oliverklee:
Okay then, closing in favor of the to-be-created PR that removes the `normalizeDocument` calls.

PR #665, opened 2019-05-24
[TASK] `CssInliner`: Keep `<wbr>` tags by default

JakeQZ:
Following #651/#652, there is now no need to remove `<wbr>` elements by default
as they are now correctly processed.  For backwards compatibility, the behaviour
is not changed for `Emogrifier`; it is only changed for `CssInliner` which is a
new class.

### DIFF
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -445,9 +445,28 @@ class CssInlinerTest extends \PHPUnit_Framework_TestCase
      *
      * @dataProvider wbrTagDataProvider
      */
-    public function inlineCssByDefaultRemovesWbrTag($html)
+    public function inlineCssByDefaultNotRemovesWbrTag($html)
     {
         $subject = $this->buildDebugSubject($html);
+
+        $subject->inlineCss('');
+
+        $expectedWbrTagCount = \substr_count($html, '<wbr');
+        $resultWbrTagCount = \substr_count($subject->render(), '<wbr');
+        self::assertSame($expectedWbrTagCount, $resultWbrTagCount);
+    }
+
+    /**
+     * @test
+     *
+     * @param string $html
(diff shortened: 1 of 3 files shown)


tests/Unit/CssInlinerTest.php
```diff
@@ -445,12 +445,31 @@ public function wbrTagDataProvider()
      *
      * @dataProvider wbrTagDataProvider
      */
-    public function inlineCssByDefaultRemovesWbrTag($html)
+    public function inlineCssByDefaultNotRemovesWbrTag($html)
```
oliverklee: I propose we drop the negation to reduce the cognitive load when reading the code: NotRemoves -> Keeps

tests/Unit/CssInlinerTest.php
```diff
@@ -445,12 +445,31 @@ public function wbrTagDataProvider()
...
     {
         $subject = $this->buildDebugSubject($html);
 
         $subject->inlineCss('');
 
+        $expectedWbrTagCount = \substr_count($html, '<wbr');
+        $resultWbrTagCount = \substr_count($subject->render(), '<wbr');
+        self::assertSame($expectedWbrTagCount, $resultWbrTagCount);
```
oliverklee: Can we move the "execution" phase to a separate line (also in the next test)?

```php
$result = $subject->render();
```

tests/Unit/CssInlinerTest.php
oliverklee: … or "ByDefaultKeeps"

JakeQZ: Done.

tests/Unit/CssInlinerTest.php
JakeQZ: Done.

PR #667, opened 2019-05-27
[CLEANUP] Drop duplicate tests from CssInlinerTest

oliverklee:
Drop all tests that already are part of `AbstractHtmlProcessorTest`,
and instead add a test that `CssInliner` extends `AbstractHtmlProcessor`.

This is the cleanup after #660.


tests/Unit/CssInlinerTest.php
```diff
@@ -196,224 +86,6 @@ public function constructorWithNoHtmlDataThrowsException($html)
         new CssInliner($html);
```
JakeQZ: `AbstractHtmlProcessorTest` has a `constructorWithNoHtmlDataThrowsException` test so is this also a duplicate that can be removed?

tests/Unit/CssInlinerTest.php
```diff
@@ -2422,160 +2094,6 @@ function (array $dataset) {
         );
```
JakeQZ: These three self-closing tag data providers are no longer referenced and can be removed.

tests/Unit/CssInlinerTest.php
```diff
@@ -2422,160 +2094,6 @@ function (array $dataset) {
...
-
-        $result = $subject->render();
-
-        self::assertNotContains('</' . $tagName, $result);
-    }
-
-    /**
-     * @test
-     */
-    public function renderAutomaticallyClosesUnclosedTag()
```
JakeQZ: We don't seem to have an equivalent test in `AbstractHtmlProcessorTest` so should this be moved there rather than removed?

tests/Unit/CssInlinerTest.php
```diff
@@ -2422,160 +2094,6 @@ function (array $dataset) {
...
-
-        $result = $subject->render();
-
-        self::assertContains('<body><p></p></body>', $result);
-    }
-
-    /**
-     * @test
-     */
-    public function renderReturnsCompleteHtmlDocument()
```
JakeQZ: We don't seem to have an equivalent test in `AbstractHtmlProcessorTest` so should this be moved there rather than removed?

tests/Unit/CssInlinerTest.php
```diff
@@ -2422,160 +2094,6 @@ function (array $dataset) {
...
-
-        $result = $subject->renderBodyContent();
-
-        self::assertSame('<p></p>', $result);
-    }
-
-    /**
-     * @test
-     */
-    public function renderBodyContentReturnsBodyContentFromPartialContent()
```
JakeQZ: We don't seem to have an equivalent test in `AbstractHtmlProcessorTest` so should this be moved there rather than removed?

tests/Unit/CssInlinerTest.php
oliverklee: Done.

tests/Unit/CssInlinerTest.php
oliverklee: Good catch! Done.

tests/Unit/CssInlinerTest.php
oliverklee: Yes, we do - it's named differently, though: `renderRepairsBrokenHtml`
https://github.com/MyIntervals/emogrifier/blob/master/tests/Unit/Emogrifier/HtmlProcessor/AbstractHtmlProcessorTest.php#L92

tests/Unit/CssInlinerTest.php
oliverklee: This is covered by several tests that test several smaller aspects of this:
-  `addsMissingHtmlTag`
- `addsMissingBodyTag'
- `renderBodyContentReturnsBodyContent`

tests/Unit/CssInlinerTest.php
oliverklee: This is covered by several tests that test several smaller aspects of this:
- `reformatsHtml`
- `addsMissingHtml5DocumentType`

oliverklee:
Worked through all review comments and push a new version.

oliverklee:
Travis currently has problem downloading PHP 5.5. The build with the other PHP versions are green. This should not block this PR.

oliverklee:
#672 should fix the PHP 5.5 issue.

oliverklee:
Repushed to include the fix for building PHP 5.5.

tests/Unit/CssInlinerTest.php
JakeQZ: ...and `addsMissingContentTypeMetaTag` - I see.

tests/Unit/CssInlinerTest.php
JakeQZ: I see.

tests/Unit/CssInlinerTest.php
JakeQZ: I see.

tests/Unit/CssInlinerTest.php
```diff
@@ -54,364 +50,11 @@ private function buildDebugSubject($html)
...
-
-    /**
-     * @test
-     *
-     * @param string $codeNotToBeChanged
-     *
-     * @dataProvider specialCharactersDataProvider
-     */
-    public function renderBodyContentKeepsSpecialCharactersInTextNodes($codeNotToBeChanged)
-    {
```
JakeQZ: I have now also spotted that `AbstractHtmlProcessorTest` does not have this test specifically for `renderBodyContent` (it has just the more general `keepsSpecialCharactersInTextNodes`) - should it (e.g. perhaps the procedure to extract the body content could affect special characters)?

oliverklee: To be sure, yes. I've added the test and will repush in a second.

JakeQZ: Looks good,

JakeQZ:
Changes are good to go.

Please can you **NOT** force-push commits.  It makes reviewing much more difficult because I can't see what the changes were since the last review.  It screws up `git` if I want a local copy of the branch and I have to resort to [Stack Overflow](https://stackoverflow.com/questions/26376832/why-does-git-say-pull-is-not-possible-because-you-have-unmerged-files) to unravel the mess.

GitHub now has "Sqash and Merge" so there is no need,

And because I requested changes in the last review, this PR is now stuck.  I can't approve it, however much I'd like to.

oliverklee:
> Please can you **NOT** force-push commits.

Will do.

PR #668, opened 2019-06-18
[WIP][FEATURE] Allow dependency injection in case of helper classes

zoliszabo:
Recently had to briefly customize the functionality `CssConcatenator` and there was no way to do it without modifying the "core" `Emogrifier.php` file in the `vendor` folder.

Hence my proposal to make the used "helper" classes inject-able:
- Made the modification in a single place only, just to showcase it;
- It allows users of the library to inject a custom subclass of `CssConcatenator`;
- It falls back to the default implementation, when there is no explicit injection of a custom implementation;
- Also modified the visibility of `$mediaRules` property (from `private` to `protected`) in `CssConcatenator`, so sub-classes have access to it.

How to use it:
```php
class CustomCssConcatenator extends CssConcatenator
{
    // Override / implement additional logic
}

...

$emogrifier = new \Pelago\Emogrifier($html);
$emogrifier->setCssConcatenator(new CustomCssConcatenator());
...
```

- Should we use interfaces (like `CssConcatenatorInterface` would declare two methods - `append()` and `getCss()`), which will be used as the argument type of the injector method and all implementations (starting with the default ones ofc) will implement those interfaces. Otherwise all additional implementations must extend the default ones provided by Emogrifier.

What do you think? Let's discuss.

Thanks!

### DIFF
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -269,7 +269,9 @@ class Emogrifier
     private $debug = false;
 
     /**
-     * Holds the eventually injected CSS concatenator dependency.
+     * Holds the eventually injected CSS concatenator dependency or the "JIT-instantiated" default implementation.
+     * Do not access this property directly, use $this->resolveCssConcatenator() instead, where you need access to the
+     * CSS concatenator object.
      *
      * @var Emogrifier\CssConcatenator
      */
@@ -1345,20 +1347,13 @@ class Emogrifier
             return;
         }
 
-        // When there is no explicitly injected dependency, just use the default one.
-        if (\is_null($this->cssConcatenator)) {
-            // support use without autoload
-            if (!\class_exists('Pelago\\Emogrifier\\CssConcatenator')) {
-                require_once __DIR__ . '/Emogrifier/CssConcatenator.php';
-            }
-            $this->cssConcatenator = new Emogrifier\CssConcatenator();
-        }
+        $cssConcatenator = $this->resolveCssConcatenator();
 
         foreach ($cssRulesRelevantForDocument as $cssRule) {
-            $this->cssConcatenator->append([$cssRule['selector']], $cssRule['declarationsBlock'], $cssRule['media']);
+            $cssConcatenator->append([$cssRule['selector']], $cssRule['declarationsBlock'], $cssRule['media']);
         }
 
-        $this->addStyleElementToDocument($this->cssConcatenator->getCss());
+        $this->addStyleElementToDocument($cssConcatenator->getCss());
     }
 
     /**
@@ -2080,5 +2075,25 @@ class Emogrifier
     public function setCssConcatenator(Emogrifier\CssConcatenator $cssConcatenator)
     {
         $this->cssConcatenator = $cssConcatenator;
+    }
+
+    /**
+     * Returns the resolved CSS concatenator object.
+     * Use this method instead of directly using $this->cssConcatenator, as that may be null.
+     *
+     * @return Emogrifier\CssConcatenator
+     */
+    protected function resolveCssConcatenator()
+    {
+        // When there is no explicitly injected dependency (or the default implementation was not yet instantiated),
+        // just set up the default one.
+        if (\is_null($this->cssConcatenator)) {
+            // support use without autoload
+            if (!\class_exists('Pelago\\Emogrifier\\CssConcatenator')) {
+                require_once __DIR__ . '/Emogrifier/CssConcatenator.php';
+            }
+            $this->cssConcatenator = new Emogrifier\CssConcatenator();
+        }
+        return $this->cssConcatenator;
     }
 }
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -269,6 +269,13 @@ class Emogrifier
     private $debug = false;
 
     /**
+     * Holds the eventually injected CSS concatenator dependency.
+     *
+     * @var Emogrifier\CssConcatenator
+     */
+    private $cssConcatenator = null;
+
+    /**
      * @param string $unprocessedHtml the HTML to process, must be UTF-8-encoded
      * @param string $css the CSS to merge, must be UTF-8-encoded
      */
@@ -1338,17 +1345,20 @@ class Emogrifier
             return;
         }
 
-        // support use without autoload
-        if (!\class_exists('Pelago\\Emogrifier\\CssConcatenator')) {
-            require_once __DIR__ . '/Emogrifier/CssConcatenator.php';
+        // When there is no explicitly injected dependency, just use the default one.
+        if (is_null($this->cssConcatenator)) {
+            // support use without autoload
+            if (!\class_exists('Pelago\\Emogrifier\\CssConcatenator')) {
+                require_once __DIR__ . '/Emogrifier/CssConcatenator.php';
+            }
+            $this->cssConcatenator = new Emogrifier\CssConcatenator();
         }
 
-        $cssConcatenator = new Emogrifier\CssConcatenator();
         foreach ($cssRulesRelevantForDocument as $cssRule) {
-            $cssConcatenator->append([$cssRule['selector']], $cssRule['declarationsBlock'], $cssRule['media']);
+            $this->cssConcatenator->append([$cssRule['selector']], $cssRule['declarationsBlock'], $cssRule['media']);
         }
 
-        $this->addStyleElementToDocument($cssConcatenator->getCss());
+        $this->addStyleElementToDocument($this->cssConcatenator->getCss());
     }
 
     /**
@@ -2060,5 +2070,15 @@ class Emogrifier
     public function setDebug($debug)
     {
         $this->debug = $debug;
+    }
+
+    /**
+     * Inject a custom CSS concatenator.
+     *
+     * @param Emogrifier\CssConcatenator $cssConcatenator Must be compatible with the default CssConcatenator.
+     */
+    public function setCssConcatenator($cssConcatenator)
+    {
+        $this->cssConcatenator = $cssConcatenator;
     }
 }
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -1346,7 +1346,7 @@ class Emogrifier
         }
 
         // When there is no explicitly injected dependency, just use the default one.
-        if (is_null($this->cssConcatenator)) {
+        if (\is_null($this->cssConcatenator)) {
             // support use without autoload
             if (!\class_exists('Pelago\\Emogrifier\\CssConcatenator')) {
                 require_once __DIR__ . '/Emogrifier/CssConcatenator.php';
@@ -2077,7 +2077,7 @@ class Emogrifier
      *
      * @param Emogrifier\CssConcatenator $cssConcatenator Must be compatible with the default CssConcatenator.
      */
-    public function setCssConcatenator($cssConcatenator)
+    public function setCssConcatenator(Emogrifier\CssConcatenator $cssConcatenator)
     {
         $this->cssConcatenator = $cssConcatenator;
     }
--- a/src/Emogrifier/CssConcatenator.php
+++ b/src/Emogrifier/CssConcatenator.php
@@ -49,7 +49,7 @@ class CssConcatenator
      *
      * @var \stdClass[]
      */
-    private $mediaRules = [];
+    protected $mediaRules = [];
 
     /**
      * Appends a declaration block to the CSS.


JakeQZ:
I have also found the need to make a hacked version of that class for debugging/testing (to add whitespace and newlines to the CSS so it is readable).

I found that for the extended class I would like the methods `getCss`, `getMediaRuleCss`, and `getRuleBlockCss`, as well as the property `$mediaRules` to be `protected` rather than `private`.

And a method to 'inject' the extended class.  I added a protected method `createCssConcatenator` to `CssInliner` that could be overridden, but a method to set it may be more convenient.  However, if it is used more than once, reusing the same object may be problematic as it would need resetting; it may be better to be able to set the class _name_ rather than class _object_.

The drawback of doing this and making methods `protected` is that it makes them public in the real sense, so it would not be possible to change the interfaces in future (unless backward-compatible) without a new major version and a release note.

zoliszabo:
> 
> I found that for the extended class I would probably like the methods `getCss`, `getMediaRuleCss`, and `getRuleBlockCss`, as well as the property `$mediaRules` to be `protected` rather than `private`. (Though actually it was possible with the [closure hack](http://ocramius.github.io/blog/accessing-private-php-class-members-without-reflection/).)
> ... 
> The drawback of doing this and making methods `protected` is that it makes them public in the real sense, so it would not be possible to change the interfaces in future (unless backward-compatible) without a new major version and a release note.

You're right about `protected` methods technically being part of the "interface" of a library, but as far as I understand from what is written out there about SemVer, the public API (of a library) is not necessarily defined by the public (& protected) methods of the code itself, but it can be a well-documented subset of it. Thus, as long as we document it well that protected methods of class X can still change, I think it should be OK even if we "open" the default implementation using `protected` instead of `private`.

Nevertheless, I think an even better approach would be to define interfaces - i.e. in this case there should be a `CssConcatenatorInterface`, which would define the two public methods any implementer class must have: `append()` & `getCss()`. Of course, extending the default implementation is still an option, if you want to reuse some of its internal" functionality, with the "risk" that you will have to pay attention for eventual changes of the parent class.


> And a method to 'inject' the extended class. I added a protected method `createCssConcatenator` to `CssInliner` that could be overridden, but a method to set it may be more convenient. However, if it is used more than once, reusing the same object may be problematic as it would need resetting; it may be better to be able to set the class _name_ rather than class _object_.

I may misunderstood you, but I am not sure I get the difference between injecting an object versus injecting a class name. In any case, I prefer injecting objects as it is more versatile (e.g. think of implementer classes that have constructor arguments - Emogrifier/CssInliner couldn't handle those).
But why not extract the "resolve" logic of the CSS concatenator object into its own overridable protected method anyway; it wouldn't hurt in any case (actually did it here: https://github.com/MyIntervals/emogrifier/pull/668/commits/6e264052ed1e69c49621b392e0951da2969a68a1).

JakeQZ:
> I may misunderstood you, but I am not sure I get the difference between injecting an object versus injecting a class name. In any case, I prefer injecting objects as it is more versatile...

Injecting a class name is like injecting a class factory (PHP allows class names to be passed around without the need for a specific factory class).  Though a class factory would have more versatility.

The point I'm trying to make is that `CssConcatenator` is not originally intended to be used like a singleton.  After it has been used, it will contain the data from that usage.  So if it were to be re-used, it would need some kind of reset method.  It is only happenchance if it is actually used only once (which I assume to be the case if the tests all pass).

The current design is that a new object instance is created each time it is required, and discarded after use (there is no need for it to persist).

So I think if consumers were to be allowed to set some property (as well the alternative of overriding `Emogrifier`/`CssInliner` with a subclassed `get`/`create`/`resolve` method), that property would have to be a class name or class factory.  Or `CssConcatenator` would have to be changed to be resettable, which I am not particularly keen on as it would mean keeping a redundant object in memory, changing the design pattern, and implementing extra methods and tests.

oliverklee:
A few thoughts:

- I like the approach.
- We should define an interface that the inject method expects.
- I agree that `CssConcatenator` must not be a singleton. However, I think it's okay if we shift the responsibility to injecting a fresh instance to the environment that provides an alternative implementation. (We should also document this in the inject method.)
- I'd prefer to have this in the new `CssInliner` class only for two reasons: 1. I'd like the `Emogrifier` to be extended as little as possible as it's complex and unwieldy. 2. Unlike `Emogrifier`, `CssInliner` is intended to be used with a fresh instance for reach HTML document, which might help with the Singleton question.

JakeQZ:
> We should define an interface that the inject method expects.

So, are we leaning towards a method to be overridden in an extended `CssInliner`, rather than a property to be set (via a get/setter)?  I think the 'method' approach would be simpler to implement.

> I think it's okay if we shift the responsibility to injecting a fresh instance to the environment that provides an alternative implementation.

The default implementation would also need to take that responsibility, though probably as simply as `return new CssConcatenator();`.

> `CssInliner` is intended to be used with a fresh instance for [<del>r</del>]each HTML document, which might help with the Singleton question.

I'm not sure that should be _allowed_ to help.  Although `CssConcatenator` is only used in `copyUninlineableCssToStyleNode` which (normally) is only called once, no-one (in future maintainance) would expect a side-effect that `copyUninlineableCssToStyleNode` produces different results if called a second time because it's reusing a dirty object.

oliverklee:
> So, are we leaning towards a method to be overridden in an extended CssInliner, rather than a property to be set (via a get/setter)? I think the 'method' approach would be simpler to implement.

I'm confused …

I was thinking about something like this:

```php
class CssInliner {
public function injectCssConcatenator(CssConcatenatorInterface $concatenator) {…}
}

class CssConcatenatorInterface implements CssConcatenatorInterface {}
```

oliverklee:
> The default implementation would also need to take that responsibility, though probably as simply as return `new CssConcatenator();`.

Agreed.

oliverklee:
And I agree that `CssConcatenator::copyUninlineableCssToStyleNode` should not have side effects. I propose we fix this separately.

JakeQZ:
> ```php
> class CssInliner {
>   public function injectCssConcatenator(CssConcatenatorInterface $concatenator) {…}
> }
> ```

Ah, ok, so a method to set an object (presumably as a property), rather than a overridable method which returns said object.

> ```php
> class CssConcatenatorInterface implements CssConcatenatorInterface {}
> ```

Presumably this should be `class CssConcatenator implements CssConcatenatorInterface`?

> And I agree that `CssConcatenator::copyUninlineableCssToStyleNode` should not have side effects. I propose we fix this separately.

It doesn't at present, but would if it were changed to use a `CssConcatenator` instance stored as a property instead of creating a `new` one each time.

oliverklee:
> Presumably this should read `class CssConcatenator implements CssConcatenatorInterface`?

Yes, you're right. (That's what happens when I reply from a train when it's too hot … ;-) )

JakeQZ:
> > And I agree that `CssConcatenator::copyUninlineableCssToStyleNode` should not have side effects. I propose we fix this separately.
> 
> It doesn't at present, but would if it were changed to use (and potentially reuse) a `CssConcatenator` instance stored as a property instead of creating a new one each time.

I've added #671 to test this.  It currently passes but the current proposed implementation in the [dependency-injection](https://github.com/MyIntervals/emogrifier/tree/dependency-injection) branch would cause it to fail, due to reusing a `CssConcatenator`.

JakeQZ:
... so I think this calls for the [factory design pattern](https://en.wikipedia.org/wiki/Factory_method_pattern).  In most OO languages this would require a factory class.  PHP would (as an alternative, still following the design pattern) allow a class name (of the objects to be created) to be set as a pseudo-factory without having to define/implement a factory interface - whether that would be benefitial is open to debate.

zoliszabo:
Thinking it through, I would still revisit the idea of simple (object) dependency injection, where the to-be-injected object must implement a well-defined interface, and the idea of having a `reset()` method for `CssConcatenator`.

In the concrete case of  `CssConcatenator`, the interface (`CssConcatenatorInterface`) would define 3 methods: `append()`, `getCss()` and **`reset()`**. The class is some kind of a collection/container (of CSS rules), so - imho - there is nothing wrong with having a way to empty the container. I see this solution less complicated and clearer than injecting factory classes or class names. Neither of these two really "force" the implementer of a custom `CssConcatenator` class to really clean up between instances (I am evil and would collect CSS rules in a static property 😈), only demanding a `reset()` method does.

JakeQZ:
> I would still revisit the idea of simple (object) dependency injection ... The class is some kind of a collection/container (of CSS rules) ...

I was also thinking that this approach would make it no longer possible to use more than one such container at the same time (if required at some future time).

However, PHP’s object `clone`-ing means this would still be possible – a second, separate container could be `clone`d from the first (in its initial, empty state).

Thus, I wonder, if we do take this approach, should `resolveCssConcatenator` always return a new `clone`, while the `cssConcatenator` property holds a ‘template’ object in the initial, empty state?  (It might then not be necessary for a `reset` method.)

My main objection is `Emogrifier`/`CssInliner` being restricted to only being able to use one `CssConcatentor` container instance throughout its lifecycle.

> Neither of these two really "force" the implementer of a custom CssConcatenator class to really clean up between instances (I am evil and would collect CSS rules in a static property 😈), only demanding a reset() method does.

I could be evil and implement `reset()` as a no-op 😉

zoliszabo:
After more thinking...

If we limit ourselves to current functionality, then simple object dependency injection with a reset method is sufficient.

BUT if we want to make it super future-proof, then I am leaning towards injecting a factory class (of course using a factory interface - `CssConcatenatorFactoryInterface`). While the above described cloning method is quite a cool trick to cover our current and eventual future use-cases, I feel the factory pattern is a bit cleaner.

If you guys agree, I will submit a new PR implementing the factory pattern way.

JakeQZ:
> If you guys agree, I will submit a new PR implementing the factory pattern way.

Although it involves more indirection and more interfaces, I think this would be the most future-proof.  We could avoid an extra class by having `CssInliner` (and `Emogrifier`) _itself_ implement `CssConcatenatorFactoryInterface` and default

```php
    $this->cssConcatenatorFactory = $this;
```

I don't know if that would be preferred to a separate default factory class?

zoliszabo:
> We could avoid an extra class by having `CssInliner` (and `Emogrifier`) _itself_ implement `CssConcatenatorFactoryInterface`...
> I don't know if that would be preferred to a separate default factory class?

While this is a neat idea, I would still vote for a separate class because of the following:
 - we do not have to duplicate code for `Emogrifier` and `CssInliner`;
 - (and more importantly) the default implementation would be a perfect example of how to create/use a custom implementation.

oliverklee:
I am reluctant to go the factory approach now. This feels over-engineered to me in order to solve a problem we do not have (yet). Unless we really need this flexibility, please let's keep it simple. (See https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it for details)

However, if there is a valid usecase coming up in the near future where we'll need multiple `CssConcatenator`s, I might be persuaded otherwise. What is the usecase, and how soon will it come up?

JakeQZ:
> What is the usecase, and how soon will it come up?

I don't have a crystal ball.  But should the situation arise it would require a new major release as a public interface is changed.

> This feels over-engineered to me in order to solve a problem we do not have (yet). Unless we really need this flexibility, please let's keep it simple. (See https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it for details.)

The simplest solution is to just add to `CssInliner`/`Emogrifier` a method that can be overridden:

```php
protected function createCssConcatenator()
{
    return new CssConcatenator();
}
```

Then if anyone wants to 'inject' a different object, they extend `CssInliner`/`Emogrifier` and override that method.

Not only is it simple, it also does not hinder future development (i.e. restricting to a single `CssConcatenator` instance or introducing unexpected method side-effects).

However, the gist I was getting is that you are both preferring to have some property that can be set, rather than some method that can be overridden…?

zoliszabo:
> However, the gist I was getting is that you are both preferring to have some property that can be set, rather than some method that can be overridden…?

It's just that I named the PR "dependency injection" and tried to find a way to make it work that way - without the need for sub-classing 😊

This being said, jokes aside, I do feel that the method overriding way is the simplest solution in our current case. It helps both with "resetting" the collection of rules and (eventually) use more concatenator objects in parallel too.

oliverklee:
> This being said, jokes aside, I do feel that the method overriding way is the simplest solution in our current case.

I concur.

oliverklee:
@zoliszabo Do you consider this PR ready to review, is it still a WIP, or was it only to show some prototype code to discuss?

src/Emogrifier.php
```diff
@@ -268,6 +268,15 @@ class Emogrifier
...
     private $debug = false;
 
+    /**
+     * Holds the eventually injected CSS concatenator dependency or the "JIT-instantiated" default implementation.
+     * Do not access this property directly, use $this->resolveCssConcatenator() instead, where you need access to the
+     * CSS concatenator object.
+     *
+     * @var Emogrifier\CssConcatenator
+     */
+    private $cssConcatenator = null;
```
JakeQZ: As I understand, we have subsequently agreed on a simpler solution, which is just to add a `createCssConcatenator` method that can be overridden to return a different or extended implementation.

src/Emogrifier/CssConcatenator.php
```diff
@@ -49,7 +49,7 @@ class CssConcatenator
      *
      * @var \stdClass[]
      */
-    private $mediaRules = [];
+    protected $mediaRules = [];
```
JakeQZ: This change is fine.

zoliszabo:
> 
> 
> @zoliszabo Do you consider this PR ready to review, is it still a WIP, or was it only to show some prototype code to discuss?

Hi @oliverklee, as @JakeQZ wrote above:
> As I understand, we have subsequently agreed on a simpler solution, which is just to add a createCssConcatenator method that can be overridden to return a different or extended implementation.

I will implement the necessary changes in 1-2 days, then ask for your review. (Unfortunately, I had no time in the past few weeks to participate in the sustained work you both did here; I will try to gear up my pace the next days.)

oliverklee:

Issue #669, opened 2019-06-25
Make the static methods in CssConcatenator non-static


oliverklee:
The public methods are not static, as there is no reason some private methods are static.

JakeQZ:
They're `static` because they don't operate on `$this` and are more like helper functions.  There's no reason I see for them _not_ to be `static`.

oliverklee:
It seems we have two different philosophies for private methods:
- make them static unless they need to be non-static
- make them non-static unless they need to be static

So we're at an impasse here, I think. :-)

Closing this ticket as I'm not really passionate about this change.

JakeQZ:
Agree it boils down to opinion, and YMMV :)

I like to emphasize the 'contract' with `static` if possible that "this method will not modify the object".  In PHP, with late static binding, even `static` methods are overridable.  In C++ they would need to be made non-`static` to allow overriding.

That said, I've noticed that `getCss` uses `[$this, 'getMediaRuleCss']` as a `callable` whereas `getMediaRuleCss` uses `[static::class, 'getRuleBlockCss']`.  In current PHP versions the end behaviour and result is the same.  The latter does require PHP 5.5+ (which we require), but the former might some time in future trigger a notice or warning about calling a static function non-statically...

oliverklee:
Oh, then we need to fix this.

PR #670, opened 2019-07-03
[CLEANUP] Call static methods statically

JakeQZ:
In `CssConcatenator`, create `callable` for `static` method using `static::class` not `$this`.

(Spotted in #669.  Instances of `static` methods being called through `$this` were previously corrected in #525.)


oliverklee:
Looks good. I propose we have a line in the changelog for this, though (in the "fixes" section).

JakeQZ:
Added changelog entry and merging - let me know if you spot any issues with that.

PR #671, opened 2019-07-03
[TASK] Test copyUninlineableCssToStyleNode side-effects

JakeQZ:
Added a test to confirm `copyUninlineableCssToStyleNode` does not have unexpected side-effects, i.e. it would produce the same result if called a second time on the same `CssInliner` instance.

This will help ensure that the solution for #668 does not create future potential maintenance issues.

### DIFF
--- a/tests/Unit/EmogrifierTest.php
+++ b/tests/Unit/EmogrifierTest.php
@@ -3334,4 +3334,34 @@ class EmogrifierTest extends \PHPUnit_Framework_TestCase
 
         self::assertContainsCss($css, $result);
     }
+
+    /**
+     * @test
+     */
+    public function copyUninlineableCssToStyleNodeHasNoSideEffects()
+    {
+        $this->subject->setHtml('<html><a>foo</a><p>bar</p></html>');
+        $css = 'a:hover { color: green; } p:hover { color: blue; }';
+
+        $parseCssRules = new \ReflectionMethod(Emogrifier::class, 'parseCssRules');
+        $parseCssRules->setAccessible(true);
+        $copyUninlineableCssToStyleNode = new \ReflectionMethod(Emogrifier::class, 'copyUninlineableCssToStyleNode');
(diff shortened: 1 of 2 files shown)


JakeQZ:
The test is added to `CssInlinerTest` but could also be added to `EmogrifierTest` if desired...?

tests/Unit/EmogrifierTest.php
```diff
@@ -3334,4 +3334,34 @@ public function emogrifyNotInDebugModeKeepsInvalidOrUnrecognizedSelectorsInMedia
...
+     */
+    public function copyUninlineableCssToStyleNodeHasNoSideEffects()
+    {
+        $this->subject->setHtml('<html><a>foo</a><p>bar</p></html>');
+        $css = 'a:hover { color: green; } p:hover { color: blue; }';
+
+        $parseCssRules = new \ReflectionMethod(Emogrifier::class, 'parseCssRules');
+        $parseCssRules->setAccessible(true);
+        $copyUninlineableCssToStyleNode = new \ReflectionMethod(Emogrifier::class, 'copyUninlineableCssToStyleNode');
+        $copyUninlineableCssToStyleNode->setAccessible(true);
```
oliverklee: I really don't like us accessing private methods as this will make refactoring a lot harder. Please let's use public methods only for the test.

JakeQZ: It's not easily possible to test the private method another way.  Perhaps we don't want to make this change permanent, but it would be useful to confirm that changes made for #668 don't result in the method having unexpected side-effects.

JakeQZ: Though looking again, the parameter `$cssRules['uninlineable']` could be set up as static data, so that only the method under test is accessed privately...

JakeQZ: I've now changed this so the data passed to the method under test is initialized as an array instead of by invoking private methods.

PR #678, opened 2019-08-29
[BUGFIX] Keep development files out of the Composer packages

oliverklee:
Mark all development-only files and directories as `export-ignore`
via the `.gitattributes` file. This will keep them from being packaged
in the release ZIPs provided by GitHub and in the packages downloaded
by Composer with the `--prefer-dist` option (which is the default).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,6 +2,7 @@
 /.github/ export-ignore
 /.gitignore export-ignore
 /.travis.yml export-ignore
+/CODE_OF_CONDUCT.md export-ignore
 /config/ export-ignore
 /phpcs.xml.dist export-ignore
 /tests/ export-ignore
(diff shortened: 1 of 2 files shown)


oliverklee:
> Should `CODE_OF_CONDUCT.md` also be excluded as it relates to "participation in our project"?

Makes sense. I'll push an update in a minute.

oliverklee:
In the article https://madewithlove.be/gitattributes/, the readme and changelog are blacklisted as well. Should we remove them as well?

JakeQZ:
> In the article https://madewithlove.be/gitattributes/, the readme and changelog are blacklisted as well. Should we remove them as well?

I'm not sure.  They strike me as being a potentially useful in the distributed package, and they are not 'development-only'.

oliverklee:
I'm fine either way.

PR #689, opened 2019-09-01
[TASK] Prepare release of version 2.2.0

oliverklee:


composer.json
```diff
@@ -89,7 +89,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.1.x-dev"
+            "dev-master": "3.0.x-dev"
```
JakeQZ: Precedently we'd have this now as `2.2.x-dev`.  I know 3.0 is planned soon, but just querying whether this is correct/OK?  Because after 3.0 is released this will ***still*** be `3.0.x-dev`...

oliverklee: Makes sense. I've changed the alias accordingly and also updated the corresponding instructions in the README.

JakeQZ: Hmm.  Confusion abounds.  Historically this has actually been set to the next minor version, i.e. when 1.2.0 was released it was set to `1.3.x-dev` (though there never was a 1.3), and when 2.0.0 was released it was set to `2.1.x-dev`.  However, when 2.1.0 was released it was not updated.

So if what we were doing before was correct, we should have `2.3.x-dev` here, and when 3.0 is released update it to `3.1.x-dev`.  But I am not sure what we should be doing...

oliverklee: I've thought about this some more (and read https://getcomposer.org/doc/articles/aliases.md). The point of the alias is to allow people to get the latest dev version of a particular branch while adhering to semantic versioning.

And we (the Emogrifier team) have the additional secret knowledge of what the next release number will most probably be. So the previous version of the PR was correct (where the branch alias was 3.0.x-dev).

So, in this case, as we'll have breaking changes directly after the release of 2.2.0, we should have 3.0.x-dev as alias again. If nobody objects, I'll revert (or remove) the most recent commit in this PR and repush.

JakeQZ: OK.  Sorry for the confusion.

oliverklee:
I've dropped the most recent commit and repushed.

PR #697, opened 2019-09-10
[BUGFIX] Allow `:not(:behavioural-pseudo-class)`

JakeQZ:
This was causing an error because `removeUnmatchablePseudoComponents()` was
converting the selector to `:not()`, which is invalid.  The intention would have
been to convert it to `:not(*)`, but this is not correct either.  If a
behavioural pseudo-class is used within the `:not` argument, then the entire
`:not(…)` component should be stripped in order to test whether the selector
could match any DOM elements.

### DIFF
--- a/src/Emogrifier/CssInliner.php
+++ b/src/Emogrifier/CssInliner.php
@@ -654,6 +654,7 @@ class CssInliner extends AbstractHtmlProcessor
 
     /**
      * Removes pseudo-elements and dynamic pseudo-classes from a CSS selector, replacing them with "*" if necessary.
+     * If such a pseudo-component is within the argument of `:not`, the entire `:not` component is removed or replaced.
      *
      * @param string $selector
      *
@@ -662,11 +663,28 @@ class CssInliner extends AbstractHtmlProcessor
      */
     private function removeUnmatchablePseudoComponents($selector)
     {
+        // The regex allows nested brackets via `(?2)`.
+        // A space is temporarily prepended because the callback can't determine if the match was at the very start.
+        $selectorWithNotsRemoved = \ltrim(\preg_replace_callback(
+            '/(\\s?+):not(\\([^\\(\\)]*+(?:(?2)[^\\(\\)]*+)*+\\))/i',
+            static function (array $matches) {
+                if (\preg_match(
+                    '/:(?!' . static::PSEUDO_CLASS_MATCHER . ')[\\w\\-:]/i',
+                    $matches[2]
+                )) {
+                    return $matches[1] !== '' ? $matches[1] . '*' : '';
+                } else {
+                    return $matches[0];
+                }
+            },
+            ' ' . $selector
+        ));
+
         $pseudoComponentMatcher = ':(?!' . static::PSEUDO_CLASS_MATCHER . '):?+[\\w\\-]++(?:\\([^\\)]*+\\))?+';
         return \preg_replace(
             ['/(\\s|^)' . $pseudoComponentMatcher . '/i', '/' . $pseudoComponentMatcher . '/i'],
             ['$1*', ''],
-            $selector
+            $selectorWithNotsRemoved
         );
     }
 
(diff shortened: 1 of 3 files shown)


JakeQZ:
Given `Emogrifier` will be deprecated in 3.0.0, is it still worth applying fixes like this (where the changes can be simply copied over) to that class?

src/Emogrifier/CssInliner.php
```diff
@@ -662,11 +663,28 @@ function (array $cssRule) {
      */
     private function removeUnmatchablePseudoComponents($selector)
     {
+        // The regex allows nested brackets via `(?2)`.
+        // A space is temporarily prepended because the callback can't determine if the match was at the very start.
+        $selectorWithNotsRemoved = \ltrim(\preg_replace_callback(
+            '/(\\s?+):not(\\([^\\(\\)]*+(?:(?2)[^\\(\\)]*+)*+\\))/i',
+            static function (array $matches) {
```
oliverklee: I propose having a well-named regular method instead of the anonymous function and some well-named local variables to make the code more explicit.

src/Emogrifier/CssInliner.php
```diff
@@ -662,11 +663,28 @@ function (array $cssRule) {
...
+            },
+            ' ' . $selector
+        ));
+
         $pseudoComponentMatcher = ':(?!' . static::PSEUDO_CLASS_MATCHER . '):?+[\\w\\-]++(?:\\([^\\)]*+\\))?+';
         return \preg_replace(
             ['/(\\s|^)' . $pseudoComponentMatcher . '/i', '/' . $pseudoComponentMatcher . '/i'],
             ['$1*', ''],
-            $selector
+            $selectorWithNotsRemoved
```
oliverklee: … or maybe `$selectorWithoutNots`

oliverklee:
> Given `Emogrifier` will be deprecated in 3.0.0, is it still worth applying fixes like this (where the changes can be simply copied over) to that class?

For bug fixes, yes, I'd say so.

src/Emogrifier/CssInliner.php
JakeQZ: I've moved the anonymous function to a method, and documented the `$matches` parameter in PHPDoc.  Hopefully this makes it clear what they are without additionally assigning the values in `$matches` to local variables...?

src/Emogrifier/CssInliner.php
JakeQZ: Changed as suggested.

src/Emogrifier/CssInliner.php
```diff
@@ -662,14 +663,46 @@ function (array $cssRule) {
...
     }
 
+    /**
+     * Helper for `removeUnmatchablePseudoComponents()` to replace/remove a selector `:not` component if its argument
+     * contains pseudo-elements or dynamic pseudo-classes.
+     *
+     * @param string[] $matches Array of capturing groups matched by the regular expression.
+     *        Index 0 contains the full match ":not(...)", preceded by a whitespace character if present.
+     *        Index 1 contains any preceding whitespace character or an empty string if none present.
+     *        Index 2 contains the `:not` argument with the surrounding brackets (i.e. "(...)").
```
oliverklee: Instead of explaining the indices in the comment, I propose we use local variables in the methos.

src/Emogrifier/CssInliner.php
```diff
@@ -662,14 +663,46 @@ function (array $cssRule) {
...
+     *
+     * @return string the full match if there were no unmatchable pseudo components within; otherwise, any preceding
+     *         whitespace followed by "*", or an empty string if there was no preceding whitespace
+     */
+    private function replaceUnmatchableNotComponent(array $matches)
+    {
+        if (\preg_match(
```
oliverklee: I also do not understand this condition from reading. Please let's put its result into a well-named variable.

src/Emogrifier/CssInliner.php
JakeQZ: Done.

src/Emogrifier/CssInliner.php
JakeQZ: Done.

PR #699, opened 2019-09-11
[CLEANUP] Drop redundant README section on installation

oliverklee:
Now there is only one section explaining how to install Emogrifier
with Composer, not two.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ selectors.
 
 ## Installation
 
-For installing emogrifier, either add pelago/emogrifier to your
+For installing emogrifier, either add `pelago/emogrifier` to your
 project's composer.json, or you can use composer as below:
 
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -118,38 +118,6 @@ calling the `emogrify` method:
...
-You can also add follow lines to your `composer.json` and run the
-`composer update` command:
-
-```json
-"require": {
-  "pelago/emogrifier": "^2.2"
-}
-```
-
-See https://getcomposer.org/ for more information and documentation.
```
JakeQZ: I agree we don't need to document how to use Composer.  But is it still worth retaining the link to getcomposer.org – moving it to the end of the "Installation" section, and perhaps adding "… on Composer" to the end of the sentence?

README.md
```diff
@@ -118,38 +118,6 @@ calling the `emogrify` method:
 * `$emogrifier->addExcludedSelector(string $selector)` - Keeps elements from
   being affected by emogrification.
 
-## Installing with Composer
```
JakeQZ: This heading also needs reming from the Table of Contents.

oliverklee: Done.

README.md
oliverklee: Agreed. Done.

oliverklee:
Updated and repushed.

Issue #695, opened 2019-09-09
Update documentation to describe `CssInliner` instead of `Emogrifier`


JakeQZ:
From 3.0, `Emogrifier` will be deprecated so the documention should detail how to use the `CssInliner` and family of classes rather than `Emogrifier`.

JakeQZ:
I've updated a project to use `CssInliner` instead of `Emogrifier` without any problems.  The following code snippets may be useful to help describe the changes that would be needed to migrate from `Emogrifier` to `CssInliner`:

Old code using `Emogrifier`:
```php
$emogrifier = new Emogrifier($html);
$emogrifier->enableCssToHtmlMapping();
$html = $emogrifier->emogrify();
```

New code using `CssInliner` and family:
```php
$html = CssToAttributeConverter::fromDomDocument(
  HtmlPruner::fromDomDocument(
    CssInliner::fromHtml($html)->inlineCss('')->getDomDocument()
  )->removeInvisibleNodes()->getDomDocument()
)->convertCssToVisualAttributes()->render();
```

oliverklee:
You can also reuse the same `$domDocument` instance (i.e., you'll need only one `getDomDocument()` call in total):

```php
$document = CssInliner::fromHtml($html)->inlineCss('')->getDomDocument();
HtmlPruner::fromDomDocument($document)->removeInvisibleNodes();
$html = CssToAttributeConverter::fromDomDocument($document)
  ->convertCssToVisualAttributes()->render();
```

By the way, why the `inlineCss('')` call with an empty string?

JakeQZ:
> You can also reuse the same `$domDocument` instance

Good point.

> why the `inlineCss('')` call with an empty string?

In my case, all the CSS is contained in a `<style>` element within the HTML.  Though passing some `$css` as a parameter would probably make for a better example.

oliverklee:
Would it make sense to have the argument for `inlineCss` optional for cases like that?

JakeQZ:
> Would it make sense to have the argument for `inlineCss` optional for cases like that? (independent of the documentation)

It could be, though it's not a big deal.  Though for consistency with `Emogrifier` it should probably be optional - it was optional in the `Emogrifier` constructor.

PR #701, opened 2019-09-11
[TASK] Deprecate the Emogrifier class

### LINKED ISSUE #680: CssInliner in release 3.0.0
1. Should we mark `CssInliner` as the new default (instead of as as tech preview) and mark the `Emogrifier` class as deprecated in 3.0?
1. As 3.0 is per semantic-versioning definition a breaking release (and `CssInliner` a tech preview before 3.0), do you think ti would be okay if we removed the automatic removal of invisible nodes from `CssInliner` for 3.0?

oliverklee:
The `CssInliner` class now is not a tech preview anymore, and it should
be used instead.

The documentation in the README will be enhanced and polished even more
in a separate change.

Closes #680
Part of #695

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#691](https://github.com/MyIntervals/emogrifier/pull/691))
 
 ### Deprecated
+- Deprecate the Emogrifier class
+  ([#701](https://github.com/MyIntervals/emogrifier/pull/701))
 
 ### Removed
 - Drop `enableCssToHtmlMapping` and `disableInvisibleNodeRemoval`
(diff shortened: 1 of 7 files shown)


oliverklee:
In the README, I mostly have moved stuff around.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#691](https://github.com/MyIntervals/emogrifier/pull/691))
 
 ### Deprecated
+- Deprecate the Emogrifier class
```
JakeQZ: The class name should be in backticks.

README.md
```diff
@@ -93,7 +151,7 @@ the complete HTML document, you can use the `emogrifyBodyContent` instead:
 $bodyContent = $emogrifier->emogrifyBodyContent();
 ```
 
-## Options
+### Options
```
JakeQZ: I wonder if this should now even be a level 4 heading, to be clear it only applies to using the legacy Emogrifier class?

CHANGELOG.md
oliverklee: Done.

README.md
oliverklee: Makes sense. Changed.

oliverklee:
Updated and repushed.

PR #707, opened 2019-09-12
[FEATURE] Add CssInliner::getMatchingUninlinableSelectors

JakeQZ:
Part of #380.

The basic functionality of the method is unit-tested.  More extensive unit tests
involving all kinds of CSS rules and HTML already exist for testing the full
content of the `<style>` element containing the uninlinable rules.

### DIFF
--- a/src/Emogrifier/CssInliner.php
+++ b/src/Emogrifier/CssInliner.php
@@ -627,6 +627,20 @@ class CssInliner extends AbstractHtmlProcessor
     }
 
     /**
+     * Gets the array of selectors present in the CSS provided to `inlineCss()` for which the declarations could not be
+     * applied as inline styles, but which may affect elements in the HTML.  The relevant CSS will have been placed in a
+     * `<style>` element.  The selectors may include those used within @media rules or those involving dynamic
+     * pseudo-classes (such as `:hover`) or pseudo-elements (such as `::after`).  `inlineCss()` must have been called
+     * first, otherwise an empty array is returned.
+     *
+     * @return string[]
+     */
+    public function getMatchingUninlinableSelectors()
+    {
+        return \array_column($this->matchingUninlinableCssRules, 'selector');
+    }
+
+    /**
      * Determines which of `$cssRules` actually apply to `$this->domDocument`, and sets them in
      * `$this->matchingUninlinableCssRules`.
      *
--- a/src/Emogrifier/CssInliner.php
+++ b/src/Emogrifier/CssInliner.php
@@ -134,7 +134,7 @@ class CssInliner extends AbstractHtmlProcessor
      *
      * @var string[][]
      */
-    private $matchingUninlinableCssRules = [];
+    private $matchingUninlinableCssRules;
 
     /**
      * Emogrifier will throw Exceptions when it encounters an error instead of silently ignoring them.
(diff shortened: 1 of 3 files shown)


src/Emogrifier/CssInliner.php
```diff
@@ -626,6 +626,20 @@ private function attributeValueIsImportant($attributeValue)
         return \strtolower(\substr(\trim($attributeValue), -10)) === '!important';
     }
 
+    /**
+     * Gets the array of selectors present in the CSS provided to `inlineCss()` for which the declarations could not be
+     * applied as inline styles, but which may affect elements in the HTML.  The relevant CSS will have been placed in a
+     * `<style>` element.  The selectors may include those used within @media rules or those involving dynamic
```
oliverklee: `@media` also should be in backticks.

src/Emogrifier/CssInliner.php
```diff
@@ -626,6 +626,20 @@ private function attributeValueIsImportant($attributeValue)
...
+     * pseudo-classes (such as `:hover`) or pseudo-elements (such as `::after`).  `inlineCss()` must have been called
+     * first, otherwise an empty array is returned.
```
oliverklee: I propose the method throws a `BadMethodCallException` if it is called before `inlineCss()` is called.

tests/Unit/CssInlinerTest.php
```diff
@@ -2618,6 +2618,59 @@ public function inlineCssNotInDebugModeKeepsInvalidOrUnrecognizedSelectorsInMedi
...
+     * @test
+     */
+    public function getMatchingUninlinableSelectorsReturnsMatchingUninlinableSelector()
+    {
+        $subject = $this->buildDebugSubject('<html><p>foo</p></html>');
+        $subject->inlineCss('p:hover { color: green; }');
+
+        $result = $subject->getMatchingUninlinableSelectors();
+
+        static::assertContains('p:hover', $result);
```
oliverklee: I propose we put the `p:hover` into a variable to make the connection more explicit. Ditto in the other tests.

src/Emogrifier/CssInliner.php
JakeQZ: Done

src/Emogrifier/CssInliner.php
JakeQZ: I wondered about this.  Now changed and added test for the expected exception.

tests/Unit/CssInlinerTest.php
JakeQZ: I've refactored the tests to use a data provider, which achieves this, and also allows additional tests to be added more easily if required - in fact I've added one more test for completeness.

src/Emogrifier/CssInliner.php
```diff
@@ -626,6 +626,25 @@ private function attributeValueIsImportant($attributeValue)
...
+     * `<style>` element.  The selectors may include those used within `@media` rules or those involving dynamic
+     * pseudo-classes (such as `:hover`) or pseudo-elements (such as `::after`).
+     *
+     * @return string[]
+     *
+     * @throws \BadMethodCallException if `inlineCss` has not been called first
+     */
+    public function getMatchingUninlinableSelectors()
+    {
+        if (!isset($this->matchingUninlinableCssRules)) {
```
oliverklee: As the field exists, I'd prefer it to (explicitly) be initialized with `null`, and to have an `=== null` check instead of the `isset`. I think that this communicates the intent more clearly.

JakeQZ: OK

JakeQZ:
Repushed...

PR #708, opened 2019-09-13
[FEATURE] Add HtmlPruner::removeRedundantClasses

JakeQZ:
Part of #380.

### DIFF
--- a/src/Emogrifier/HtmlProcessor/HtmlPruner.php
+++ b/src/Emogrifier/HtmlProcessor/HtmlPruner.php
@@ -2,10 +2,13 @@
 
 namespace Pelago\Emogrifier\HtmlProcessor;
 
+use Pelago\Emogrifier\Utilities\ArrayIntersector;
+
 /**
  * This class can remove things from HTML.
  *
  * @author Oliver Klee <github@oliverklee.de>
+ * @author Jake Hotson <jake.github@qzdesign.co.uk>
  */
 class HtmlPruner extends AbstractHtmlProcessor
 {
@@ -51,37 +54,59 @@ class HtmlPruner extends AbstractHtmlProcessor
      * This method also has the (presumably beneficial) side-effect of minifying (removing superfluous whitespace from)
      * `class` attributes.
      *
-     * @param string[] $classesToKeep list of class names that should not be removed
+     * @param string[] $classesToKeep class names that should not be removed
      *
      * @return self fluent interface
      */
-    public function removeRedundantClasses(array $classesToKeep)
+    public function removeRedundantClasses(array $classesToKeep = [])
     {
         $nodesWithClassAttribute = $this->xPath->query('//*[@class]');
 
         if ($classesToKeep !== []) {
-            // https://stackoverflow.com/questions/6329211/php-array-intersect-efficiency
-            // It's more efficient to invert the array and use `array_intersect_key()` when doing many intersections.
-            $classesToKeepAsKeys = \array_flip($classesToKeep);
-
-            foreach ($nodesWithClassAttribute as $node) {
-                $nodeClasses = \preg_split('/\\s++/', \trim($node->getAttribute('class')));
-                $nodeClassesToKeep = \array_flip(\array_intersect_key(
-                    \array_flip($nodeClasses),
-                    $classesToKeepAsKeys
-                ));
-                if ($nodeClassesToKeep !== []) {
-                    $node->setAttribute('class', \implode(' ', $nodeClassesToKeep));
-                } else {
-                    $node->removeAttribute('class');
-                }
-            }
+            $this->removeClassesFromNodes($nodesWithClassAttribute, $classesToKeep);
         } else {
-            foreach ($nodesWithClassAttribute as $node) {
-                $node->removeAttribute('class');
-            }
+            // Avoid unnecessary processing if there are no classes to keep.
+            $this->removeClassAttributeFromNodes($nodesWithClassAttribute);
         }
 
         return $this;
+    }
+
+    /**
+     * Removes classes from the `class` attribute of each element in `$nodeList`, except any in `$classesToKeep`,
+     * removing the `class` attribute itself if the resultant list is empty.
+     *
+     * @param \DOMNodeList $nodeList
+     * @param string[] $classesToKeep
+     *
+     * @return void
+     */
+    private function removeClassesFromNodes(\DOMNodeList $nodeList, array $classesToKeep)
+    {
+        $classesToKeepIntersector = new ArrayIntersector($classesToKeep);
+
+        foreach ($nodeList as $node) {
+            $nodeClasses = \preg_split('/\\s++/', \trim($node->getAttribute('class')));
+            $nodeClassesToKeep = $classesToKeepIntersector->intersectWith($nodeClasses);
+            if ($nodeClassesToKeep !== []) {
+                $node->setAttribute('class', \implode(' ', $nodeClassesToKeep));
+            } else {
+                $node->removeAttribute('class');
+            }
+        }
+    }
+
+    /**
+     * Removes the `class` attribute from each element in `$nodeList`.
+     *
+     * @param \DOMNodeList $nodeList
+     *
+     * @return void
+     */
+    private function removeClassAttributeFromNodes(\DOMNodeList $nodeList)
+    {
+        foreach ($nodeList as $node) {
+            $node->removeAttribute('class');
+        }
     }
 }
--- a/src/Emogrifier/HtmlProcessor/HtmlPruner.php
+++ b/src/Emogrifier/HtmlProcessor/HtmlPruner.php
@@ -41,4 +41,47 @@ class HtmlPruner extends AbstractHtmlProcessor
 
         return $this;
     }
+
+    /**
+     * Removes classes that are no longer required (e.g. because there are no longer any CSS rules that reference them)
+     * from `class` attributes.
+     *
+     * Note that this does not inspect the CSS, but expects to be provided with a list of classes that are still in use.
+     *
+     * This method also has the (presumably beneficial) side-effect of minifying (removing superfluous whitespace from)
+     * `class` attributes.
+     *
+     * @param string[] $classesToKeep list of class names that should not be removed
+     *
+     * @return self fluent interface
+     */
+    public function removeRedundantClasses(array $classesToKeep)
+    {
+        $nodesWithClassAttribute = $this->xPath->query('//*[@class]');
+
+        if ($classesToKeep !== []) {
+            // https://stackoverflow.com/questions/6329211/php-array-intersect-efficiency
+            // It's more efficient to invert the array and use `array_intersect_key()` when doing many intersections.
+            $classesToKeepAsKeys = \array_flip($classesToKeep);
+
+            foreach ($nodesWithClassAttribute as $node) {
+                $nodeClasses = \preg_split('/\\s++/', \trim($node->getAttribute('class')));
+                $nodeClassesToKeep = \array_flip(\array_intersect_key(
+                    \array_flip($nodeClasses),
+                    $classesToKeepAsKeys
+                ));
+                if ($nodeClassesToKeep !== []) {
+                    $node->setAttribute('class', \implode(' ', $nodeClassesToKeep));
+                } else {
+                    $node->removeAttribute('class');
+                }
+            }
+        } else {
+            foreach ($nodesWithClassAttribute as $node) {
+                $node->removeAttribute('class');
+            }
+        }
+
+        return $this;
+    }
 }
--- a/src/Emogrifier/HtmlProcessor/HtmlPruner.php
+++ b/src/Emogrifier/HtmlProcessor/HtmlPruner.php
@@ -54,7 +54,7 @@ class HtmlPruner extends AbstractHtmlProcessor
      * This method also has the (presumably beneficial) side-effect of minifying (removing superfluous whitespace from)
      * `class` attributes.
      *
-     * @param string[] $classesToKeep class names that should not be removed
+     * @param string[] $classesToKeep names of classes that should not be removed
      *
      * @return self fluent interface
      */
--- a/src/Emogrifier/HtmlProcessor/HtmlPruner.php
+++ b/src/Emogrifier/HtmlProcessor/HtmlPruner.php
@@ -60,53 +60,53 @@ class HtmlPruner extends AbstractHtmlProcessor
      */
     public function removeRedundantClasses(array $classesToKeep = [])
     {
-        $nodesWithClassAttribute = $this->xPath->query('//*[@class]');
+        $elementsWithClassAttribute = $this->xPath->query('//*[@class]');
 
         if ($classesToKeep !== []) {
-            $this->removeClassesFromNodes($nodesWithClassAttribute, $classesToKeep);
+            $this->removeClassesFromElements($elementsWithClassAttribute, $classesToKeep);
         } else {
             // Avoid unnecessary processing if there are no classes to keep.
-            $this->removeClassAttributeFromNodes($nodesWithClassAttribute);
+            $this->removeClassAttributeFromElements($elementsWithClassAttribute);
         }
 
         return $this;
     }
 
     /**
-     * Removes classes from the `class` attribute of each element in `$nodeList`, except any in `$classesToKeep`,
+     * Removes classes from the `class` attribute of each element in `$elements`, except any in `$classesToKeep`,
      * removing the `class` attribute itself if the resultant list is empty.
      *
-     * @param \DOMNodeList $nodeList
+     * @param \DOMNodeList $elements
      * @param string[] $classesToKeep
      *
      * @return void
      */
-    private function removeClassesFromNodes(\DOMNodeList $nodeList, array $classesToKeep)
+    private function removeClassesFromElements(\DOMNodeList $elements, array $classesToKeep)
     {
         $classesToKeepIntersector = new ArrayIntersector($classesToKeep);
 
-        foreach ($nodeList as $node) {
-            $nodeClasses = \preg_split('/\\s++/', \trim($node->getAttribute('class')));
-            $nodeClassesToKeep = $classesToKeepIntersector->intersectWith($nodeClasses);
-            if ($nodeClassesToKeep !== []) {
-                $node->setAttribute('class', \implode(' ', $nodeClassesToKeep));
+        foreach ($elements as $element) {
+            $elementClasses = \preg_split('/\\s++/', \trim($element->getAttribute('class')));
+            $elementClassesToKeep = $classesToKeepIntersector->intersectWith($elementClasses);
+            if ($elementClassesToKeep !== []) {
+                $element->setAttribute('class', \implode(' ', $elementClassesToKeep));
             } else {
-                $node->removeAttribute('class');
+                $element->removeAttribute('class');
             }
         }
     }
 
     /**
-     * Removes the `class` attribute from each element in `$nodeList`.
+     * Removes the `class` attribute from each element in `$elements`.
      *
-     * @param \DOMNodeList $nodeList
+     * @param \DOMNodeList $elements
      *
      * @return void
      */
-    private function removeClassAttributeFromNodes(\DOMNodeList $nodeList)
+    private function removeClassAttributeFromElements(\DOMNodeList $elements)
     {
-        foreach ($nodeList as $node) {
-            $node->removeAttribute('class');
+        foreach ($elements as $element) {
+            $element->removeAttribute('class');
         }
     }
 }
--- a/tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
+++ b/tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
@@ -85,4 +85,255 @@ class HtmlPrunerTest extends TestCase
 
         self::assertNotContains('<div', $subject->render());
     }
+
+    /**
+     * @test
+     */
+    public function removeRedundantClassesProvidesFluentInterface()
+    {
+        $subject = HtmlPruner::fromHtml('<html></html>');
+
+        $result = $subject->removeRedundantClasses([]);
+
+        self::assertSame($subject, $result);
+    }
+
+    /**
+     * @return string[][][]
+     */
+    public function classesToKeepDataProvider()
+    {
+        return [
+            'no classes to keep' => [[]],
+            '1 class to keep' => [['foo']],
+        ];
+    }
+
+    /**
+     * @test
+     *
+     * @param string[] $classesToKeep
+     *
+     * @dataProvider classesToKeepDataProvider
+     */
+    public function removeRedundantClassesPreservesHtmlWithoutClasses(array $classesToKeep)
+    {
+        $html = '<p style="color: green;">hello</p>';
+        $subject = HtmlPruner::fromHtml('<html>' . $html . '</html>');
+
+        $subject->removeRedundantClasses($classesToKeep);
+
+        self::assertContains($html, $subject->render());
+    }
+
+    /**
+     * @return (string|string[])[][]
+     */
+    public function nonMatchedClassesDataProvider()
+    {
+        return [
+            '1 attribute, 1 class, no classes to keep' => [
+                'html' => '<p class="foo">hello</p>',
+                'classes to keep' => [],
+                'classes expected to be removed' => ['foo'],
+            ],
+            '2 attributes, 1 different class each, no classes to keep' => [
+                'html' => '<p class="foo">hello</p><p class="bar">world</p>',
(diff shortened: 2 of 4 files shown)


src/Emogrifier/HtmlProcessor/HtmlPruner.php
```diff
@@ -41,4 +41,47 @@ public function removeInvisibleNodes()
 
         return $this;
     }
+
+    /**
```
oliverklee: By the way, would you like to add yourself as an `@author` to the class and the test case?

src/Emogrifier/HtmlProcessor/HtmlPruner.php
```diff
@@ -41,4 +41,47 @@ public function removeInvisibleNodes()
...
+     * Removes classes that are no longer required (e.g. because there are no longer any CSS rules that reference them)
+     * from `class` attributes.
+     *
+     * Note that this does not inspect the CSS, but expects to be provided with a list of classes that are still in use.
+     *
+     * This method also has the (presumably beneficial) side-effect of minifying (removing superfluous whitespace from)
+     * `class` attributes.
+     *
+     * @param string[] $classesToKeep list of class names that should not be removed
```
oliverklee: The "list of" is redundant here.

src/Emogrifier/HtmlProcessor/HtmlPruner.php
```diff
@@ -41,4 +41,47 @@ public function removeInvisibleNodes()
...
+                if ($nodeClassesToKeep !== []) {
+                    $node->setAttribute('class', \implode(' ', $nodeClassesToKeep));
+                } else {
+                    $node->removeAttribute('class');
+                }
+            }
+        } else {
+            foreach ($nodesWithClassAttribute as $node) {
+                $node->removeAttribute('class');
+            }
```
oliverklee: I propose we move both branches into to well-named methods.

src/Emogrifier/HtmlProcessor/HtmlPruner.php
```diff
@@ -41,4 +41,47 @@ public function removeInvisibleNodes()
...
+     * Note that this does not inspect the CSS, but expects to be provided with a list of classes that are still in use.
+     *
+     * This method also has the (presumably beneficial) side-effect of minifying (removing superfluous whitespace from)
+     * `class` attributes.
+     *
+     * @param string[] $classesToKeep list of class names that should not be removed
+     *
+     * @return self fluent interface
+     */
+    public function removeRedundantClasses(array $classesToKeep)
```
oliverklee: Would it make sense to have give `$classesToKepp` a default value of `[]`?

src/Emogrifier/HtmlProcessor/HtmlPruner.php
```diff
@@ -41,4 +41,47 @@ public function removeInvisibleNodes()
...
+    {
+        $nodesWithClassAttribute = $this->xPath->query('//*[@class]');
+
+        if ($classesToKeep !== []) {
```
oliverklee: Do we need this condition? Or is the second branch just a special case of the first one?

src/Emogrifier/HtmlProcessor/HtmlPruner.php
```diff
@@ -41,4 +41,47 @@ public function removeInvisibleNodes()
...
+            // https://stackoverflow.com/questions/6329211/php-array-intersect-efficiency
+            // It's more efficient to invert the array and use `array_intersect_key()` when doing many intersections.
+            $classesToKeepAsKeys = \array_flip($classesToKeep);
+
+            foreach ($nodesWithClassAttribute as $node) {
+                $nodeClasses = \preg_split('/\\s++/', \trim($node->getAttribute('class')));
```
oliverklee: What does the second plus sign mean?

src/Emogrifier/HtmlProcessor/HtmlPruner.php
```diff
@@ -41,4 +41,47 @@ public function removeInvisibleNodes()
...
+                $nodeClassesToKeep = \array_flip(\array_intersect_key(
+                    \array_flip($nodeClasses),
+                    $classesToKeepAsKeys
+                ));
```
oliverklee: I find this hard to grok with the multiple `array_flip`s. Maybe we can wrap this in a well-named method to reduce the level of detail required to understand this?

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -85,4 +85,255 @@ public function removeInvisibleNodesRemovesNodesWithDisplayNone($displayNone)
...
+    }
+
+    /**
+     * @return string[][][]
+     */
+    public function classesToKeepDataProvider()
+    {
+        return [
+            'no classes to keep' => [[]],
+            '1 class to keep' => [['foo']],
```
oliverklee: To cover all equivalence classes, we'll also need a test case for 2 classes to keep (for the minimum version of "multiple classes").

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -85,4 +85,255 @@ public function removeInvisibleNodesRemovesNodesWithDisplayNone($displayNone)
...
+        $html = '<p style="color: green;">hello</p>';
+        $subject = HtmlPruner::fromHtml('<html>' . $html . '</html>');
+
+        $subject->removeRedundantClasses($classesToKeep);
+
+        self::assertContains($html, $subject->render());
+    }
+
+    /**
+     * @return (string|string[])[][]
```
oliverklee: I propose making this more redundant, but easier to read:
`string[][]|string[][][]`

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -85,4 +85,255 @@ public function removeInvisibleNodesRemovesNodesWithDisplayNone($displayNone)
...
+    {
+        $subject = HtmlPruner::fromHtml('<html>' . $html . '</html>');
+
+        $subject->removeRedundantClasses($classesToKeep);
+
+        self::assertNotContains('class', $subject->render());
+    }
+
+    /**
+     * @return (string|string[])[][]
```
oliverklee: Ditto.

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -85,4 +85,255 @@ public function removeInvisibleNodesRemovesNodesWithDisplayNone($displayNone)
...
+     */
+    public function nonMatchedClassesDataProvider()
+    {
+        return [
+            '1 attribute, 1 class, no classes to keep' => [
+                'html' => '<p class="foo">hello</p>',
+                'classes to keep' => [],
+                'classes expected to be removed' => ['foo'],
```
oliverklee: I like having these descriptive sub-array keys.

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -85,4 +85,255 @@ public function removeInvisibleNodesRemovesNodesWithDisplayNone($displayNone)
...
+
+    /**
+     * @test
+     *
+     * @param string $html
+     * @param string[] $classesToKeep
+     *
+     * @dataProvider nonMatchedClassesDataProvider
+     */
+    public function removeRedundantClassesRemovesClassAttributesContainingNoClassesToKeep($html, array $classesToKeep)
```
oliverklee: I find it confusing that the test has only two parameters, but the data provider provides three. I know that the DP gets recycling in another place, and don't have a good idea to solve this at the moment, though.

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -85,4 +85,255 @@ public function removeInvisibleNodesRemovesNodesWithDisplayNone($displayNone)
...
+            ],
+            'last class in attribute is to be removed' => [
+                'html' => '<p class="foo bar baz">hello</p>',
+                'classes to keep' => ['foo', 'bar'],
+                'classes expected to be removed' => ['baz'],
+            ],
+        ];
+        $matchedClassesWithExtraWhitespaceData = $this->matchedClassesWithExtraWhitespaceDataProvider();
+
+        return $nonMatchedClassesData + $matchedClassesData + $matchedClassesWithExtraWhitespaceData;
```
oliverklee: Instead of having a data provider with a different kinds of data sets, I propose we have three separate tests that use each data subset and focus on one aspect of the behavior.

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -85,4 +85,255 @@ public function removeInvisibleNodesRemovesNodesWithDisplayNone($displayNone)
...
+                'classes expected to be removed' => ['bar'],
+            ],
+            '1 attribute, 2 classes separated by newline, both to be kept' => [
+                'html' => "<p class=\"foo\nbar\">hello</p>",
+                'classes to keep' => ['foo', 'bar'],
+            ],
+            '1 attribute, 2 classes separated by newline, 1 to be kept' => [
+                'html' => "<p class=\"foo\nbar\">hello</p>",
+                'classes to keep' => ['foo'],
+                'classes expected to be removed' => ['bar'],
```
oliverklee: Instead of relying on the default values for the parameters, please let's make the third value explicit. Then the reader does not need to know the test method interface to understand the data from the data provider.

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -85,4 +85,255 @@ public function removeInvisibleNodesRemovesNodesWithDisplayNone($displayNone)
...
+                    $expectedInstanceCount,
+                    \substr_count($result, $class),
+                    'asserting \'' . $result . '\' contains ' . $expectedInstanceCount . ' instance(s) of "' . $class
+                    . '"'
+                );
+            }
+        }
+        foreach ($classesExpectedToBeRemoved as $class) {
+            self::assertNotContains($class, $result);
+        }
```
oliverklee: Please let's split this up into separate tests.

src/Emogrifier/HtmlProcessor/HtmlPruner.php
JakeQZ: Sure, will do 😀

src/Emogrifier/HtmlProcessor/HtmlPruner.php
JakeQZ: OK, will remove it.

src/Emogrifier/HtmlProcessor/HtmlPruner.php
JakeQZ: OK, will refactor...

src/Emogrifier/HtmlProcessor/HtmlPruner.php
JakeQZ: Yes, I think so...

src/Emogrifier/HtmlProcessor/HtmlPruner.php
JakeQZ: > Or is the second branch just a special case of the first one?

It is indeed.  The first branch would achieve the same result, but do a whole lot of work for nothing - i.e. for every class attribute, doing a `preg_split`, computing the intersection with an empty array, testing that the result is empty (it of course always will be), then finally deducing that the attribute should be removed.

src/Emogrifier/HtmlProcessor/HtmlPruner.php
JakeQZ: It's the [possessive quantifier](https://www.regular-expressions.info/possessive.html) - it tells the regex engine not to worry about backtracking, so makes it more optimal and helps mitigate against [catastrophic failure](https://www.regular-expressions.info/catastrophic.html).

If the regex was instead `/\s+\n/`, the possessive quantifier could not be used, as the regex engine may need to backtrack to see if one of the whitespace characters matched was in fact a newline - however that regex could be rewritten as `/[^\S\n]++\n/` to ensure the first part won't match a newline, then the possessive quantifier can be used.  Other examples include rewriting `/\(.*\)/` as `/\([^\)]*+\)` (matching something in brackets) - with `/\(.*\)/` the regex engine could have to backtrack a long way to find the closing bracket (or lack of).

src/Emogrifier/HtmlProcessor/HtmlPruner.php
JakeQZ: Or perhaps a simple class could be provided to abstract the detail, so we could have

```php
            $classesToKeepIntersector = new BatchArrayIntersector($classesToKeep);
            ....
                $nodeClassesToKeep = $classesToKeepIntersector->intersectWith($nodeClasses);
```

I thought somebody might have already written one, but [could not find anything](https://www.google.com/search?q=github+simple+php+class+for+batch+array+intersection+using+flipped+arrays+and+array_intersect_key+instead+of+array_intersect).

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: OK, will add it.

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: That wouldn't be correct - it would mean that either all elements of the 2nd level array are strings, or they're all arrays of strings.  Whereas in fact, some of them are strings and some of them are arrays of strings.  The only viable alternative I can think of would be `mixed[][]`...

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: I thought it might be possible, and would help readability, particularly with two arrays of classes.  So I tried it and found it was possible 😎

Though as they are for human consumption only, `'html'` should really be `'HTML'`.  I'll change that...

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: From memory, if we add an unused parameter, PHPCS (or PHPMD) will complain.

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: OK, I'll look into reworking it...

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: OK

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: OK.  We can have `removeRedundantClassesRemovesClassesNotToKeep` and `removeRedundantClassesNotRemovesClassesNotToKeep`.  It may then be realized that not all the datasets are applicable to one or other, which might help with

> Instead of having a data provider with a different kinds of data sets, I propose we have three separate tests that use each data subset and focus on one aspect of the behavior.

JakeQZ:
I've just replied to the comments for now.  I'll look at committing some changes tomorrow... 🌙

JakeQZ:
> I find this hard to grok with the multiple `array_flip`s. Maybe we can wrap this in a well-named method to reduce the level of detail required to understand this?

> Or perhaps a simple class could be provided to abstract the detail

On the basis of the adage “if a job’s worth doing, it’s worth doing well”, I’ve added such a class.  I dropped the prefix `Batch…` as that may imply a method that does the batching, whereas in fact it has a method intended to be called repeatedly as part of some batch processing.  See #710.

src/Emogrifier/HtmlProcessor/HtmlPruner.php
JakeQZ: I'm adding a comment to the code clarifying this.

JakeQZ:
I've made changes based on the review suggestions and repushed.

JakeQZ:
> I've made changes based on the review suggestions and repushed.

In this change I introduced these private methods:

```php
private function removeClassesFromNodes(\DOMNodeList $nodeList, array $classesToKeep);
private function removeClassAttributeFromNodes(\DOMNodeList $nodeList);
```

I chose `…Nodes` in the method name because the parameter is a `DOMNodeList`, but now I think about it, the methods should probably be named `…Elements` because they require that the nodes are elements (only elements can have attributes).  And perhaps then the parameter should be `$elements` rather than `$nodeList`, and the local variables renamed `…node…` to `…element…`...?

(We also have various instances in the code base of private methods and local variables using ‘node’ in the name, where they are necessarily elements, and perhaps should consider changing the names for clarity, but I’d say not for 3.0.0 – except #717.)

oliverklee:
> but now I think about it, the methods should probably be named `…Elements` because they require that the nodes are elements (only elements can have attributes). And perhaps then the parameter should be `$elements` rather than `$nodeList`, and the local variables renamed `…node…` to `…element…...`?

Yes, this makes sense and sounds good.

JakeQZ:
> > the methods should probably be named `…Elements` … the parameter should be `$elements` rather than `$nodeList`, and the local variables renamed `…node…` to `…element…...`?
> 
> Yes, this makes sense and sounds good.

I've renamed accordingly and repushed.

src/Emogrifier/HtmlProcessor/HtmlPruner.php
```diff
@@ -41,4 +44,69 @@ public function removeInvisibleNodes()
 
         return $this;
     }
+
+    /**
+     * Removes classes that are no longer required (e.g. because there are no longer any CSS rules that reference them)
```
oliverklee: Should there be a comma after "e.g."?

src/Emogrifier/HtmlProcessor/HtmlPruner.php
```diff
@@ -41,4 +44,69 @@ public function removeInvisibleNodes()
...
+     * from `class` attributes.
+     *
+     * Note that this does not inspect the CSS, but expects to be provided with a list of classes that are still in use.
+     *
+     * This method also has the (presumably beneficial) side-effect of minifying (removing superfluous whitespace from)
+     * `class` attributes.
+     *
+     * @param string[] $classesToKeep class names that should not be removed
```
oliverklee: Nitpick: class names -> names of classes

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -85,4 +86,281 @@ public function removeInvisibleNodesRemovesNodesWithDisplayNone($displayNone)
...
+
+        $result = $subject->render();
+        foreach ($classesToKeep as $class) {
+            $expectedInstanceCount = \substr_count($html, $class);
+            self::assertSame(
+                $expectedInstanceCount,
+                \substr_count($result, $class),
+                'asserting \'' . $result . '\' contains ' . $expectedInstanceCount . ' instance(s) of "' . $class
+                . '"'
+            );
```
oliverklee: I propose we move this into a well-name static `assert*` method.

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -85,4 +86,281 @@ public function removeInvisibleNodesRemovesNodesWithDisplayNone($displayNone)
...
+     */
+    public function removeRedundantClassesMinifiesClassAttributes($html, array $classesToKeep)
+    {
+        $subject = HtmlPruner::fromHtml('<html>' . $html . '</html>');
+
+        $subject->removeRedundantClasses($classesToKeep);
+
+        \preg_match_all('/class="([^"]*+)"/', $subject->render(), $classAttributeMatches);
+        foreach ($classAttributeMatches[1] as $classAttributeValue) {
+            self::assertNotRegExp('/^\\s|\\s{2}|\\s$/', $classAttributeValue);
```
oliverklee: Can we move this into a static `assert*` method with a descriptive name?

src/Emogrifier/HtmlProcessor/HtmlPruner.php
JakeQZ: There's some discussion on [StackExchange](https://english.stackexchange.com/questions/16172/should-i-always-use-a-comma-after-e-g-or-i-e) that suggests it may be a matter of style, but also an [answer with an example in brackets](https://english.stackexchange.com/questions/16172/should-i-always-use-a-comma-after-e-g-or-i-e#answer-16188) (as here) where there is no comma, and also a [reference link showing a comma only when the example is not in brackets](https://theoatmeal.com/comics/ie).  So I'll leave it as it is.

src/Emogrifier/HtmlProcessor/HtmlPruner.php
JakeQZ: Done.

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: Done.

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: Done.

JakeQZ:
> Just some minor remarks.

Have made suggested changes.

> And could you please rebase?

Well, I used `git rebase master` and it seems to have gone somewhat awry.  Initially I had not done a `git pull` on `master` so it was not up to date.  But after doing so, it kept trying to merge the initial changes from this PR on top of the later changes.  When finally all sorted, I see that the file diff now shows all of the other changes from `master` since this PR was started - exactly what we were trying to avoid!

I've saved the command prompt transcript which I can post to identify what I might have done wrong...

oliverklee:
> Well, I used `git rebase master` and it seems to have gone somewhat awry

I'll do my best to help.

oliverklee:
Some ideas to resolve the git problem:
1. change to master and pull (just to be on the safe side)
1. drop the merge commits from the branch (using `git rebase -i HEAD~25`
1. rebase from master again and resolve any rebase conflicts in the individual commits

oliverklee:
@JakeQZ And if you like, I can also resolve the problem on my machine, force-push and then provide the steps I executed.

oliverklee:
Should this close #380 when merged?

JakeQZ:
>     2. drop the merge commits from the branch (using `git rebase -i HEAD~25` 
>     3. rebase from master again and resolve any rebase conflicts in the individual commits

I think I've sorted it.  It seems that after `git rebase master`, the local copy of the branch must be `--force push`ed.  If an ordinary `push` is attempted, you're told you need to do a `pull` first which then ends up with the entire set of commits on the branch being duplicated (or triplicated), and the 'patching base' not being moved at all.

JakeQZ:
> Should this close #380 when merged?

Not yet - there's the final method which takes a `CssInliner` parameter still to do...

oliverklee:
>  If an ordinary `push` is attempted, you're told you need to do a `pull` first which then ends up with the entire set of commits on the branch being duplicated (or triplicated), and the 'patching base' not being moved at all.

Oh, yes, that message. The action the message recommends is completely bogus. Just last week, I told exactly that to the attendees at a Git workshop. :-)

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -85,4 +86,306 @@ public function removeElementsWithDisplayNoneRemovesElementsWithDisplayNone($dis
...
+    /**
+     * Asserts that the number of occurrences of `$needle` within the string `$haystack` is as expected.
+     *
+     * @param int $expectedCount
+     * @param string $haystack
+     * @param string $needle
+     *
+     * @throws \PHPUnit_Framework_AssertionFailedError
+     */
+    public static function assertSubstringCount($expectedCount, $haystack, $needle)
```
oliverklee: The two helper functions don't need to be public; private should suffice here.

JakeQZ: OK, I've made them private and repushed.

PR #710, opened 2019-09-14
[FEATURE] Add an ArrayIntersector class

JakeQZ:
Part of #708.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+- Add an `ArrayIntersector` class
+  ([#708](https://github.com/MyIntervals/emogrifier/pull/708),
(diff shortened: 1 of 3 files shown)


JakeQZ:
The class is very simple, almost a one-liner.  There is a bit more to the tests, however!

JakeQZ:
It occurs to me now that this may be not be in the best namespace - if `Pelago\Emogrifier\HtmlProcessor` is intended for, um, HTML processors, then perhaps this should be in a namespace like `Pelago\Emogrifier\Utils` or `Pelago\Emogrifier\Utilities`.  And the `Pelago\Emogrifier\CssConcatenator` class may also be better placed in such namespace.

@oliverklee, what do you think?

oliverklee:
I was thinking about this, too. `Pelago\Emogrifier\Utilities` sounds good for both classes.

JakeQZ:
> `Pelago\Emogrifier\Utilities` sounds good for both classes.

OK.  Added #711 for this.

PR #716, opened 2019-09-16
[CLEANUP] Autoformat the code with php-cs-fixer

oliverklee:


tests/Unit/CssInlinerTest.php
```diff
@@ -33,7 +33,7 @@ class CssInlinerTest extends TestCase
                 <p class="p-7"><a
                     href="https://example.org/"
                     data-ascii-1="! &quot; # $ % &amp; &#039; ( ) * + , - . / : ; &lt; = &gt; ?"
-                    data-ascii-2="@ [ \ ] ^ _ ` { | } ~"
+                    data-ascii-2="@ [ \\ ] ^ _ ` { | } ~"
```
JakeQZ: Good catch.  A consequence of using some code to generate the string and copying-and-pasting the result.

PR #722, opened 2019-09-26
[TASK] Test descendant attribute prefix selector

### LINKED ISSUE #443: CSS selector to XPath: Descendant combinator not converted when descendant is an attribute selector

JakeQZ:
Closes #443.  This was actually fixed by #506.  The tests added for `Emogrifier`
confirm this for the specific case; those (same) added for `CssInliner` confirm
there’s no similar issue with the Symfony CssSelector component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -219,7 +219,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ignored, [#507](https://github.com/MyIntervals/emogrifier/pull/507))
 - Allow attribute selectors in descendants
   ([#506](https://github.com/MyIntervals/emogrifier/pull/506),
-  [#381](https://github.com/MyIntervals/emogrifier/issues/381))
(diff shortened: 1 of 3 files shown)


JakeQZ:
I've just updated the CHANGELOG entry for #506 (under 2.1.0) to reference #443 rather than create a new entry.  Is that OK?

PR #724, opened 2019-09-27
[FEATURE] Add HtmlPruner::removeRedundantClassesAfterCssInlined

### LINKED ISSUE #380: remove the class="xxx" from html after converting to inline-css

JakeQZ:
Also added section for `HtmlPruner` to the README.

Closes #380.

### DIFF
--- a/tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
+++ b/tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
@@ -2,6 +2,7 @@
 
 namespace Pelago\Tests\Unit\Emogrifier\HtmlProcessor;
 
+use Pelago\Emogrifier\CssInliner;
 use Pelago\Emogrifier\HtmlProcessor\AbstractHtmlProcessor;
 use Pelago\Emogrifier\HtmlProcessor\HtmlPruner;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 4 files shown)


tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -359,6 +360,182 @@ public function removeRedundantClassesMinifiesClassAttributes($html, array $clas
...
+     * @param string[] $classesToBeKept
+     *
+     * @dataProvider provideClassesInUninlinableRules
+     */
+    public function removeRedundantClassesAfterCssInlinedNotRemovesClassesInUninlinableRules(
+        $html,
+        $css,
+        array $classesToBeKept = []
+    ) {
+        \extract($this->buildSubjectAndCssInlinerWithCssInlined('<html>' . $html . '</html>', $css));
```
oliverklee: Please let's not use `extract` and `compact` as static code analyzers have a really hard time with these functions. Instead, please let's use `list` and explicit arrays.

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -359,6 +360,182 @@ public function removeRedundantClassesMinifiesClassAttributes($html, array $clas
...
+
+        $subject->removeRedundantClassesAfterCssInlined($cssInliner);
+
+        $result = $subject->render();
+        foreach ($classesToBeKept as $class) {
+            self::assertContains($class, $result);
```
oliverklee: I propose we add a `private static assertContainsAll` method for this (and `assertContainsNone`).

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: I was wondering what you'd think about that and aware that using `list` was an alternative solution to the 'multiple return values problem'.  Will change accordingly...

tests/Unit/Emogrifier/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: There might, from memory, be tests in #708 that could also use this.  Will see...

JakeQZ:
Have made suggested changes and repushed...

JakeQZ:
> Looks good! Could you please rebase to resolve the conflict?

Have done.  (Now forewarned and forearmed, `git rebase master` worked seamlessly, with forced push.)

PR #727, opened 2019-09-27
[FEATURE] Check on Travis that PHP-CS-Fixer will not change anything

oliverklee:


composer.json
```diff
@@ -69,6 +69,7 @@
         "php:fix": "php-cs-fixer --config=config/php-cs-fixer.php fix config/ src/ tests/",
         "ci:php:lint": "find config src tests -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l",
         "ci:php:sniff": "phpcs config src tests",
+        "ci:php:fixer": "php-cs-fixer --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff-format=udiff config/ src/ tests/",
```
JakeQZ: Should this not be added to `ci:static`, rather than added to `travis.yml`, so that this check will also run locally?

oliverklee: Oh, good point. I'll add it in a minute.

oliverklee:
> If it works (I've no idea) this will be good. Will the command return an error (non-zero) if it detects an issue?

Yes, it will. (I've tested this on Travis CI.)

.travis.yml
```diff
@@ -62,3 +62,8 @@ script:
   else
     echo "Skipped PHP_CodeSniffer due to insufficient PHP version: $(composer php:version)";
   fi;
+
+- >
+  echo;
+  echo "Running PHP-CS-Fixer";
+  composer ci:php:fixer;
```
JakeQZ: I wonder - as a separate issue/PR - if we might be able to have Travis just run `composer ci`, so that we can avoid the duplication.  The only issue I can see is requiring PHP 7 for CodeSniffer.  Maybe when we've dropped PHP 5.6 support...?

oliverklee: This would make sense.

Currently, the `ci` script runs `ci:static` and `ci:dynamic`. We'd need to change to `ci` script to include the individual scripts, not `ci:static` and `ci:dynamic` anymore so the scripts can keep their current order of execution (which I'd like to do). Currently, the order of the scripts follows these principles:

1. heed the dependencies (e.g., if the PHP linting fails, executing or sniffing the PHP does not make sense anymore)
1. functional checks (e.g., unit tests) before stylistic issues
1. faster checks before slower checks (so the feedback is relatively fast)

JakeQZ: I've added #736 for this.

JakeQZ:
> I'll look forward to it catching me out on something... 😎

Well, it's already found two issues.  But both of them I thought should have been picked up by the CodeSniffer which runs first: missing comma at end of multiline array; extra newline between methods.

I'm pretty sure I remember specifically adjusting the PHPCS config to catch missing commas on multiline arrays, so it might need investigating why it's stopped working...

PR #731, opened 2019-09-28
[CLEANUP] Replace most `static::` with `self::`

### LINKED ISSUE #730: Use `self::` instead of `static::` when overriding is not intended

oliverklee:
`static::` should only be used where overriding the corresponding
method/field is intended.

This makes the intent of the code more clear. (And in addition,
`self::` is a bit shorter than `static::`.)

Fixes #730

### DIFF
--- a/src/Emogrifier/Utilities/CssConcatenator.php
+++ b/src/Emogrifier/Utilities/CssConcatenator.php
@@ -74,7 +74,7 @@ class CssConcatenator
             $lastRuleBlock->selectorsAsKeys += $selectorsAsKeys;
         } else {
             $hasSameSelectorsAsLastRule = $lastRuleBlock !== false
-                && static::hasEquivalentSelectors($selectorsAsKeys, $lastRuleBlock->selectorsAsKeys);
+                && self::hasEquivalentSelectors($selectorsAsKeys, $lastRuleBlock->selectorsAsKeys);
             if ($hasSameSelectorsAsLastRule) {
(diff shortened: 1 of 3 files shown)


src/Emogrifier/Utilities/CssConcatenator.php
```diff
@@ -74,7 +74,7 @@ public function append(array $selectors, $declarationsBlock, $media = '')
             $lastRuleBlock->selectorsAsKeys += $selectorsAsKeys;
         } else {
             $hasSameSelectorsAsLastRule = $lastRuleBlock !== false
-                && static::hasEquivalentSelectors($selectorsAsKeys, $lastRuleBlock->selectorsAsKeys);
+                && self::hasEquivalentSelectors($selectorsAsKeys, $lastRuleBlock->selectorsAsKeys);
```
JakeQZ: My first thought was that I don't think any of these in `CssConcatenator` should be changed.  The proposal in #668 is to allow this class to be extended - there is even a visibility change of a property from `private` to `protected` proposed.  We should allow the methods to be overridden via late static binding.

However, on closer examination, I see that the particular methods are all `private`, so they can't be overridden (unless I'm missing something that wouldn't be the case if they were non-static methods), and there is currently no proposal in #668 to change their visibility.  (That may change, but we can revisit if it does.)

So in conclusion, I think this change is correct for consistency with the (current) methods' visibilities.

PR #732, opened 2019-09-28
[FEATURE] Preserve `@import` rules

JakeQZ:
Implementation for #338 in the `CssInliner` class.


JakeQZ:
I'll apply the equivalent changes to `Emogrifier` when approved for `CssInliner`.

PR #737, opened 2019-09-28
[FEATURE] Keep `display:none` elements with `-emogrifier-keep`

### LINKED ISSUE #252: Invisible node removal: preheaders and responsive design

JakeQZ:
The additional part of the XPath expression was obtained by running
`:not(.-emogrifier-keep)` through `CssSelectorConverter::toXPath()`.  It tests
for attribute presence before checking the value, presumably as an omptimization
to avoid unnecessary processing of a null/empty value.  On the presumed basis
that this is a good idea, a similar check has been added for the `style`
attribute.

Resolves #252.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,24 @@ $prunedHtml = HtmlPruner::fromDomDocument($cssInliner->getDomDocument())
   ->removeRedundantClassesAfterCssInlined($cssInliner)->render();
 ```
 
+The `removeElementsWithDisplayNone` method will not remove any elements which
+have the class `-emogrifier-keep`.  So if, for example, there are elements which
+by default have `display: none` but are revealed by an `@media` rule, or which
+are intended as a preheader, you can add that class to those elements.  The
+paragraph in this HTML snippet will not be removed even though it has
+`display: none` (which has presumably been applied by `CssInliner::inlineCss()`
+from a CSS rule `.preheader { display: none; }`):
+
+```html
+<p class="preheader -emogrifier-keep" style="display: none;">
+	Hello World!
+</p>
+```
+
+The `removeRedundantClassesAfterCssInlined` (or `removeRedundantClasses`)
+method, if invoked after `removeElementsWithDisplayNone`, will remove the
+`-emogrifier-keep` class.
+
 ### Options
 
 There are several options that you can set on the `CssInliner` instance before
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ from a CSS rule `.preheader { display: none; }`):
 
 ```html
 <p class="preheader -emogrifier-keep" style="display: none;">
-	Hello World!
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -189,6 +189,24 @@ $prunedHtml = HtmlPruner::fromDomDocument($cssInliner->getDomDocument())
...
+The `removeElementsWithDisplayNone` method will not remove any elements which
+have the class `-emogrifier-keep`.  So if, for example, there are elements which
+by default have `display: none` but are revealed by an `@media` rule, or which
+are intended as a preheader, you can add that class to those elements.  The
+paragraph in this HTML snippet will not be removed even though it has
+`display: none` (which has presumably been applied by `CssInliner::inlineCss()`
+from a CSS rule `.preheader { display: none; }`):
+
+```html
+<p class="preheader -emogrifier-keep" style="display: none;">
```
oliverklee: I'm wondering: Why the leading hyphen in `-emogrifier-keep`?

JakeQZ: This is to give it a ‘vendor prefix’ as outlined in the [W3C](https://www.w3.org/TR/CSS/#future-proofing) [specification](https://www.w3.org/TR/CSS2/syndata.html#vendor-keywords).

When parsing the CSS, etc., Emogrifier is acting very much as a user agent, and following requirements like “user agents must …” and “user agents should …”.  Although the specification generally refers to using vendor prefixes for keywords and property names, it seems appropriate to extend the principle to proprietary class names.  The leading hyphen should ensure there will not be a clash with any user-defined class names.

oliverklee: Ah, makes sense. Thanks for the explanation.

PR #741, opened 2019-09-29
[CLEANUP] Give an anonymous function a name

### LINKED ISSUE #133: Use fewer anonymous functions for callables

oliverklee:
This will help reduce cognitive load when reading the code, and it will
also help static code analysis.

Fixes #133

### DIFF
--- a/src/Emogrifier.php
+++ b/src/Emogrifier.php
@@ -998,16 +998,7 @@ class Emogrifier
     {
         $css = $cssImportRules;
 
-        $cssRulesRelevantForDocument = \array_filter(
-            $cssRules,
-            function (array $cssRule) {
-                $selector = $cssRule['selector'];
-                if ($cssRule['hasUnmatchablePseudo']) {
-                    $selector = $this->removeUnmatchablePseudoComponents($selector);
-                }
-                return $this->existsMatchForCssSelector($selector);
-            }
-        );
+        $cssRulesRelevantForDocument = \array_filter($cssRules, [$this, 'existsMatchForSelectorInCssRule']);
 
         // avoid including unneeded class dependency if there are no rules
         if ($cssRulesRelevantForDocument !== []) {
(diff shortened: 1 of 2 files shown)


oliverklee:
I'll particularly appreciate feedback on the name of the method.

oliverklee:
The php-cs-fixer CI step has kicked in as intended: https://travis-ci.org/MyIntervals/emogrifier/jobs/591055052

src/Emogrifier.php
```diff
@@ -1030,6 +1021,20 @@ function (array $cssRule) {
         }
     }
 
+    /**
+     * @param string[] $cssRule
+     *
+     * @return bool
+     */
```
JakeQZ: This should have `@throws \InvalidArgumentException` in the PHPDoc.

src/Emogrifier/CssInliner.php
```diff
@@ -660,16 +660,23 @@ public function getMatchingUninlinableSelectors()
...
+    }
+
+    /**
+     * @param string[] $cssRule
+     *
+     * @return bool
+     *
+     * @throws SyntaxErrorException
+     */
+    private function existsMatchForSelectorInCssRule(array $cssRule)
```
JakeQZ: The name looks good and is consistent with the naming of the existing method `existsMatchForCssSelector`.

However, I'm wondering whether it should be clarified (probably in the PHPDoc rather than the name - I'm struggling to put this succinctly) that any dynamic pseudo-classes will be assumed to apply, and that if the selector matches a pseudo-element, it will test for a match with its originating element.

src/Emogrifier.php
oliverklee: Done.

src/Emogrifier/CssInliner.php
oliverklee: I have added a PHPDoc description for this.

oliverklee:
Rebased, improved and repushed.

Issue #747, opened 2019-09-29
Add tests for remaining supported pseudo-classes


JakeQZ:
The following are (presumably) supported by Symfony CssSelector, and the regex `PSEUDO_CLASS_MATCHER` allows them, but we do not have any tests confirming this:

- `:nth-last-child`
- `:nth-last-of-type` (with a type)
- `:only-child`

When confirmed as supported by adding tests, we can list them as supported in the README - see #723.

PR #750, opened 2019-09-29
[TASK] Test :nth-last-child

JakeQZ:
Also updated the regex for supported pseudo-classes (`PSEUDO_CLASS_MATCHER`) so
that those supported are explicitly specified (rather than ‘anything ending with
`-child` or anything ending with `-type(`’), without acually affecting the valid
pseudo-class names it will match.  Except in the `Emogrifier` class – this
doesn’t support `:nth-last-child` so that has been removed (to the effect that
rules with it will now be copied to the `<style>` element rather than attempted
to be inlined).

And added `:nth-last-child` to the list of supported pseudo-classes in the
README.  (Also, for consistency, added `()` to the other `:nth…` pseudo-classes
in that section.)

Part of #747.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+- Test `:nth-last-child`
+  ([#747](https://github.com/MyIntervals/emogrifier/issues/747),
+  [#750](https://github.com/MyIntervals/emogrifier/pull/750))
 - Test general sibling combinator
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+- Test `:nth-last-child`
```
JakeQZ: Thinking about it, it's not just tests being added.  These are new features by virtue of the Symfony CssSelector component.  So these last 3 CHANGELOG entries should in fact read “Support and test …”.  I can change that in a separate PR.

PR #751, opened 2019-09-29
[TASK] Test :nth-last-of-type

JakeQZ:
Also removed it from the regex for supported pseudo-classes
(`Emogrifier::PSEUDO_CLASS_MATCHER`) for the legacy `Emogrifier` class which
does not support it (to the effect that rules with it will now be copied to the
`<style>` element rather than attempted to be inlined).

And added `:nth-last-of-type` to the list of supported pseudo-classes in the
README.

Part of #747.


JakeQZ:
Rebased.

PR #754, opened 2019-09-30
[TASK] Test :only-child

### LINKED ISSUE #747: Add tests for remaining supported pseudo-classes

JakeQZ:
And updated README to indicate it is supported.

Also removed ‘support’ for `:only-child` from `Emogrifier` class by adjusting
the `PSEUDO_CLASS_MATCHER` regex, so that rules with such selectors will be
copied to a `<style>` element.  (It is not supported in that class.)

Closes #747.


JakeQZ:
> Could you please rebase?

Have done.

Issue #748, opened 2019-09-29
Support all pseudo-classes that Symfony CssSelector does


JakeQZ:
The following are (presumably) supported by Symfony CssSelector, but the regex in `PSEUDO_CLASS_MATCHER` does not allow them:

- `:any-link`
- `:empty`
- `:first-of-type` (with a type)
- `:last-of-type` (with a type)
- `:only-of-type` (with a type)
- `:optional`
- `:required`

We should add tests to confirm they are supported, and, if indeed they are, modify the regex to allow them.  ~If any are in fact not supported, we should ensure the regex does not allow them (so that rules using them are copied to the `<style>` element).~  UPDATE: Since #750 the regex now only allows pseudo-classes that are knowingly supported.

When supported, we can list them as supported in the README - see #723.

PR #757, opened 2019-09-30
[FEATURE] Support: first-of-type

JakeQZ:
Part of #748.

PR #758, opened 2019-09-30
[FEATURE] Support :last-of-type

JakeQZ:
Part of #748.


JakeQZ:
Will need rebasing after #756 and #757 are merged.

PR #762, opened 2019-09-30
[TASK] Relax the dependency on symfony/css-selector

oliverklee:
This reduces the probability of dependency conflicts with other packages.

`symfony/css-selector` 2.8 is the first version that provides the
functionality required by Emogrifier.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
         "php": "^5.6.0 || ~7.0.0 || ~7.1.0 || ~7.2.0 || ~7.3.0",
         "ext-dom": "*",
         "ext-libxml": "*",
-        "symfony/css-selector": "^3.4.0 || ^4.0.0"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -40,7 +40,7 @@
         "php": "^5.6.0 || ~7.0.0 || ~7.1.0 || ~7.2.0 || ~7.3.0",
         "ext-dom": "*",
         "ext-libxml": "*",
-        "symfony/css-selector": "^3.4.0 || ^4.0.0"
+        "symfony/css-selector": "^2.8 | ^3.0 || ^4.0"
```
JakeQZ: This should have `||` rather than `|` - the single pipe though it will currently work is deprecated - see https://github.com/composer/composer/issues/6755.

oliverklee: Oh, that was a typo. Good catch! On it.

oliverklee:
Repushed.

PR #764, opened 2019-09-30
[TASK] Only render the body for most tests

### LINKED ISSUE #213: Change most CssInliner tests to use renderBodyContent
This should make the tests a lot shorter.

oliverklee:
Tests that are not about the complete rendered HTML (and also not
explictly about the `render()` method) now use the `renderBodyContent()`
method instead in order to better communicate that they are not about
the complete HTML.

Resolves #213

### DIFF
--- a/tests/Unit/Emogrifier/HtmlProcessor/CssToAttributeConverterTest.php
+++ b/tests/Unit/Emogrifier/HtmlProcessor/CssToAttributeConverterTest.php
@@ -48,7 +48,7 @@ class CssToAttributeConverterTest extends TestCase
     /**
      * @test
      */
-    public function renderWithoutConvertCssToVisualAttributesCallNotAddsVisuablAttributes()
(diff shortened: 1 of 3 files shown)


oliverklee:
Oh, this caused some lines to be too long. On it.

tests/Unit/Emogrifier/HtmlProcessor/CssToAttributeConverterTest.php
```diff
@@ -48,7 +48,7 @@ public function fromDomDocumentReturnsInstanceOfCalledClass()
     /**
      * @test
      */
-    public function renderWithoutConvertCssToVisualAttributesCallNotAddsVisuablAttributes()
+    public function renderWithoutConvertCssToVisualAttributesCallNotAddsVisualAttributes()
```
JakeQZ: Good spot!

PR #765, opened 2019-09-30
[TASK] Test :optional and :required

### LINKED ISSUE #748: Support all pseudo-classes that Symfony CssSelector does

JakeQZ:
Added a “broken: …” comment placeholder to indicate that support for `:required`
and `:optional` with the Symfony CssSelector component has been checked and
found to be lacking, suggesting tests to be added if and when it is supported in
future.

Added tests to confirm rules with `:required` or `:optional` are nonetheless
copied to a `<style>` element.

In the README, added `:required` and `:optional` to the list of unsupported
static pseudo-classes that are copied to a `<style>` element rather than being
inlined.

Closes #748.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+- Test that rules with `:optional` or `:required` are copied to the `<style>`
+  element ([#748](https://github.com/MyIntervals/emogrifier/issues/748),
+  [#764](https://github.com/MyIntervals/emogrifier/pull/764))
 - Test that rules with `:only-of-type` are copied to the `<style>` element
   ([#748](https://github.com/MyIntervals/emogrifier/issues/748),
   [#760](https://github.com/MyIntervals/emogrifier/pull/760))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 - Test that rules with `:optional` or `:required` are copied to the `<style>`
   element ([#748](https://github.com/MyIntervals/emogrifier/issues/748),
-  [#764](https://github.com/MyIntervals/emogrifier/pull/764))
+  [#765](https://github.com/MyIntervals/emogrifier/pull/765))
 - Test that rules with `:only-of-type` are copied to the `<style>` element
   ([#748](https://github.com/MyIntervals/emogrifier/issues/748),
   [#760](https://github.com/MyIntervals/emogrifier/pull/760))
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+- Test that rules with `:optional` or `:required` are copied to the `<style>`
+  element ([#748](https://github.com/MyIntervals/emogrifier/issues/748),
+  [#764](https://github.com/MyIntervals/emogrifier/pull/764))
```
oliverklee: Looks like I was faster with my PR by a few seconds. So this is not 765, not 764. :-)

JakeQZ: Corrected 😉

tests/Unit/CssInlinerTest.php
```diff
@@ -2113,7 +2117,7 @@ public function matchingSelectorWithPseudoComponentCssRuleDataProvider()
      */
     public function inlineCssKeepsRuleWithPseudoComponentInMatchingSelector($css)
     {
-        $subject = $this->buildDebugSubject('<html><p><a id="a" class="a" href="a">foo</a></p></html>');
+        $subject = $this->buildDebugSubject('<html><p><a id="a" class="a" href="a">foo</a><input></p></html>');
```
oliverklee: I'd prefer to have the `input` tag explicitly self-closing, i.e., to change it from `<input>` to `<input/>`. And while we're at it, can we add the required attributes to the `input` attribut eso the HTML would make sense in the real world, e.g. `<input type="text"/>?

tests/Unit/EmogrifierTest.php
```diff
@@ -2075,7 +2077,7 @@ public function matchingSelectorWithPseudoComponentCssRuleDataProvider()
      */
     public function emogrifyKeepsRuleWithPseudoComponentInMatchingSelector($css)
     {
-        $this->subject->setHtml('<html><p><a id="a" class="a" href="a">foo</a></p></html>');
+        $this->subject->setHtml('<html><p><a id="a" class="a" href="a">foo</a><input></p></html>');
```
oliverklee: Ditto.

oliverklee:
> Made `<input>` element more real-worldy

:-D

PR #767, opened 2019-09-30
[TASK] Also sniff the code against the PSR-12 standard

oliverklee:
The latest version of PHP_CodeSniffer now supports PSR-12,
which supersedes PSR-2.

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -6,8 +6,10 @@
 
     <config name="installed_paths" value="../../slevomat/coding-standard"/>
 
-    <!--The complete PSR-2 ruleset-->
-    <rule ref="PSR2"/>
+    <!-- The complete PSR-12 ruleset (excluding a rule that requires PHP >= 7.1 -->
+    <rule ref="PSR12">
+        <exclude name="PSR12.Properties.ConstantVisibility"/>
+    </rule>
 
     <!-- Arrays -->
     <rule ref="Generic.Arrays.DisallowLongArraySyntax"/>
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -6,8 +6,9 @@
 
     <config name="installed_paths" value="../../slevomat/coding-standard"/>
 
-    <!-- The complete PSR-12 ruleset (excluding a rule that requires PHP >= 7.1 -->
+    <!-- The complete PSR-12 ruleset -->
     <rule ref="PSR12">
+        <!-- Exclude a rule that requires PHP >= 7.1. -->
         <exclude name="PSR12.Properties.ConstantVisibility"/>
     </rule>
 
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -1608,13 +1608,18 @@ class CssInlinerTest extends TestCase
                 foreach ($possibleSurroundingCss as $descriptionAfter => $cssAfter) {
                     // every combination would be a ridiculous c.1000 datasets - choose a select few
                     // test all possible CSS before once
-                    if (($cssBetween === '' && $cssAfter === '')
+                    if (
+                        ($cssBetween === ''
+                        && $cssAfter === '')
                         // test all possible CSS between once
-                        || ($cssBefore === '' && $cssAfter === '')
+                        || ($cssBefore === ''
+                        && $cssAfter === '')
                         // test all possible CSS after once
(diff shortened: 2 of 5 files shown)


oliverklee:
This still needs some work, it seems …

oliverklee:
… and maybe wait for the next bug-fix release of PHP_CodeSniffer.

phpcs.xml.dist
```diff
@@ -6,8 +6,10 @@
 
     <config name="installed_paths" value="../../slevomat/coding-standard"/>
 
-    <!--The complete PSR-2 ruleset-->
-    <rule ref="PSR2"/>
+    <!-- The complete PSR-12 ruleset (excluding a rule that requires PHP >= 7.1 -->
+    <rule ref="PSR12">
+        <exclude name="PSR12.Properties.ConstantVisibility"/>
```
JakeQZ: I should have read the comment above :/.  I just wasted a few minutes trying to understand what this was about.

phpcs.xml.dist
```diff
@@ -6,8 +6,10 @@
 
     <config name="installed_paths" value="../../slevomat/coding-standard"/>
 
-    <!--The complete PSR-2 ruleset-->
-    <rule ref="PSR2"/>
+    <!-- The complete PSR-12 ruleset (excluding a rule that requires PHP >= 7.1 -->
```
JakeQZ: Missing ")" but maybe the comment for constant visibility should be separated and moved just before the relevant `<exclude>`...

tests/Unit/CssInlinerTest.php
```diff
@@ -1606,13 +1606,18 @@ public function orderedRulesAndSurroundingCssDataProvider()
                 foreach ($possibleSurroundingCss as $descriptionAfter => $cssAfter) {
                     // every combination would be a ridiculous c.1000 datasets - choose a select few
                     // test all possible CSS before once
-                    if (($cssBetween === '' && $cssAfter === '')
+                    if (
```
JakeQZ: Could this be resolved by merely adding a newline after `if (`...?

tests/Unit/CssInlinerTest.php
```diff
@@ -1606,13 +1606,18 @@ public function orderedRulesAndSurroundingCssDataProvider()
...
+                        ($cssBetween === ''
+                        && $cssAfter === '')
```
JakeQZ: ... I find having a newline within the bracketed `AND` expressions makes the code less readable, not more.  Is this really not allowed?

tests/Unit/EmogrifierTest.php
```diff
@@ -1547,13 +1547,18 @@ public function orderedRulesAndSurroundingCssDataProvider()
                 foreach ($possibleSurroundingCss as $descriptionAfter => $cssAfter) {
                     // every combination would be a ridiculous c.1000 datasets - choose a select few
                     // test all possible CSS before once
-                    if (($cssBetween === '' && $cssAfter === '')
+                    if (
+                        ($cssBetween === ''
+                        && $cssAfter === '')
```
JakeQZ: Ditto.

tests/Unit/CssInlinerTest.php
oliverklee: I'll test this.

This is the relevant section from the standard: https://www.php-fig.org/psr/psr-12/#51-if-elseif-else

oliverklee:
This is the relevant section from the standard: https://www.php-fig.org/psr/psr-12/#51-if-elseif-else

phpcs.xml.dist
oliverklee: Done.

tests/Unit/CssInlinerTest.php
oliverklee: Unfortunately, this did not work. :-(

oliverklee: We could use introduce some local variables (in a separate PR) to make the condition less enormous, though.

oliverklee:
Updated and repushed.

tests/Unit/CssInlinerTest.php
JakeQZ: > Unfortunately, this did not work. :-(

OK.

> We could use introduce some local variables (in a separate PR) to make the condition less enormous, though.

That would help make it more readable and perhaps also avoid the need for the comments within the `if` expression.

PR #770, opened 2019-10-01
[BUGFIX] Fix PhpStorm code inspection warnings

### LINKED ISSUE #729: FIx PhpStorm code inspection warnings
… including the checks provided by "PHP Inspections (EA Ultimate)"

oliverklee:
Fixes #729


src/Emogrifier.php
```diff
@@ -843,7 +843,7 @@ private function normalizeStyleAttributesOfAllNodes()
     private function normalizeStyleAttributes(\DOMElement $node)
     {
         $normalizedOriginalStyle = \preg_replace_callback(
-            '/[A-z\\-]+(?=\\:)/S',
+            '/[a-zA-Z\\-]+(?=:)/S',
```
JakeQZ: Good catch - it happened to work more by luck becuase `:` is before `A` in ASCII.

However, I don't think the correction is quite right - AKAIK property names may also contain underscores or numbers (though I'm not aware of any), if they are classed as [CSS identifiers](https://www.w3.org/TR/CSS21/syndata.html#characters) - the regex used in `HtmlPruner::removeRedundantClassesAfterCssInlined` matches a CSS identifier (after a period).

src/Emogrifier.php
```diff
@@ -1003,7 +1003,7 @@ private function copyUninlinableCssToStyleNode(array $cssRules, $cssImportRules)
         // avoid including unneeded class dependency if there are no rules
         if ($cssRulesRelevantForDocument !== []) {
             // support use without autoload
-            if (!\class_exists('Pelago\\Emogrifier\\Utilities\\CssConcatenator')) {
+            if (!\class_exists(CssConcatenator::class)) {
```
JakeQZ: This won't work - if autoloading is not used (which is the reason for its existence in the first place), it will cause an error that `CssConcatenator` is not defined.

src/Emogrifier/CssInliner.php
```diff
@@ -380,7 +380,7 @@ private function getAllNodesWithStyleAttribute()
     private function normalizeStyleAttributes(\DOMElement $node)
     {
         $normalizedOriginalStyle = \preg_replace_callback(
-            '/[A-z\\-]+(?=\\:)/S',
+            '/[a-zA-Z\\-]+(?=:)/S',
```
JakeQZ: Possible issue as with the same regex in `Emogrifier`.

src/Emogrifier.php
oliverklee: Yes, this will work. The `::class` notation is only a shortcut for the name (as imported via the `use` statement above or the current namespace). This also works even if the class cannot be loaded or does not exist because this is on the naming level only.

JakeQZ: OK.  You're right.  (For peace of mind, I tested and confirmed, at least for PHP 7.3.)

JakeQZ:
> Looks generally fine, but there are a couple of issues - in particular with regard maintaining support for the `Emogrifier` class without an autoloader.

The autloader issue is a non-issue, but do you think the regex for property names should be updated to better match the spec?

src/Emogrifier.php
oliverklee: Rebased, changed accordingly and repushed.

PR #773, opened 2019-10-01
[TASK] Drop support for PHP 5.6

oliverklee:
Now PHP >= 7.0 is required.

Changes to the code and to the php-cs-fixer configuration will come
in separate changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.15.3",
         "squizlabs/php_codesniffer": "^3.5.0",
-        "slevomat/coding-standard": "5.0.4",
+        "slevomat/coding-standard": "^5.0.4",
         "phpmd/phpmd": "^2.7.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -37,14 +37,15 @@
...
-        "php": "^5.6.0 || ~7.0.0 || ~7.1.0 || ~7.2.0 || ~7.3.0",
+        "php": "~7.0.0 || ~7.1.0 || ~7.2.0 || ~7.3.0",
         "ext-dom": "*",
         "ext-libxml": "*",
         "symfony/css-selector": "^2.8 || ^3.0 || ^4.0"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.15.3",
         "squizlabs/php_codesniffer": "^3.5.0",
+        "slevomat/coding-standard": "5.0.4",
```
JakeQZ: Should this not be with a `^`?

oliverklee: Indeed. On it.

oliverklee: Done and repushed.

PR #777, opened 2019-10-01
[CLEANUP] Make use of PHP 7.0 language features

oliverklee:


oliverklee:
Hold on, I think we can enable even more sniffs here. On it.

JakeQZ:
> Hold on, I think we can enable even more sniffs here. On it.

OK...

oliverklee:
Done.

src/Emogrifier/CssInliner.php
```diff
@@ -154,7 +156,7 @@ class CssInliner extends AbstractHtmlProcessor
      *
      * @throws SyntaxErrorException
      */
-    public function inlineCss($css = '')
+    public function inlineCss(string $css = ''): self
```
JakeQZ: Just a query, is the coding standard specifically to have no space between `)` and `:` for the return value type hint?  (Personally, I'd prefer a space, but I can get used to its absence.)

src/Emogrifier/CssInliner.php
```diff
@@ -233,7 +235,7 @@ public function disableStyleBlocksParsing()
      *
      * @return void
      */
-    public function addAllowedMediaType($mediaName)
+    public function addAllowedMediaType(string $mediaName)
```
JakeQZ: `void` return value type hints will have to wait until PHP 7.1 :/

src/Emogrifier/CssInliner.php
```diff
@@ -781,7 +783,7 @@ private function splitCssAndMediaQuery($css)
      */
     private function copyInlinableCssToStyleAttribute(\DOMElement $node, array $cssRule)
     {
-        $newStyleDeclarations = $this->parseCssDeclarationsBlock($cssRule['declarationsBlock']);
+        $newStyleDeclarations = $this->parseCssDeclarationsBlock((string)$cssRule['declarationsBlock']);
```
JakeQZ: In what situation is `$cssRule['declarationsBlock']` not a `string`?  Maybe we need to look into that...

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -58,11 +60,8 @@ private function __construct()
      *
      * @throws \InvalidArgumentException if $unprocessedHtml is anything other than a non-empty string
      */
-    public static function fromHtml($unprocessedHtml)
+    public static function fromHtml(string $unprocessedHtml): AbstractHtmlProcessor
```
JakeQZ: Is `: static` not possible?

tests/Unit/CssInlinerTest.php
```diff
@@ -1614,16 +1616,16 @@ public function orderedRulesAndSurroundingCssDataProvider()
                     // test all possible CSS before once
                     if (
                         ($cssBetween === ''
-                        && $cssAfter === '')
+                            && $cssAfter === '')
```
JakeQZ: This is slightly more readable that with the previous change in #767 :)

src/Emogrifier/CssInliner.php
oliverklee: It will always be a string, but the static code analysis of PhpStorm does not know this and hence needs some help …

tests/Unit/CssInlinerTest.php
oliverklee: ;-)

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
oliverklee: Oh, that should indeed be `static`. I think that was some fallout of the automatic code conversion.

oliverklee: I've just tested it: `: static` indeed does not work. :-(

src/Emogrifier/CssInliner.php
oliverklee: From the PSR-12 document:

> The colon and declaration MUST be on the same line as the argument list closing parenthesis with no spaces between the two characters.

src/Emogrifier/CssInliner.php
JakeQZ: Is there no other way to tell PhpStorm, e.g. with a `/** @var string[] $cssRule */` preceding, or perpahs `/** @var string $declarationsBlock */ $declarationsBlock = cssRule['declarationsBlock']`?

I don't like the idea of having an unncessary cast in the actual PHP code in order to work around an issue with an IDE analysis tool.

oliverklee: We could introduce a local variable for this and add a type annotation for it. Would that be a viable solution?

oliverklee: Oh, I see that's actually the second solution you have proposed. I'll implement it in a second.

src/Emogrifier/CssInliner.php
JakeQZ: Fair enough :)

oliverklee:
Updated and repushed.

src/Emogrifier/CssInliner.php
JakeQZ: > /** @var string[] $cssRule */

That first suggestion may be not so good, as they're probably not all strings - and from what I've read PhpStorm possibly doesn't support the `(string|int)[]` syntax.

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
JakeQZ: :/ Maybe in a future PHP version,

src/Emogrifier/CssInliner.php
```diff
@@ -783,7 +783,9 @@ private function splitCssAndMediaQuery(string $css): array
      */
     private function copyInlinableCssToStyleAttribute(\DOMElement $node, array $cssRule)
     {
-        $newStyleDeclarations = $this->parseCssDeclarationsBlock((string)$cssRule['declarationsBlock']);
+        /** @var string $declarationsBlock */
```
JakeQZ: Does it need newlines in the DocBlock like
```php
/**
 * @var string $declarationsBlock
 */
```
or does it also work as a single line?

oliverklee: Single-line doc blocks work just as fine.

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
JakeQZ: But would `: self` work...?

oliverklee: It would. :-) #784

Issue #742, opened 2019-09-29
Use dedicated classes instead of arrays with special keys


oliverklee:
This will make the code more explicit and help static code analysis.

Particularly, we should get rid of `compact`, `list` and family.

SignpostMarv:
it appears that `CssConcatenator::getOrCreateMediaRuleToAppendTo()` may require a dedicated class in order to satisfy some `MixedPropertyFetch` issues in psalm:
https://github.com/MyIntervals/emogrifier/blob/6f9d875f0054ab3d8cd6a972a7d517ced19a2040/src/Utilities/CssConcatenator.php#L103

Issue #537, opened 2018-03-23
Check the code with psalm


oliverklee:
https://github.com/vimeo/psalm

SignpostMarv:
psalm has identified a set of `TestCase::assertInstanceOf()` checks as redundant, which renders three test cases redundant:
* https://github.com/MyIntervals/emogrifier/blob/b8fa86b0fa6d4424e9ab4234587065c01f085ee1/tests/Unit/HtmlProcessor/HtmlPrunerTest.php#L24
* https://github.com/MyIntervals/emogrifier/blob/b8fa86b0fa6d4424e9ab4234587065c01f085ee1/tests/Unit/HtmlProcessor/HtmlPrunerTest.php#L34
* https://github.com/MyIntervals/emogrifier/blob/b8fa86b0fa6d4424e9ab4234587065c01f085ee1/tests/Unit/HtmlProcessor/HtmlPrunerTest.php#L44

This is likely because the corresponding methods are flags as `@return static` and so long as psalm doesn't identify a different type in these methods, it *should* be safe to remove the tests.
* https://github.com/MyIntervals/emogrifier/blob/b8fa86b0fa6d4424e9ab4234587065c01f085ee1/src/HtmlProcessor/AbstractHtmlProcessor.php#L63
* https://github.com/MyIntervals/emogrifier/blob/b8fa86b0fa6d4424e9ab4234587065c01f085ee1/src/HtmlProcessor/AbstractHtmlProcessor.php#L82

JakeQZ:
It's also possibly because we're asserting something that's provably true by static code analysis!

SignpostMarv:
the question would be whether to leave the assertions in to enforce future behaviour, or remove all three test methods as the static analysis "proves" the appropriate type is in play.

JakeQZ:
I'd prefer to have more checks than fewer, even if they overlap/duplicate, i.e. keep the assertions.

In a similar vein, I recall a comment about Psalm a few days ago where you mentioned something that was already covered by PHPCS (or -fixer).  I'd like to have all the tools cover all the possible issues, regardless of overlap.  That way you can be more sure nothing is missed (one tool may be better in different ways at picking up a particular issue than another).

SignpostMarv:
related query to checking with psalm; whether or not to enable the [shepherd](https://shepherd.dev/) plugin to have the type-coverage badge in readme.md

oliverklee:
[...]

PR #778, opened 2019-10-01
Psalm

SignpostMarv:


oliverklee:
This will probably greatly benefit from #777.

SignpostMarv:
So there's an issue in that slevomat/coding-standard 5.0.4 requires php 7.1 & there's a sniff used that isn't in the latest version that supports php 7.0.

Rather than tinkering with the travis config, I'd recommend dropping php 7.0, given that it's been EOL for 8 months.

oliverklee:
> So there's an issue in that slevomat/coding-standard 5.0.4 requires php 7.1 & there's a sniff used that isn't in the latest version that supports php 7.0.

If that's the case,  then we need to revert part of #773. I wonder why the Travis build has not failed, though.

As for the minimum PHP version, we plan to require PHP >= 7.1 for Emogrifier 5.0 (i.e., for the next-after-next breaking release). As we have announced that PHP >= 7.0 will suffice for Emogrifier 4.0.0, I'd like to not break that promise.

(We could release 4.0.0 quite soon so we can drop PHP 7.0, though.)

SignpostMarv:
it's also worth noting that 7.1 goes EOL as of next year.

oliverklee:
I have created #779 to resolve the problem with PHP 7.0.

SignpostMarv:
I'll rebase once that's merged :)

oliverklee:
Thanks, that's perfectly fine. There is no hurry.

By the way, I'd prefer to have relatively low-impact PRs, e.g., one for adding psalm to the CI chain (without any code changes) and for adding a baseline, then others for fixing issues found by psalm, another for adding psalm-specific annotations etc.

SignpostMarv:
> By the way, I'd prefer to have relatively low-impact PRs, e.g.,

split ef33fbf into 3 on the rebase?

oliverklee:
> split ef33fbf into 3 on the rebase?

Yes, thanks, that would be great.

composer.json
```diff
@@ -37,17 +37,19 @@
...
         "ext-libxml": "*",
         "symfony/css-selector": "^2.8 || ^3.0 || ^4.0"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.15.3",
-        "squizlabs/php_codesniffer": "^3.5.0",
-        "slevomat/coding-standard": "^5.0.4",
         "phpmd/phpmd": "^2.7.0",
-        "phpunit/phpunit": "^5.7.27"
+        "phpunit/phpunit": "^7.5",
```
JakeQZ: Now that we've dropped PHP 5.x support, we can at least update PHPUnit to `^6.5.14` and use the namespaced classes.  However, PHPUnit 7.x requires PHP 7.1.

composer.json
```diff
@@ -37,17 +37,19 @@
...
-        "php": "~7.0.0 || ~7.1.0 || ~7.2.0 || ~7.3.0",
+        "php": "~7.1.0 || ~7.2.0 || ~7.3.0",
         "ext-dom": "*",
         "ext-libxml": "*",
         "symfony/css-selector": "^2.8 || ^3.0 || ^4.0"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.15.3",
-        "squizlabs/php_codesniffer": "^3.5.0",
-        "slevomat/coding-standard": "^5.0.4",
```
JakeQZ: Agree alphabetical order is better than, um, random.

composer.json
```diff
@@ -37,17 +37,19 @@
...
         "phpmd/phpmd": "^2.7.0",
-        "phpunit/phpunit": "^5.7.27"
+        "phpunit/phpunit": "^7.5",
+        "psalm/plugin-phpunit": "^0.7.0",
+        "slevomat/coding-standard": "^5.0.4",
+        "squizlabs/php_codesniffer": "^3.5.0",
+        "vimeo/psalm": "^3.5"
```
JakeQZ: As `psalm/plugin-phpunit` depends on `vimeo/psalm`, do we need this?

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -300,14 +296,32 @@ private function ensurePhpUnrecognizedSelfClosingTagsAreXml($html)
...
-        $htmlElement->appendChild($this->domDocument->createElement('body'));
+        $htmlElement->appendChild($this->getDomDocument()->createElement('body'));
+    }
+
+    /**
+     * Ensures that AbstractHtmlProcessor::$xPath has been set
+     *
+     * @throws \UnexpectedValueException if AbstractHtmlProcessor::$xPath has not been set
+     */
+    protected function ensureXPath(): \DOMXPath
```
JakeQZ: I think `getXPath` would be a better name – even though it may throw an exception, it still (normally) does exactly what a ‘getter’ does.  @oliverklee, what do you think?

composer.json
oliverklee: I have created #783 for the PHPUnit upgrade.

composer.json
oliverklee: I'd like to have dependencies explicit here and generally not depend on transitive dependencies.

SignpostMarv: If you ever added maglnet/composer-require-checker to your ci checklist, you'd be prompted to explicitly require it.

composer.json
SignpostMarv: I'll make a separate PR for the config change, unless someone else makes the change while I'm at the day job?

composer.json
SignpostMarv: There's an earlier commit that limited psalm to 3.2 so it could be used alongside php 7.0.

composer.json
JakeQZ: Hmm... What if `psalm/plugin-phpunit` 0.7.1 decides that it requires `vimeo/psalm` 4.0?  I've seen things like this happen.  Our build would suddenly become broken despite us not changing anything.  That's why I suggest that we only require the immediate dependencies, and let them figure out what they need in turn.  It is however possible to use [suggests](https://getcomposer.org/doc/03-cli.md#suggests) with Composer, which may be the solution here.

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
JakeQZ: From the two thumbs-ups, seems we are in agreement here 😎

composer.json
SignpostMarv: > What if `psalm/plugin-phpunit` 0.7.1 decides that it requires `vimeo/psalm` 4.0?

then it wouldn't be installed if vimeo/psalm was required as "^3.x"

JakeQZ: > > What if `psalm/plugin-phpunit` 0.7.1 decides that it requires `vimeo/psalm` 4.0?
> 
> then it wouldn't be installed if vimeo/psalm was required as "^3.x"

That's my point.  Composer would 'error' because it could not satisfy the dependencies.

JakeQZ: > > > What if `psalm/plugin-phpunit` 0.7.1 decides that it requires `vimeo/psalm` 4.0?
> > 
> > 
> > then it wouldn't be installed if vimeo/psalm was required as "^3.x"
> 
> That's my point. Composer would 'error' because it could not satisfy the dependencies. And even if allowed to continue, the build would fall over.

But that might be exactlty what you want...

JakeQZ: But you could end up with a mysteriously broken build despite having not made any changes.  Or more likely, you've made a change which has triggered the build and you're tearing you hair out trying to figure out just how you've broken it.

JakeQZ: But I concede 😣

SignpostMarv: > That's my point. Composer would 'error' because it could not satisfy the dependencies. And even if allowed to continue, the build would fall over.

Not sure how it would error? composer's conflict resolution would see the constraint of ^0.7, see that the hypothetical 0.7.1 requested 4.x & 0.7.0 requested 3.5.x, then see that psalm 3.5 was also required, and thus install psalm 3.5.x alongside the phpunit plugin 0.7.0.

not sure I've understood your argument :s

JakeQZ: > Not sure how it would error? composer's conflict resolution would see the constraint of ^0.7, see that the hypothetical 0.7.1 requested 4.x & 0.7.0 requested 3.5.x, then see that psalm 3.5 was also required, and thus install psalm 3.5.x alongside the phpunit plugin 0.7.0

Good point.  Struggling for a logical argument, next best I can come up with is that suppose `psalm/plugin-phpunit` v0.8.0 no longer requires `vimeo/psalm`.  Then we are left with an orphaned dependency.

I think I know from experience that having this theoretically unnecessary dependency could lead to problems in future.  Precisely what experience that was I'm finding hard to pinpoint/recollect...

composer.json
oliverklee: Go for it! :coffee:

composer.json
SignpostMarv: > Struggling for a logical argument, next best I can come up with is that suppose `psalm/plugin-phpunit` v0.8.0 no longer requires `vimeo/psalm`.

Not likely in this particular case (it being a plugin for psalm that it essentially suppresses a bunch of issues that crop up when using phpunit, i.e. dataProvider methods being flagged as unused because they're not explicitly called etc.)).

> Precisely what experience that was I'm finding hard to pinpoint/recollect...

I do know what you mean here though- every now and then I have to review all my dependencies to ensure they're actually needed.

JakeQZ: > > Precisely what experience that was I'm finding hard to pinpoint/recollect...
> 
> I do know what you mean here though- every now and then I have to review all my dependencies to ensure they're actually needed.

This is a more general point about [dependecy hell](https://en.wikipedia.org/wiki/Dependency_hell) (which Composer attempts to address), and the fact that I've experienced the ‘hell’, but right now can’t recall details of specific examples.

composer.json
SignpostMarv: see #788

composer.json
SignpostMarv: negated by rebased sprint of psalm implementation.

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
SignpostMarv: implemented on rebased psalm sprint.

SignpostMarv:
This is why I generally commit all psalm bug fixes en-masse.

psalm.baseline.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files>
+  <file src="tests\Unit\Utilities\CssConcatenatorTest.php">
+    <PossiblyNullPropertyAssignmentValue occurrences="1">
```
SignpostMarv: iirc, this particular issue relates to a class property being set in `TestCase::setUp()` rather than a constructor.

what I can't recall is whether or not that particular scenario was resolved in a later release of psalm/plugin-phpunit.

tl:dr; update the baseline when you're able to upgrade psalm/plugin-phpunit to see if the issue goes away.

src/HtmlProcessor/HtmlPruner.php
```diff
@@ -90,7 +90,7 @@ private function removeClassesFromElements(\DOMNodeList $elements, array $classe
     {
         $classesToKeepIntersector = new ArrayIntersector($classesToKeep);
 
-        /** @var \DOMNode $element */
+        /** @var \DOMElement $element */
```
JakeQZ: Nice.  I hate seeing elements described as mere nodes in this code base 👍

tests/Support/Traits/AssertCss.php
```diff
@@ -88,12 +88,12 @@ private static function assertNotContainsCss(string $needle, string $haystack)
      * @param string $needle
      * @param string $haystack
      */
-    private static function assertContainsCssCount(
+    private function assertContainsCssCount(
```
JakeQZ: Why no longer `static`?

tests/Support/Traits/AssertCss.php
```diff
@@ -88,12 +88,12 @@ private static function assertNotContainsCss(string $needle, string $haystack)
...
         int $expectedCount,
         string $needle,
         string $haystack
     ) {
-        self::assertSame(
+        $this->assertSame(
```
JakeQZ: `assertSame` is a `static` method and should be called staticly.  Is there an issue with Psalm that we need a workaround for here?

tests/Unit/CssInlinerTest.php
```diff
@@ -2768,7 +2768,7 @@ public function multiLineMediaQueryWithWindowsLineEndingsIsAppliedOnlyOnce()
 
         $subject->inlineCss($css);
 
-        self::assertContainsCssCount(1, $css, $subject->render());
+        $this->assertContainsCssCount(1, $css, $subject->render());
```
JakeQZ: Ditto. Etc.

tests/Unit/CssInlinerTest.php
```diff
@@ -3215,6 +3215,9 @@ public function copyUninlinableCssToStyleNodeHasNoSideEffects()
         $expectedHtml = $subject->render();
 
         $styleElement = $domDocument->getElementsByTagName('style')->item(0);
+
+        self::assertInstanceOf(\DOMElement::class, $styleElement);
+
```
JakeQZ: I don't think we need the empty lines around this addition, which is very closely related to both the preceding and following statements.

tests/Unit/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -368,7 +383,7 @@ public function removeRedundantClassesNotRemovesClassesToKeep(string $html, arra
         $result = $subject->renderBodyContent();
         foreach ($classesToKeep as $class) {
             $expectedInstanceCount = \substr_count($html, $class);
-            self::assertSubstringCount($expectedInstanceCount, $result, $class);
+            $this->assertSubstringCount($expectedInstanceCount, $result, $class);
```
JakeQZ: Again, if I'm not mistaken, this is a `static` method and should not be called via `$this`.  Same comment applies: is there an issue with Psalm?

tests/Unit/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -606,7 +629,7 @@ private static function assertContainsAll(array $needles, string $haystack)
      *
      * @throws AssertionFailedError
      */
-    private static function assertSubstringCount(int $expectedCount, string $haystack, string $needle)
+    private function assertSubstringCount(int $expectedCount, string $haystack, string $needle)
```
JakeQZ: Why make this non-static?

In the whole of PHPUnit, the assert methods are all `static`.  So when we introduce 'extension' methods, we retain consistency my making them `static` too.

src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -103,12 +103,40 @@ private function setHtml(string $html)
      * Provides access to the internal DOMDocument representation of the HTML in its current state.
      *
      * @return \DOMDocument
+     *
+     * @throws \UnexpectedValueException
      */
     public function getDomDocument(): \DOMDocument
     {
+        if (null === $this->domDocument) {
```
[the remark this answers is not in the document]
SignpostMarv: Force of habit of coding under yoda conditions.

src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -103,12 +103,40 @@ private function setHtml(string $html)
...
+                '::setDomDocument() has not yet been called on ' .
+                static::class
+            );
+        }
+
         return $this->domDocument;
     }
 
+    /**
+     * Provides access to the the internal DOMXPath instance corresponding to AbstractHtmlProcessor::$domDocument
```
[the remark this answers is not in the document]
SignpostMarv: The intent of this comment was to highlight the relationship between the `DOMXpath` instance and the `DOMDocument`, can reword at your preference though.

psalm.baseline.xml
SignpostMarv: > Maybe add an XML comment to that effect?

I'd thought of that, but I'm pretty sure the XML comment would be destroyed whenever psalm wrote to the baseline.

composer.json
```diff
@@ -74,6 +76,7 @@
         "ci:php:sniff": "phpcs config src tests",
         "ci:php:fixer": "php-cs-fixer --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff-format=udiff config/ src/ tests/",
         "ci:php:md": "phpmd src text config/phpmd.xml",
+        "ci:php:psalm": "psalm --show-info=false",
```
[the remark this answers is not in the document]
SignpostMarv: Had overlooked this, was running `composer run ci` locally.

There's also vimeo/psalm#1192  to take into account, the fix for which wasn't yet introduced in 3.2 (which you're stuck using because of various dependencies).

src/HtmlProcessor/AbstractHtmlProcessor.php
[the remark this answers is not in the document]
SignpostMarv: which actually occurs to me; I think there's a php-cs-fixer option to set yoda conditions explicitly to false?

tests/Unit/CssInlinerTest.php
SignpostMarv: ditto `NonStaticSelfCall` etc.

tests/Unit/HtmlProcessor/HtmlPrunerTest.php
SignpostMarv: yup, NonStaticSelfCall  https://github.com/MyIntervals/emogrifier/commit/015728397f9dd995afbbb20ee1bd6483967cfabc#diff-db03589574d7c5d074966c97658929d9

tests/Unit/HtmlProcessor/HtmlPrunerTest.php
SignpostMarv: They oddly didn't come up with `NonStaticSelfCall`.

oliverklee:
@SignpostMarv Thanks for this huge effort!

I'd like to keep "separate things in separate PRs", i.e., one separate PR for adding psalm to the build and adding the baseline, and then one (or more) separate PRs for fixing things found by psalm and adding psalm-specific annotations. Optimally, those separate PRs will start with one commit each.

oliverklee:
(Smaller PRs will also make the reviews easier and help get the parts merged a lot faster.)

SignpostMarv:
@oliverklee @JakeQZ When I split this into multiple PRs, should the `NonStaticSelfCall` be split onto it's own PR?



> Optimally, those separate PRs will start with one commit each.

At present this would result in 30~ PRs.

oliverklee:
> Optimally, those separate PRs will start with one commit each.

Sorry, my comment was a bit ambiguous. I meant "Define well-cut packages for each PR, and if necessary, squash several commits for the same package together before creating the PR for that package."

oliverklee:
> When I split this into multiple PRs, should the `NonStaticSelfCall` be split onto it's own PR?

In principle, yes. (Although I do not agree with that particular change. I'll add a comment to that commit in a second.)

psalm.baseline.xml
JakeQZ: Oh, so does Psalm potentially also write to this file (which looks like a config file to me)?  If so, we could nonetheless re-insert the comment when it does (or simply reject that specific change when doing a diff).

composer.json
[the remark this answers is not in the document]
JakeQZ: OK.  Let's do this as a separate PR.  I've added #790 so that it is captured and this conversation can be resolved.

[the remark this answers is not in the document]
JakeQZ: Oh, I see you've pushed a commit to this PR adding the step to Travis, but it's currently failing due to some remaining issues.  I suggest also addressing the remaining issues as that (or another) separate PR - this one is beginning to become hard to follow and keep track of via the GitHub UI.

src/HtmlProcessor/AbstractHtmlProcessor.php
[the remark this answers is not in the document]
JakeQZ: > I think there's a php-cs-fixer option to set yoda conditions explicitly to false?

I've added #791 to capture this :)

src/HtmlProcessor/AbstractHtmlProcessor.php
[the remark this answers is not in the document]
JakeQZ: Yes, it's true that the `DOMXPath` is tied to the `DOMDocument`, however the specific class property that the method accesses is `xPath` rather than `domDocument`.

tests/Support/Traits/AssertCss.php
JakeQZ: Looks like an issue with Psalm then.  `assertSame` is definitely a `static` method in PHPUnit.

I've added #792 to capture this.  For now, can we revert changes relating to calls via `self::` vs `$this` (and also any changes making methods `static` vs non-`static`), so that we can get this PR merged, and then move on to address the remaining issues separately?

tests/Unit/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: > They oddly didn't come up with `NonStaticSelfCall`.

See #792.

tests/Unit/HtmlProcessor/HtmlPrunerTest.php
JakeQZ: This may be because of a Psalm bug (see #792), but may perhaps be because you've also changed the `assertSubstringCount` method signature to be non-`static` (https://github.com/SignpostMarv/emogrifier/blob/psalm/tests/Unit/HtmlProcessor/HtmlPrunerTest.php#L632)?

JakeQZ:
> @oliverklee @JakeQZ When I split this into multiple PRs, should the `NonStaticSelfCall` be split onto it's own PR?
> 
> > Optimally, those separate PRs will start with one commit each.
> 
> At present this would result in 30~ PRs.

Though a moot point (because I'm pretty sure that code changes are not required for `NonStaticSelfCall` - see #792), if there were a need to change the `static`-ness of several methods, that would be fine in a single PR, because it would be addressing one specific issue.

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Support `ci:php:lint` on Windows
   ([#740](https://github.com/MyIntervals/emogrifier/issues/740),
   [#780](https://github.com/MyIntervals/emogrifier/pull/780))
+- Check the code with psalm
+  ([#537](https://github.com/MyIntervals/emogrifier/issues/537),
+  [#778](https://github.com/MyIntervals/emogrifier/pull/778))
```
JakeQZ: We currently have a policy of adding new CHANGELOG entries to the top of the relevant section.  (I was caught out by this in my first contribution to the project.)

(Whether that is the best policy is another issue.  The way we're doing things currently, it does mean that the list of changes/fixes in the release note ends up being in reverse-chronological order, which does seem natural to me.)

JakeQZ:
I've added #790, #791 and #792 to capture some additional issues.  Hopefully this will help us more swiftly get this PR merged.  We can then move on to look at addressing those issues.

psalm.baseline.xml
oliverklee: It does (if the corresponding command line argument is provided). I highly recommend not editing this file manually as it is supposed to be automatically created.

src/CssInliner.php
```diff
@@ -1069,7 +1099,7 @@ private function copyUninlinableCssToStyleNode(string $cssImportRules)
         $css = $cssImportRules;
 
         // avoid including unneeded class dependency if there are no rules
-        if ($this->matchingUninlinableCssRules !== []) {
+        if ($this->matchingUninlinableCssRules !== [] && $this->matchingUninlinableCssRules !== null) {
```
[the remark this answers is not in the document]
SignpostMarv: Given that null is the default, I'd agree- however, the vertical diff is a little easier to read in this instance.

src/HtmlProcessor/AbstractHtmlProcessor.php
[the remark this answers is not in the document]
SignpostMarv: leave this as pending for resolution of #789 ?

src/CssInliner.php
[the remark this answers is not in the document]
JakeQZ: Though once this PR is merged, the ease of reading the diff will become largely irrelevant (and GitHub, at least, does highlight the actual line additions with a darker green background)...

src/HtmlProcessor/AbstractHtmlProcessor.php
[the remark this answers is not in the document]
JakeQZ: Looking again, I think this actually is fine as it is.  I had initially thought it was the result of an oversight copy-and-pasting from `getDomDocument()`.

psalm.baseline.xml
JakeQZ: > I highly recommend not editing this file manually as it is supposed to be automatically created.

If it's automatically created, then we should not be adding this file to the repository at all.  Or am I missing something?

SignpostMarv: > > I highly recommend not editing this file manually as it is supposed to be automatically created.
> 
> If it's automatically created, then we should not be adding this file to the repository at all. Or am I missing something?

running `psalm --update-baseline` or `psalm --set-baseline=./psalm.baseline.xml` will cause the file to be written to.

the bugfix commits on this PR were created by working from the bottom up, removing an issue type or code entry, then running psalm to find the code that caused that chunk of the baseline to be added.

SignpostMarv: > If so, we could nonetheless re-insert the comment when it does (or simply reject that specific change when doing a diff).

this method is also consistent with the manual changes performed first in 130782e (has to be re-done every time with psalm 3.2 to avoid travis failing) and in 2b6ec42 (doesn't affect build status, but is still annoying).

SignpostMarv:
rebased minus those commits referenced in #792

psalm.baseline.xml
SignpostMarv: resolved in af615a9, flagged to be monitored by #793

src/CssInliner.php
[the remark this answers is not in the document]
SignpostMarv: resolved in 2f91880

src/HtmlProcessor/AbstractHtmlProcessor.php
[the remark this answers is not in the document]
SignpostMarv: resolved via #794

tests/Unit/CssInlinerTest.php
SignpostMarv: reverted via today's rebase.

tests/Unit/HtmlProcessor/HtmlPrunerTest.php
SignpostMarv: reverted in rebase referenced https://github.com/MyIntervals/emogrifier/pull/778#issuecomment-538681924

tests/Unit/HtmlProcessor/HtmlPrunerTest.php
SignpostMarv: reverted in rebase referenced https://github.com/MyIntervals/emogrifier/pull/778#issuecomment-538681924

tests/Unit/CssInlinerTest.php
SignpostMarv: resolved in 164d2f6

JakeQZ:
I'd rather https://github.com/SignpostMarv/emogrifier/commit/dadd1da42d4ee6857031975003325cb7a28521f2 was in a separate PR, so that we can compartmentalize the changes, if that's OK?

SignpostMarv:
> I'd rather [SignpostMarv@dadd1da](https://github.com/SignpostMarv/emogrifier/commit/dadd1da42d4ee6857031975003325cb7a28521f2) was in a separate PR, so that we can compartmentalize the changes, if that's OK?

it is ?

JakeQZ:
> it is ?

Appears on this branch.  Don't worry I'll look later.

SignpostMarv:
> Appears on this branch.

commit message references the discussions in this thread.

JakeQZ:
Sorry, hadn't spotted the side-PR...

SignpostMarv:
> Sorry, hadn't spotted the side-PR...

confusing ui is confusing.

JakeQZ:
#794 is now merged to master so you can rebase.

(I tried to correct capitalization on the commit message too, but GitHub seemed to ignore it [case-insensitive comparison for changes somewhere?], I hope “yoda” doesn’t mind.)

src/HtmlProcessor/AbstractHtmlProcessor.php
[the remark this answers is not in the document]
SignpostMarv: resolved in df750c6

SignpostMarv:
PR has been rebased, psalm issues have been resolved (unless otherwise noted in the xml comments) up to tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php as of f3556f8

will pick up the remainder tomorrow.

JakeQZ:
Super.  At quick glance, the only remaining issue I see the CHANGLELOG entry order.  Hopefully @oliverklee will be avaiable to be your captain tomorrow – this is my first encounter with Psalm 🌙

tests/Unit/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -611,7 +635,7 @@ private static function assertSubstringCount(int $expectedCount, string $haystac
         self::assertSame(
             $expectedCount,
             \substr_count($haystack, $needle),
-            'asserting \'' . $haystack . '\' contains ' . $expectedCount . ' instance(s) of "' . $needle . '"'
+            'asserting \'' . $haystack . '\' contains ' . (string)$expectedCount . ' instance(s) of "' . $needle . '"'
```
JakeQZ: Good catch, this is a required cast.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -399,12 +417,12 @@ static function (array $dataset) {
      * Concatenates pairs of datasets (in a similar way to SQL `JOIN`) such that each new dataset consists of a 'row'
      * from a left-hand-side dataset joined with a 'row' from a right-hand-side dataset.
      *
-     * @param string[][] $leftDatasets
-     * @param string[][] $rightDatasets
+     * @param array<string, array<int, string>> $leftDatasets
+     * @param array<string, array<int, string>> $rightDatasets
```
JakeQZ: I think this syntax is clearer for object-like arrays, if supported by PSR-5 (which as far as I know is incomplete anyway - see https://github.com/phpDocumentor/phpDocumentor2/issues/557#issuecomment-390924081).  We had some relevant discussion recently at https://github.com/MyIntervals/emogrifier/pull/708/commits/d7d7b16eca7c7301d4022377cb4463c6d9f44c42#r324370364

JakeQZ: Though for arrays with string keys the proposal at https://github.com/phpDocumentor/phpDocumentor2/issues/557#issuecomment-390924081 looks better.

SignpostMarv: When I come across docblocks like these that aren't quite specific enough for psalm, there are two choices:
1. use an array shape
2. use a more strictly-typed array (as was done here)

Some php analysis tools don't support array shapes and will consider such returns/variables/parameters as untyped (mixed, mixed[] etc.). adding an array shape generally requires adding an extra line replicating the `@param` to some degree.

In this particular case, the psalm issue was related to the possibility of non-string keys being found in the arrays, as `string[][]` is equivalent to `array<array-key, array<array-key, string>>`, where `array-key` can be either `string` or `int`, whereas the implementation uses only string keys on the outter array and int keys on the inner array (hence `array<string, array<int, string>>`).

For typed arrays vs. shaped arrays, see 23689aa vs. f3556f8

As for what to do with associative arrays, see the commits listed as referenced under #742.

SignpostMarv:
have identified a set of non-array issues in ~63964ee~ 3076187 that seem similar enough in intent to #742 to be retained for review by that issue rather than #537

oliverklee:
Everybody, please let's split this into smaller PRs and merge those. This PR is far too big and changes things that are too different for me to be comfortable to merge it.

In addition, for new topics and issues, please let's add dedicated issues for those instead of having them in existing tickets/PRs that already are closed or are about something only related.

SignpostMarv:
> Everybody, please let's split this into smaller PRs and merge those.

PR by file is probably a good approach, given that I've been going up from the bottom of the baseline.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
oliverklee: Generally, I'd like to keep the `string[]` notation over the `array<string>`/`array<int, string>` notation as long as possible because PhpStorm currently only supports the first notation. (There already is a ticket at JetBrains for this feature request.)

CHANGELOG.md
oliverklee: The thinking behind the reverse-chronological order is that then the order of the releases (newest on top) is the same as the order of the changes within the release.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
SignpostMarv: > Generally, I'd like to keep the `string[]` notation over the `array<string>`/`array<int, string>` notation as long as possible because PhpStorm currently only supports the first notation. (There already is a ticket at JetBrains for this feature request.)

noted, will decorate with psalm-specific docblock tags in future.

SignpostMarv:
> This PR is far too big and changes things that are too different for me to be comfortable to merge it.

closing this PR and replicating the branch so the current state of this branch can be retained for reference.

src/CssInliner.php
```diff
@@ -1093,15 +1123,22 @@ private function copyUninlinableCssToStyleNode(string $cssImportRules)
...
-        $styleAttribute = $this->domDocument->createAttribute('type');
+        $styleElement = $this->getDomDocument()->createElement('style', $css);
+        $styleAttribute = $this->getDomDocument()->createAttribute('type');
         $styleAttribute->value = 'text/css';
         $styleElement->appendChild($styleAttribute);
 
         $headElement = $this->getHeadElement();
+
+        if ($headElement === null) {
+            throw new \UnexpectedValueException('Could not find <head>');
```
JakeQZ: For other exceptions we have used a unique UNIX timestamp (the time at which the programmer put in the code, unique for each instance in the source code) as an exception code.  This could help precisely identify the source if an issue is reported.

tests/Support/Traits/AssertCss.php
```diff
@@ -29,6 +29,9 @@ private static function getCssNeedleRegExp(string $needle): string
         $needleMatcher = \preg_replace_callback(
             '/\\s*+([{},])\\s*+|(^\\s++)|(>)\\s*+|(?:(?!\\s*+[{},]|^\\s)[^>])++/',
             static function (array $matches): string {
+                /** @psalm-var array{0:string, 1:string, 2:string, 3:string} */
+                $matches = $matches;
+
```
JakeQZ: Can this instead be `@psalm-param` for the function/closure?

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
JakeQZ: Can we not have both `@param` and `@psalm-param` here?

tests/Support/Traits/AssertCss.php
SignpostMarv: see commit message for f6ea5a1

src/CssInliner.php
oliverklee: Oh, looks like I forgot to add the exception code. I'll need to do this (or someone else).

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
SignpostMarv: that's the intention yes; I was going based on my experience of what types of param hints cause issues with other tools, but was not aware of phpstorm's particular issues.

src/CssInliner.php
JakeQZ: > Oh, looks like I forgot to add the exception code. I'll need to do this (or someone else).

I've checked and all current (3) instances within `src` in `master` have a suitable code.  The only exceptions (literally, not `Exception`s) are cases where an `exception` is rethrown.

I was just commenting on the new addition(s) in this PR branch.

SignpostMarv: noted, will get around to it after #795 is merged.

JakeQZ:
> 
> 
> > Everybody, please let's split this into smaller PRs and merge those.
> 
> PR by file is probably a good approach, given that I've been going up from the bottom of the baseline.

Can I suggest by specific Psalm check rather than by file?

i.e. Enable one-by-one the checks in `<issueHandlers>` in `psalm.xml` and address the code changes required for that specific issue.  And each `issueHandler` in a separate PR.  With a single CHANGELOG entry in the `Fixed` section "Issues detected by Psalm <name-of-issue-handler>".

If there are some `<issueHandlers>` that can be added without making any code changes, then we could add those as a single PR, and probably do not need a CHANGELOG entry since we already have "Check the code with Psalm" and there was nothing else to fix.

@SignpostMarv, @oliverklee, does that make sense?

SignpostMarv:
> i.e. Enable one-by-one the checks in `<issueHandlers>` in `psalm.xml` and address the code changes required for that specific issue.

issues error by default.

SignpostMarv:
> Can I suggest by specific Psalm check rather than by file?

PR by file would mean continuous blocks would be removed from ./psalm.baseline.xml, generally having a single block to merge- making rebasing and merging more straightforward.

PR by type can introduce issues of other types, possibly introducing issues a PR was already made for; there's a couple of those in this PR.

JakeQZ:
> issues error by default.

So we could enable one issue, and fix the errors for that specific issue, and have that as a single PR?  Rinse and repeat?

Those that cause errors we cannot address right now (e.g. `this` vs `self` due to version constraints) will have to wait until PHP 7.0 support is dropped (or we allow the Psalm step to be skipped for PHP 7.0 in Travis, but still run for PHP 7.1+).

SignpostMarv:
> So we could enable one issue, and fix the errors for that specific issue, and have that as a single PR? Rinse and repeat?

prepping an explanation on #795 in response to query there.

JakeQZ:
> prepping an explanation on #795 in response to query there.

OK, thanks.  Just been reading https://erik.booij.me/a-year-with-psalm/

SignpostMarv:
> prepping an explanation on #795 in response to query there.

following on from [that explanation](https://github.com/MyIntervals/emogrifier/pull/795#issuecomment-538756309), what one *could* do is:
1. set an issue type to "info",
2. run psalm with `--set-baseline`
3. remove the issue type entry added in step 1
4. run `psalm --show-info=false`
5. fix issues.

what I've been doing is:
1. remove an issue type from a particular file in the baseline, or remove a single instance if there are many
2. run `psalm --show-info=false`
3. fix issues

JakeQZ:
> following on from [that explanation](https://github.com/MyIntervals/emogrifier/pull/795#issuecomment-538756309)

Thanks for that.  I'm beginning to get a grasp on how it works 🙂

> what one _could_ do is

I'm just wanting to be sure we're taking the best approach to make all our lives easier, both in terms of being able to keep track of what changes are being made to the code base and why, and to minimize the effort actually required to get to a good baseline.

src/CssInliner.php
oliverklee: I've added #797 for this.

JakeQZ:
> ...

Really, any way that breaks down the changes into small mangeable chunks would be good :)

@SignpostMarv, we very much appreciate the effort you've been putting in here, it's been amazing.

PR #783, opened 2019-10-02
[TASK] Upgrade to PHPUnit 6.5

oliverklee:


oliverklee:
> This is a start, but we should also use the PHPUnit namspaced classes now...

Don't we already? Or am I missing something?

JakeQZ:
> > This is a start, but we should also use the PHPUnit namspaced classes now...
> 
> Don't we already? Or am I missing something?

We used them partially with 5.7 but they were dummy aliases/wrappers.  See e.g. https://github.com/MyIntervals/emogrifier/pull/778/files#diff-10 for furhter changes to complate the migration.

oliverklee:
I've just created #786 to fix this.

PR #784, opened 2019-10-02
[CLEANUP] Use self as the return type for fromHtml

oliverklee:


src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -60,7 +60,7 @@ private function __construct()
      *
      * @throws \InvalidArgumentException if $unprocessedHtml is anything other than a non-empty string
      */
-    public static function fromHtml(string $unprocessedHtml): AbstractHtmlProcessor
+    public static function fromHtml(string $unprocessedHtml): self
```
JakeQZ: Um, also for `::fromDomDocument` please :)

oliverklee: Oh. Right. (Note to myself: Have coffee and breakfast first. Code after that, not before.)

oliverklee: Note: I'd like to keep the `@return static` in the PHPDoc to make it explicit that we return an instance of the corresponding called class.

oliverklee:
Updated and repushed.

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
JakeQZ: > keep the `@return static` in the PHPDoc to make it explicit

I strongly agree.  There is still not much correlation between PHPDoc and type hints and [probably never will be](https://wiki.php.net/rfc/arrayof).

JakeQZ: > Have coffee and breakfast first

Or as the newspaper-reading types do: look at the main article AND the side article :)

src/Emogrifier/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -45,7 +45,7 @@ abstract class AbstractHtmlProcessor
     /**
      * The constructor.
      *
-     * Please use ::fromHtml instead.
+     * Please use `::fromHtml` or `::fromDomDocument` instead.
```
JakeQZ: Good correction.  Coffee clearly working :)

PR #786, opened 2019-10-02
[CLEANUP] Stop using replaced PHPUnit classes

oliverklee:
We now use the namespaced classes in all places.


JakeQZ:
Also, I wonder if we should use `ExpectationFailedException` rather than `AssertionFailedError`...?

oliverklee:
> There are a bunch of instances in `AssertCssTest.php` that should also be accordingly changed...

I already changed those in 249d81c31dee81c987b0be5151bcadbffab36c21.

oliverklee:
> Also, I wonder if we should use `ExpectationFailedException` rather than `AssertionFailedError`...?

Yes, that would be good. Let's do that in a follow-up PR.

oliverklee:
Rebased to resolve the conflicts and repushed.

JakeQZ:
> > There are a bunch of instances in `AssertCssTest.php` that should also be accordingly changed...
> 
> I already changed those in [249d81c](https://github.com/MyIntervals/emogrifier/commit/249d81c31dee81c987b0be5151bcadbffab36c21).

Yes, and I reviewed.  Some kind of dementia must be taking effect...

oliverklee:
I have created issue #787 for using `ExpectationFailedException`.

Issue #791, opened 2019-10-04
Add PHPCS check to disallow Yoda conditions


JakeQZ:
See https://github.com/MyIntervals/emogrifier/pull/778#discussion_r330913843 and https://en.wikipedia.org/wiki/Yoda_conditions

JakeQZ:
Resolved by #794.

PR #794, opened 2019-10-05
[TASK] disable php-cs-fixer yoda conditions

SignpostMarv:
regarding: #791, https://github.com/MyIntervals/emogrifier/pull/778#discussion_r330913843

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+- disable php-cs-fixer yoda conditions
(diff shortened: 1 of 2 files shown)


SignpostMarv:
> Could we have a CHANGELOG entry ('Added' section I think)?
> 
> Thanks.

d'oh

SignpostMarv:
*used the `--force`*

JakeQZ:
> _used the `--force`_

That's fine in this case.  I think I should be able to uppercase the first character of the CHANGELOG title while you focus on more important work 😎

JakeQZ:
> _used the `--force`_

And now just spotted the pun :)

Issue #793, opened 2019-10-04
track issues relating to raising the minimum php version to 7.1 or greater.


SignpostMarv:
- [x] raise minimum supported php version to 7.1
- [ ] update vimeo/psalm to latest version that supports 7.1
- - [ ] update psalm/plugin-phpunit to latest version that supports 7.1
- - [ ] remove `requireVoidReturnType="false"` from `psalm.xml`, re: 9d36ae4
- - [ ] review f6ea5a1, apply intended `@psalm-param` typehint & see if `MixedArgumentTypeCoercion` is resolved (or file issue with psalm)
- - [ ] review if updates resolve #792 without requiring code changes.
- - [ ] review if updates resolve the closure issue identified in #806 without requiring code changes
- [ ] update phpunit/phpunit to latest version that supports 7.179

oliverklee:
We now require PHP >= 7.2. So this is all done now. Closing.

Issue #792, opened 2019-10-04
Investigate NonStaticSelfCall Psalm issue


JakeQZ:
See https://github.com/MyIntervals/emogrifier/pull/778#discussion_r330914954.  It seems perhaps that Psalm is sporadically reporting false negatives with the `NonStaticSelfCall` check enabled.  We may need to disable this check in the Psalm config, before Psalm can be included in the Travis CI checks (#790).  It may perhaps be the case that this is fixed in a version we can't yet use due to dependency constrainsts (e.g. supporting PHP 7.0).

oliverklee:
@JakeQZ Good research! It looks like there's already a PR for this: https://github.com/psalm/psalm-plugin-phpunit/pull/50

JakeQZ:
There's a new release ([0.9.2](https://github.com/psalm/psalm-plugin-phpunit/releases/tag/0.9.2)) with the fix, but this would require a later version of Psalm (^3.6.2), which in turn would require a later version of `sebastian/diff` which is not compatible with PHPUnit 6.x - the latest version that still supports PHP 7.0.

So I would suggest this needs to wait until Emogrifier 5.0 (dropping PHP 7.0 support) to be resolved - unless Phive may be able to help with the dependency problem (we don't need Psalm to be able to run PHPUnit, and we only need to run Psalm against the current-minus-one PHP version).  Hence I've set 5.0 Milestone.

Additionally, later versions of Psalm introduce more checks, with additional code issues detected which would need investigating (though the issues reported could be added to the baseline temporarily), so updating Psalm is not a trivial change.

oliverklee:
Phive might indeed help.

oliverklee:
These are all resolved now. Closing.

PR #795, opened 2019-10-06
[TASK] Retain psalm issues for future review

SignpostMarv:
cherry-picking the commits from #778 that pertain to issues that either need to be retained until resolved by discussion on other issues, or are being retained as a result of said discussion.

### DIFF
--- a/psalm.baseline.xml
+++ b/psalm.baseline.xml
@@ -154,13 +154,14 @@
       <code>$lastRuleBlock-&gt;declarationsBlock</code>
     </MixedPropertyFetch>
   </file>
+  <!-- NonStaticSelfCall is retained, to be reviewed with #792 -->
   <file src="tests/Support/Traits/AssertCss.php">
     <MixedArgument occurrences="3">
       <code>$matches[1]</code>
       <code>$matches[3]</code>
       <code>$matches[0]</code>
     </MixedArgument>
-    <NonStaticSelfCall occurrences="1"/> <!-- retained, to be reviewed with #792 -->
+    <NonStaticSelfCall occurrences="1"/>
   </file>
   <file src="tests/Unit/CssInlinerTest.php">
     <InvalidOperand occurrences="3">
@@ -216,6 +217,7 @@
       <code>assertInstanceOf</code> <!-- retained as per discussion in #537 -->
     </RedundantConditionGivenDocblockType>
   </file>
+  <!-- NonStaticSelfCall is retained, to be reviewed with #792 -->
   <file src="tests/Unit/HtmlProcessor/HtmlPrunerTest.php">
     <InvalidOperand occurrences="1">
       <code>$expectedCount</code>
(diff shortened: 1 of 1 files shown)


JakeQZ:
I've added labels and edited the title to indicate this contains commits for reference, and is not something to be reviewed and merged right now - as I understand that to be the case.

SignpostMarv:
this particular PR isn't a WIP, it's to flag issues that shouldn't be removed in PRs without first checking other issues
* `NonStaticSelfCall` - existing issues will likely be permanently retained, pending outcome of #792 
* `RedundantConditionGivenDocblockType` - some issues are being permanently retained, as a result of discussion in #537
* flagged issues for `src/Utilities/CssConcatenator.php` are probably under the purview of #742, so shouldn't be resolved unless specifically as part of that issue.

JakeQZ:
> this particular PR isn't a WIP, it's to flag issues that shouldn't be removed in PRs without first checking other issues

OK.  Maybe I could have chosen a better title prefix.  I just wanted to make it clear that

> SignpostMarv wants to merge 4 commits into MyIntervals:master from SignpostMarv:retain-psalm-issues-for-future-review

is misleading.

SignpostMarv:
> > SignpostMarv wants to merge 4 commits into MyIntervals:master from SignpostMarv:retain-psalm-issues-for-future-review
> 
> is misleading.

ah, clarification; this is in relation to #793, as running `psalm --set-baseline=./psalm.baseline.xml` after upgrading psalm may remove baseline entries due to bugfixes in later releases of psalm (i.e. you'd be going from 3.2.12 to ^3.5), or because issues were reclassified by psalm so the comments would need to be added elsewhere, as opposed to tapping `[s]` when `git commit -p`

SignpostMarv:
"future review" in this instance meaning "when upgrading psalm / project to php 7.1"

JakeQZ:
OK.  I'm still confused as to why `psalm.baseline.xml` needs to be in the repo at all - it looks more like a logfile to me?  Or does it also act as a cache to speed up future runs?  (I note I also have a file called `.php_cs.cache` which is not in the repo.)  But like I said, I'm new to Psalm 😕

SignpostMarv:
`psalm.xml` is the config, where you can set types of issues to be set to globally `info` or `suppress`. one can also set individual files or directories to be ignored.

`psalm.baseline.xml` keeps a record of *specific instances* of issues, effectively setting those instances to `info`.

travis runs `psalm --show-info=false`, passing the psalm check if no new issues are introduced.

`psalm --update-baseline` effectively removes issues that're resolved **without recording new issues introduced**

`psalm --set-baseline=[path-to-file]` records all current issues, removing issues that're resolved by virtue of them not being included in the output.

any issue not currently referenced in the baseline will necessarily be "new" and will cause travis to fail, issues which might not be caught if the baseline were not employed and `<issueHandlers>` or `<ignoreFiles`> were used instead.

tl:dr; baseline is more of a record than a log and helps identify newly introduced issues.

SignpostMarv:
I'll not be able to work on this while I'm sat at the day job, but could this be merged for this evening ?

SignpostMarv:
thanks :D

oliverklee:
You're welcome. Thanks for working on the psalm issues! :coffee:

JakeQZ:
> 
> 
> I've added labels and edited the title to indicate this contains commits for reference, and is not something to be reviewed and merged right now - as I understand that to be the case.

Apologies.  It seems I completely misunderstood this PR.

I ***will*** get up to speed with Psalm, I promise...

🥧(humble)

PR #799, opened 2019-10-07
[BUGFIX] partially satisfy psalm for tests/Unit/HtmlProcessor/HtmlPrunerTest.php

SignpostMarv:


tests/Unit/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -611,7 +635,7 @@ private static function assertSubstringCount(int $expectedCount, string $haystac
         self::assertSame(
             $expectedCount,
             \substr_count($haystack, $needle),
-            'asserting \'' . $haystack . '\' contains ' . $expectedCount . ' instance(s) of "' . $needle . '"'
+            'asserting \'' . $haystack . '\' contains ' . (string)$expectedCount . ' instance(s) of "' . $needle . '"'
```
JakeQZ: This is a positive change, as I've already said :)

SignpostMarv:
should note I'm cherry-picking here and leaving it up to travis to catch any `composer run ci` errors.

JakeQZ:
> it up to travis to catch any `composer run ci` errors.

That's what it's there for.  It won't (or shouldn't) allow us to commit a broken build :)

oliverklee:
> That's what it's there for. It won't (or shouldn't) allow us to commit a broken build :)

There is a possible edge case: If two commits from separate PRs clash with each other without causing a Git conflict and the second PR is merged without rebasing and running Travis CI again, the master build will fail. I've had this every now an then in some projects.

JakeQZ:
> I mean as opposed to redoing the psalm commits from scratch :P

How we're doing things right now seems to be working well.  The PRs are coming through **_thin_** and fast - at lightning speed! :))

PR #801, opened 2019-10-07
[BUGFIX] partially satisfy psalm for tests/Unit/CssInlinerTest.php

SignpostMarv:


tests/Unit/CssInlinerTest.php
```diff
@@ -3215,6 +3216,7 @@ public function copyUninlinableCssToStyleNodeHasNoSideEffects()
         $expectedHtml = $subject->render();
 
         $styleElement = $domDocument->getElementsByTagName('style')->item(0);
+        self::assertInstanceOf(\DOMElement::class, $styleElement);
```
JakeQZ: It looks a bit messy having this here, but this is a messy test in the first place, so I'm fine with this.

SignpostMarv:
> Again, we need separate `@param` and `@psalm-param` entries in the DocBlock...

resolved in 062f73b

tests/Unit/CssInlinerTest.php
SignpostMarv: it's to cause the test to fail in the event that `DOMNodeList::item()` returns null.

oliverklee:
> BTW, we probably don't need the full path of the file in the commit title, as there's almost always only going to be one file with that name. A succincter title helps a reader browsing through a list of commits.

For example: [CLEANUP] Fix psalm warnings in CssInlinerTest

JakeQZ:
> 
> 
> > BTW, we probably don't need the full path of the file in the commit title, as there's almost always only going to be one file with that name. A succincter title helps a reader browsing through a list of commits.
> 
> For example: [CLEANUP] Fix psalm warnings in CssInlinerTest

I amended the title a bit, though not quite like that.  Hope what I did was OK...

JakeQZ:
> I amended the title a bit, though not quite like that. Hope what I did was OK...

Is there any way to edit commit messages/titles once the've been merged to `master`?  There should be but I couldn't find a way...

oliverklee:
> Is there any way to edit commit messages/titles once they've been merged to master?

No (at least not with a protected master branch).

JakeQZ:
> 
> 
> > Is there any way to edit commit messages/titles once they've been merged to master?
> 
> No (at least not with a protected master branch).

That's what I feared.  And why on a couple of these PRs I suggested you do the final squash, merge and commit.  Because I've already screwed up the commit title/message on a couple of merges.  But given the need to move forward, and the lightning speed at which @SignpostMarv seems to be working, I felt I had to commit changes I was perfectly happy with and which I could not think of any reason you (@oliverklee) would not be.

GitHub/`git` is good at making sure everything is peer-reviewed, CI-tested, etc.  But the one thing it is strikingly lacking is having the final commit title/message peer-reviewed.  Or any ability to correct it later (incorrect code changes can always be sorted with a further commit).  We should perhaps consider raising an issue with GitHub itself?  (The sqaush-and-merge feature is good and useful, but it does apparently introduce such problems.)

PR #802, opened 2019-10-07
[TASK] track issues relating to php 7.1

SignpostMarv:
~This commit resolves a set of MixedArgument issues.
This commit *should* be type-hinting the closure parameter via
`@psalm-param array{0:string, 1:string, 2:string, 3:string} $matches`,
but as of vimeo/psalm 3.2.12, this raises a MixedArgumentTypeCoercion
issue.~

~regarding: #537, #793~

regarding: #806, https://github.com/MyIntervals/emogrifier/pull/802#pullrequestreview-298296128

### DIFF
--- a/psalm.baseline.xml
+++ b/psalm.baseline.xml
@@ -157,9 +157,9 @@
   <!-- NonStaticSelfCall is retained, to be reviewed with #792 -->
   <file src="tests/Support/Traits/AssertCss.php">
     <MixedArgument occurrences="3">
-      <code>$matches[1]</code>
-      <code>$matches[3]</code>
-      <code>$matches[0]</code>
+      <code>$matches[1]</code> <!-- retained, to be reviewed with #793  -->
+      <code>$matches[3]</code> <!-- retained, to be reviewed with #793  -->
+      <code>$matches[0]</code> <!-- retained, to be reviewed with #793  -->
     </MixedArgument>
     <NonStaticSelfCall occurrences="1"/>
   </file>


tests/Support/Traits/AssertCss.php
```diff
@@ -29,6 +29,9 @@ private static function getCssNeedleRegExp(string $needle): string
         $needleMatcher = \preg_replace_callback(
             '/\\s*+([{},])\\s*+|(^\\s++)|(>)\\s*+|(?:(?!\\s*+[{},]|^\\s)[^>])++/',
             static function (array $matches): string {
+                /** @psalm-var array{0:string, 1:string, 2:string, 3:string} */
+                $matches = $matches;
```
JakeQZ: Sorry, I really don't like the idea of adding redundant code.

Perhaps a solution to this seeming impasse would be to move the closure to a well-named `private static` method?

We could do this as a separate PR, then come back to this.  I don't mind having a go at that, unless @SignpostMarv  you'd like to yourself?

SignpostMarv: the closure isn't the issue, psalm is the issue- specficially, https://github.com/vimeo/psalm/commit/aa3df3f7a511ae4e12dedaa8f21552c47814fc5b#diff-4355ecf0bcc89fe8c145b64be0e02ddf

Shall we add this issue to the list of issues to be reviewed under #793 ?

JakeQZ: > Shall we add this issue to the list of issues to be reviewed under #793 ?

Seems like perhaps the best way forward to me.  Thanks :)

SignpostMarv: resolved by 87d4b69c12b9fb28a776c888d4dcbea4b623b524

PR #803, opened 2019-10-07
[CLEANUP] Fix psalm issues in HtmlPruner

SignpostMarv:
This commit resolves the UndefinedMethod issues identified in
src/HtmlProcessor/HtmlPruner.php by altering the type hints to match
the child type of DOMNode more closely matching what is expected.

regarding: #537

### DIFF
--- a/psalm.baseline.xml
+++ b/psalm.baseline.xml
@@ -111,8 +111,7 @@
     </MixedArgumentTypeCoercion>
   </file>
   <file src="src/HtmlProcessor/HtmlPruner.php">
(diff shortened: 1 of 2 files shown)


oliverklee:
> However, I'm aware that the type hints were originally added by PhpStorm. @oliverklee, could there be an issue there?

No, that should be fine. (The type hints were added _for_ PhpStorm, not by it, by the way. :-) )

JakeQZ:
> The type hints were added _for_ PhpStorm, not by it, by the way. :-)

Useful to know :)

Issue #733, opened 2019-09-28
Add a class to be used instead of StdClass in CssConcatenator


oliverklee:

JakeQZ:
Once the `Emogrifier` class is removed, we can do this, as we require installation with Composer and an autoloader (or something to that effect) to use the new classes and have additional dependencies automatically `include`d.

Issue #373, opened 2017-03-02
Update the CONTRIBUTING file


oliverklee:
It looks like this still leaves too many questions open.

JakeQZ:
From experience, if a `rebase` is required, the procedure can go awry if not performed carefully.  The following is a tried-and-tested procedure to rebase a ‘feature’ branch to the latest `master`:

1. Make sure that your local copy of the repository has the most up-to-date
  revisions of `master` (this is important, otherwise you may end up rebasing to
  an older base point):
  ```sh
  git checkout master
  git pull
  ```
2. Switch to the (feature) branch to be rebased and make sure your copy is up to
  date:
  ```sh
  git checkout feature/something-cool
  git pull
  ```
3. Consider taking a copy of the folder tree at this stage, this may help when
  resolving conflicts in the next step.
1. Begin the rebasing process
  ```sh
  git rebase master
  ```
5. Resolve the conflicts in the reported files.  (This will typically require
  reversing the order of the new entries in `CHANGELOG.md`.)  You may use a
  folder ‘diff’ against the copy taken at step 3 to assist, but bear in mind
  that at this stage `git` is partway through rebasing, so some files will have
  been merged and include the latest changes from `master`, whilst others might
  not – in any case, you should ignore changes to files not reported as having
  conflicts.
1. Mark the conflicting files as resolved and continue the rebase
  ```sh
  git add *.md *.php
  git rebase --continue
  ```
  (You can alternartively specify individual files with a full relative path
  rather than using wildcards.)
7. Force-push the rebased (feature) branch to the remote repository
  ```sh
  git push --force
  ```
  The `--force` option is important.  Without it, you’ll get an error with a
  hint suggesting a `git pull` is required:
  ```
  hint: Updates were rejected because the tip of your current branch is behind
  hint: its remote counterpart. Integrate the remote changes (e.g.
  hint: 'git pull ...') before pushing again.
  hint: See the 'Note about fast-forwards' in 'git push --help' for details.
  ```
  ***DO NOT*** follow the hint and execute `git pull`.  This will result in the
  set of all commits on the feature branch being duplicated, and the ‘patching
  base’ not being moved at all.

JakeQZ:
x-ref: https://github.com/MyIntervals/emogrifier/pull/804#discussion_r332281248

JakeQZ:
Some notes about potential pitfalls, mostly for classification/recollection at this stage as we indulge in Psalm - off the top of my head:

[...]

PR #804, opened 2019-10-07
[FEATURE] Add helper classes to be used instead of stdClass in CssConcatenator

SignpostMarv:


psalm.baseline.xml
```diff
@@ -128,32 +128,6 @@
...
-    <MixedPropertyAssignment occurrences="1">
-      <code>$lastRuleBlock</code>
-    </MixedPropertyAssignment>
-    <MixedPropertyFetch occurrences="4">
-      <code>$lastRuleBlock-&gt;declarationsBlock</code>
-      <code>$lastRuleBlock-&gt;selectorsAsKeys</code>
-      <code>$lastRuleBlock-&gt;selectorsAsKeys</code>
-      <code>$lastRuleBlock-&gt;declarationsBlock</code>
-    </MixedPropertyFetch>
-  </file>
```
SignpostMarv: removed due to implementation of dedicated class, as a candidate for #742

src/Utilities/CssConcatenator.php
```diff
@@ -42,16 +42,9 @@
...
     /**
-     * Array of media rules in order.  Each element is an object with the following properties:
-     * - string `media` - The media query string, e.g. "@media screen and (max-width:639px)", or an empty string for
-     *   rules not within a media query block;
-     * - \stdClass[] `ruleBlocks` - Array of rule blocks in order, where each element is an object with the following
-     *   properties:
-     *   - mixed[] `selectorsAsKeys` - Array whose keys are selectors for the rule block (values are of no
-     *     significance);
-     *   - string `declarationsBlock` - The property declarations, e.g. "margin-top: 0.5em; padding: 0".
+     * Array of media rules in order.
```
SignpostMarv: Comments were moved to the appropriate dedicated class/property/parameter.

src/Utilities/CssConcatenator.php
```diff
@@ -42,16 +42,9 @@
...
      *
-     * @var \stdClass[]
+     * @var CssConcatenator\MediaRule[]
```
SignpostMarv: this could technically be `array<int, CssConcatenator\MediaRule>`, but psalm seems happy without it, so best not to confuse phpstorm :P

src/Utilities/CssConcatenator.php
```diff
@@ -98,19 +91,16 @@ public function getCss(): string
...
         $lastMediaRule = \end($this->mediaRules);
         if ($lastMediaRule !== false && $media === $lastMediaRule->media) {
             return $lastMediaRule;
         }
 
-        $newMediaRule = (object)[
-            'media' => $media,
-            'ruleBlocks' => [],
-        ];
+        $newMediaRule = new CssConcatenator\MediaRule($media);
```
SignpostMarv: constructor defaults ruleBlocks to `[]`, so passing both in seemed redundant, as the change couldn't be described with a 2-line diff without having the constructor first accept an array.

src/Utilities/CssConcatenator/MediaRule.php
```diff
@@ -0,0 +1,45 @@
...
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\Utilities\CssConcatenator;
+
+/**
+ * Implementation of media rules for \Pelago\Emogrifier\Utilities\CssConcatenator
+ *
+ * @internal
+ *
+ * @author Jake Hotson <jake.github@qzdesign.co.uk>
```
SignpostMarv: retained from original source

src/Utilities/CssConcatenator/MediaRule.php
```diff
@@ -0,0 +1,45 @@
...
+ * @author SignpostMarv
```
SignpostMarv: felt this was appropriate due to the transposition of `stdClass` casting & `compact()` to dedicated classes, if that's not an issue?

src/Utilities/CssConcatenator/MediaRule.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+class MediaRule
+{
+    /**
+     * The media query string, e.g. "@media screen and (max-width:639px)", or an empty string for
+     *   rules not within a media query block;
+     *
+     * @var string
+     */
+    public $media;
```
SignpostMarv: possible `@todo` docblock tags:
* amend to `public string $media` when support for <= 7.3 is dropped
* review whether one should use `@readonly` when `vimeo/psalm` is upgraded as part of #793

src/Utilities/CssConcatenator/MediaRule.php
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Array of rule blocks in order, where each element is an object with the following
+     *
+     * @var RuleBlock[]
+     *
+     * @psalm-var array<int, RuleBlock>
+     */
+    public $ruleBlocks = [];
```
SignpostMarv: possible `@todo` docblock tags

* review whether one should use `@readonly` when `vimeo/psalm` is upgraded as part of #793

src/Utilities/CssConcatenator/MediaRule.php
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @param string $media
```
SignpostMarv: This docblock tag appears to be redundant, being needed solely to satisfy phpcs?

src/Utilities/CssConcatenator/RuleBlock.php
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * The property declarations, e.g. "margin-top: 0.5em; padding: 0".
+     *
+     * @var string
+     */
+    public $declarationsBlock;
+
+    /**
+     * @param mixed[] $selectorsAsKeys
+     * @param string $declarationsBlock
```
SignpostMarv: This docblock tag appears to be redundant, being needed solely to satisfy phpcs?

src/Utilities/CssConcatenator/RuleBlock.php
```diff
@@ -0,0 +1,45 @@
...
+{
+    /**
+     * Array whose keys are selectors for the rule block (values are of no
+     *     significance);
+     *
+     * @var mixed[]
+     *
+     * @psalm-var array<string, mixed>
+     */
+    public $selectorsAsKeys = [];
```
SignpostMarv: possible `@todo` docblock tags

* review whether one should use `@readonly` when `vimeo/psalm` is upgraded as part of #793

SignpostMarv: if the values are of no significance, should this be adjusted to `array<string, int>` as implemented, or `array<string, null>` to avoid the temptation to add random ints with special purpose?

src/Utilities/CssConcatenator/RuleBlock.php
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * The property declarations, e.g. "margin-top: 0.5em; padding: 0".
+     *
+     * @var string
+     */
+    public $declarationsBlock;
```
SignpostMarv: possible `@todo` docblock tags:
* amend to `public string $declarationsBlock` when support for <= 7.3 is dropped
* review whether one should use `@readonly` when `vimeo/psalm` is upgraded as part of #793

src/Utilities/CssConcatenator/RuleBlock.php
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @param mixed[] $selectorsAsKeys
+     * @param string $declarationsBlock
+     *
+     * @psalm-param array<string, mixed> $selectorsAsKeys
```
SignpostMarv: impacted by result of discussion on `RuleBlock::$selectorsAsKeys`

SignpostMarv:
> @SignpostMarv, is this something you'd like us to merge now, or hold back on?

I believe that answers that question :D

SignpostMarv:
> There are some open issues about using dedicated classes - notably #733.

didn't catch that!

src/Utilities/CssConcatenator.php
```diff
@@ -145,12 +135,12 @@ private static function getMediaRuleCss(\stdClass $mediaRule): string
     }
 
     /**
-     * @param \stdClass $ruleBlock Object with properties as described for elements of the `ruleBlocks` property of
-     *                            elements of `$mediaRules`.
+     * @param CssConcatenator\RuleBlock $ruleBlock Object with properties as described for elements of the `ruleBlocks`
+     *                            property of elements of `$mediaRules`.
```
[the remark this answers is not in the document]
SignpostMarv: this odd indentation cropped up a fair bit through the original psalm PR.

JakeQZ:
> ~didn't catch that!~ resolved in [eeb2967](https://github.com/MyIntervals/emogrifier/commit/eeb296770fa4394320f6688e8897d629f04d506d)

No way could you have been realistically expected to.  I just happened to be aware that the issue existed.

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#776](https://github.com/MyIntervals/emogrifier/pull/776))
 
 ### Changed
+- `CssConcatenator::$mediaRules` was given dedicated classes
```
oliverklee: I'm a big fan of using the active voice. Please let's change this to: "Give …"

src/Utilities/CssConcatenator.php
oliverklee: Please let's `use` the new classes in `CssConcatenator` so the prefix is not needed.

JakeQZ: It's good to have a list of `use`s at the top of the file to clearly identify what the dependencies are :)

JakeQZ:
> 
> 
> > ~didn't catch that!~ resolved in [eeb2967](https://github.com/MyIntervals/emogrifier/commit/eeb296770fa4394320f6688e8897d629f04d506d)
> 
> No way could you have been realistically expected to. I just happened to be aware that the issue existed.

BTW, avoiding using the `--force` could make reviewing easier.  Or maybe it's an issue with `git commit --amend`.  I don't appear to be able to see this change in isolation now.

CHANGELOG.md
JakeQZ: https://github.com/joelparkerhenderson/git_commit_message

oliverklee:
> Can I suggest we do this as a separate PR?

No, I'd really like to have the code be added together with the tests so we don't have untested code in the repository at any time.

src/Utilities/CssConcatenator.php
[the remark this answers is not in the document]
JakeQZ: The only other realistic alternative I can see is to not indent at all, but put a line break between all the `@param`s, but I've not seen anyone doing that whatsoever.

oliverklee:
We could however split it this way:

1. add the first class + tests
1. add the second class + tests
1. use the new classes and adapt the baseline

src/Utilities/CssConcatenator.php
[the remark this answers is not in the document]
oliverklee: We can most probably fit this in line by removing the prefix and changing the wording.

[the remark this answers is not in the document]
oliverklee: … and also updating the description to match what the parameter now contains.

src/Utilities/CssConcatenator.php
JakeQZ: Absolutely.

Something's broken with Firefox + Github.  I already commented here.

We only had the `[]` initialization because we were using an object-like-array, AFAICR.

src/Utilities/CssConcatenator/MediaRule.php
JakeQZ: OK, thanks :))

src/Utilities/CssConcatenator/MediaRule.php
JakeQZ: By all means, please leave your footprint.  You will be remembered in history :)

If you want to add an email address, you can.  I understand the reasons for not doing so, of course :)

src/Utilities/CssConcatenator/MediaRule.php
JakeQZ: > possible `@todo` docblock tags

Some analysis tools may reject `@todo` in a DocBlock.  I'm pretty sure I've found some to reject `TODO` in a comment.

>     * amend to `public string $media` when support for <= 7.3 is dropped

There'll be lots of similar instances to update when that time comes.  Nothing more to do here.

>     * review whether one should use `@readonly` when `vimeo/psalm` is upgraded as part of #793

Possibly add a note there, but the fact you've x-referenced it should be sufficient :)

src/Utilities/CssConcatenator/MediaRule.php
JakeQZ: Um, also for tools like PhpDocumentor.  They are the reason we had DocBlocks in the first place.

src/Utilities/CssConcatenator/RuleBlock.php
JakeQZ: > if the values are of no significance

That's right, they aren't.  From memory they might be `bool`s (that's the usual construct in this situation).  But I'm still yet to review this PR as a whole...

SignpostMarv:
> Amazing! Now we need unit tests for the new classes

Wouldn't these classes inherit the coverage of the unit tests for CssConcatenator?

src/Utilities/CssConcatenator/RuleBlock.php
JakeQZ: I already replied to this.  GitHub/Firefox going wrong again.  It's needed for tools like PhpDocumentor.

src/Utilities/CssConcatenator/MediaRule.php
SignpostMarv: email address is in the git log :P

src/Utilities/CssConcatenator.php
SignpostMarv: possibly not firefox, I had an issue with a review comment not posting in chrome.

src/Utilities/CssConcatenator/RuleBlock.php
SignpostMarv: the thinking I'm going with here is that a scalar param docblock with no description adds no value that can't already be extracted through reflection/AST

src/Utilities/CssConcatenator/RuleBlock.php
SignpostMarv: at least one of them is a zero.

SignpostMarv:
> BTW, avoiding using the `--force` could make reviewing easier. Or maybe it's an issue with `git commit --amend`. I don't appear to be able to see this change in isolation now.

Yeah, it's an issue with `--force`. 'twas done as I'm not a fan of filling a commit log with reversion commits via PRs.

CHANGELOG.md
SignpostMarv: will take a look tomorrow after I see how step 1 of #807 goes (brain sleepy)

src/Utilities/CssConcatenator.php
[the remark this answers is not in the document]
JakeQZ: > 
> 
> We can most probably fit this in line by removing the prefix and changing the wording.

Might work in this instance, but for future we perhaps need some style guidelines for for formatting DocBlocks (I don't think this is something checkable by any tool right now, given that PSR-5 is still WIP) - see #373.

JakeQZ:
> I'm not a fan of filling a commit log with reversion commits via PRs.

What is merged to `master` will be squashed-and-merged so I don't see a big issue here, apart from that GitHub squash-and-merge is a possibly incomplete feature and needs ability to set the (overall) commit title and message before review.  That is another issue I will look into.

src/Utilities/CssConcatenator/MediaRule.php
JakeQZ: > email address is in the git log :P

Not many people will look at that (I never have).  It's entirely up to you whether you want to include it in the source code, and not a decision I'd take too lightly.  Have a think.  Can be added (or removed) later if you change your mind, though if added first then wanted to be removed, could be impossible removing all traces in copies of the repo etc.  So maybe leave it off for now (i.e. no further changes to make for this conversation in this PR), and can be added in later if you like.

JakeQZ:
> 
> 
> > Amazing! Now we need unit tests for the new classes
> 
> Wouldn't these classes inherit the coverage of the unit tests for CssConcatenator?

GitHub web app wouldn't let me see the source code for the new classes without being cluttered with conversation comments (which are good - it's another issue with GitHub), so I did a `git pull`/`git checkout` on your branch.

I can now see clearly that the new classes are simply property holders (as expected), with a constructor (as expected) and there is literally nothing to test with them.

So AFAICS there is absolutely no requirement for any changes within `tests` :)

(But definitely worth checking up on :>)

src/Utilities/CssConcatenator/RuleBlock.php
JakeQZ: > 
> 
> at least one of them is a zero.

Sharp spot.  Presumably because of a combination of `array_flip` and `$x[] =` pushing to the array.  I've not looked closely at the code right now.

src/Utilities/CssConcatenator/RuleBlock.php
JakeQZ: It also shows that the author of the code has thought about it (and upon reflection realized that their naming of the method/parameter is so good that there is nothing more to add in the DocBlock).  For that reason, I think having a skeletal DocBlock is important, and it should have all ribs.

JakeQZ: A missing DocBlock looks like the programmer forgot to put it in at all.

src/Utilities/CssConcatenator.php
JakeQZ: Both my comments are now here, but one of them is somewhere else.  Weird?  Maybe because this conversation has been copied to another place, or because a certain commit has been rebased and the review comments have been copied to another PR.  This is confusing the H out of me.

I've literally just seen the other one, but closed the tab, and even with "Undo Closed Tabs Button" addon cannot revive it.

I'd really like to be able to work without significant UX bugs in the tools used, i.e. GitHub.

JakeQZ:
I've managed to review this and made comments/suggestions.

> We could however split it this way:
> 
>     1. add the first class + tests
> 
>     2. add the second class + tests
> 
>     3. use the new classes and adapt the baseline

Part of the problem as I see it now is UX bugs in GitHub itself.

But, yes, splitting this up would be good.  And @oliverklee's suggestion sounds good too.  (I was struggling to think of any way to split this.)

oliverklee:
> I can now see clearly that the new classes are simply property holders (as expected), with a constructor (as expected) and there is literally nothing to test with them.
>
> So AFAICS there is absolutely no requirement for any changes within tests :)

I disagree. If something fails, we'd like to know from the tests whether the problem lies in `CssConcatenator` itself or one of the new classes.

src/Utilities/CssConcatenator/RuleBlock.php
oliverklee: Please let's make the properties of the new classes `private` and use proper getters/setters so the properties are encapsulated.

src/Utilities/CssConcatenator/MediaRule.php
oliverklee: I'd prefer to have a real name there instead of a nickname, though.

oliverklee:
I can take care of creating the two PRs for adding the two new classes with tests for each.

src/Utilities/CssConcatenator/RuleBlock.php
SignpostMarv: This might be the bias of having used psalm for so long, but why do the properties need to be encapsulated if static analysis ensures the properties are used appropriately?

oliverklee: For several reasons (in addition to encapsulation of object properties having been a recognized best practice for a really long time):

- it allows us to change the internal representation of the data without having to change the API (and hence without changing the classes that use our class).
- It allows to have type safety during runtime (as arguments and return values can be typed, but fields cannot (yet)). Currently, any other class can write any data into the properties.
- It allows us to add logic to the getters or setters: additional validation, trimming, caching …

SignpostMarv: > * it allows us to change the internal representation of the data without having to change the API

the classes are/should be flagged as internal

> * It allows to have type safety during runtime

in-repo code that would violate type safety during runtime would raise issues in psalm.

> * It allows us to add logic to the getters or setters: additional validation, trimming, caching …

defer getters/settters until additional logic is needed, as it is an internal class?

oliverklee: > the classes are/should be flagged as internal

Even with internal classes, being able to update the internal representation withing having to change the API is a good thing.

@SignpostMarv  Another question: Do you have concerns with encapsulating the fields (i.e., you see potential pitfalls), or is it just not important to you? If it's the former, we'll need to address those concerns, and if it's the latter, then I'd prefer to spend my time implementing this than discussing something that actually is not a problem for anyone concerned.

oliverklee: Then we follow different philosophies here: I think that unit tests, static analysis and manual code reviews complement each other, and that covering code with tests is paramount. (The optimal case is test-driven development.)

SignpostMarv: > (And if I understand you correctly, you do not have any concerns, but just prefer another approach.)

yup- it's like with the tests that were identified as redundant- I'd be the type to remove those because static analysis says they're unneeded.

JakeQZ: > Then we follow different philosophies here: I think that unit tests, static analysis and manual code reviews complement each other, and that covering code with tests is paramount. (The optimal case is test-driven development.)

I'm with @oliverklee here.  Even though some tests and checks might overlap or even aim to do the same thing, they'll inevitably work in different ways, and have their own unique set of limitations.  If there's no overlap, things may fall through the gaps that you couldn't see :)

So I'd much rather have a huge overlap than the possibility of an tiny underlap which turns out to be gaping chasm.  It helps ensure software robustness.  A possible drawback from that approach is overloading Travis, but that's another issue.

BTW, it is good to discuss what approach/methodology is best, and question "are we doing things in the right way?".  Helps achieve perfection (which, philosophically, is never quite possible, but something we should always strive towards), and provides extracurricular things to think about.  I'm all for that :)

SignpostMarv: "the best code is the code not written" :P

oliverklee:

PR #805, opened 2019-10-07
[BUGFIX] partially satisfy psalm for src/HtmlProcessor/AbstractHtmlProcessor.php

SignpostMarv:


JakeQZ:
Yoda is speaking :))

src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -103,9 +103,19 @@ private function setHtml(string $html)
      * Provides access to the internal DOMDocument representation of the HTML in its current state.
      *
      * @return \DOMDocument
+     *
+     * @throws \UnexpectedValueException
      */
     public function getDomDocument(): \DOMDocument
     {
+        if (null === $this->domDocument) {
```
JakeQZ: Yoda not like this :)

src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -103,9 +103,19 @@ private function setHtml(string $html)
...
+            throw new \UnexpectedValueException(
+                self::class .
+                '::setDomDocument() has not yet been called on ' .
+                static::class
+            );
```
JakeQZ: Could we have an exception code based on the current UNIX timestamp?  (It will also mark your change on the timeline of all history, up until at least 2037, etc. :>)

JakeQZ: Also I prefer to have the `.` operators on the beginning of the next line rather than end of the preceding.  But I don't think we have any stylistic rule about that - @oliverklee ?

It might become clearer when you add the 2nd argument - is the code easier to read with the `.` operator at the start of a new line?  YMMV, and it's no big deal...

src/HtmlProcessor/AbstractHtmlProcessor.php
SignpostMarv: looks like this can't be cleanly cherry-picked, so it'll be addressed directly.

SignpostMarv: resolved in 451e379

src/HtmlProcessor/AbstractHtmlProcessor.php
SignpostMarv: > It might become clearer when you add the 2nd argument - is the code easier to read with the `.` operator at the start of a new line? YMMV, and it's no big deal...

my general preference is parenthesis.

SignpostMarv: resolved in 0bf5794 and 0adb26f

JakeQZ: > my general preference is parenthesis.

Good idea.  LGTM now :)

Issue #544, opened 2018-04-02
Refactor the CSS parsing


oliverklee:
We should move the CSS parsing either to a separate class or use an existing package for this, e.g. one of these:

- https://packagist.org/packages/sabberworm/php-css-parser
- https://packagist.org/packages/soloproyectos-php/css-parser

oliverklee:
@JakeQZ Do you have experience with CSS parsing libraries and can recommend one?

JakeQZ:
I've looked at a few PHP CSS minification tools (one of which I use), and they've all used their own regex parsing.  I'm not familiar with either of the above.

JakeQZ:
I've read the readmes of the two suggested possibilities, but have not looked in any detail...

>     * https://packagist.org/packages/sabberworm/php-css-parser

Looks like it would parse the CSS and organize into an OO data structure along the lines of the W3C spec.  Seems to misname 'Declaration' as 'Rule', but the principle and organization appear correct, so that should be a non-issue.  Property values have their own classes (like URL, Color) - we really only care about the string (provided the `!important` flag is separated), but hopefully that won't be an issue.  Actively maintained.  Looks worth investigating further.

>     * https://packagist.org/packages/soloproyectos-php/css-parser

Looks a bit like a jQuery-like extension for PHP DOM, rather than a CSS parser.  Doesn't appear to have been updated for 4 years.  At first glance, doesn't look applicable/useful to Emogrifier.

JakeQZ:
> > ```
> > * https://packagist.org/packages/sabberworm/php-css-parser
> > ```
> 
> Looks like it would parse the CSS and organize into an OO data structure along the lines of the W3C spec.

I've started to look at this in more detail, and begun trying out some changes to our implementation to use it.  So far so good.

I think as a first step, we could use this to parse the raw CSS, then collate the data structure returned and put into the structure we currently use internally, as is returned by `CssInliner::parseCssRules`.

So, initially, we would use this as far as `CssInliner::parseCssRules`, and that could make for a self-contained improvement.  Moving forward from there, we can then look at propagating the `Sabberworm` data structures and objects further along the stream to replace the associative arrays we currently use...

JakeQZ:
Having subconsciously thought about https://github.com/MyIntervals/emogrifier/pull/912#issuecomment-653879492

> move the CSS parsing either to a separate class or use an existing package

 for a while, I think we should probably do both:

<del>or</del><ins>and</ins>

[...]

PR #810, opened 2019-10-08
[WIP][FEATURE] Add a RuleBlock class

oliverklee:
Part of #742.

### DIFF
--- a/src/Utilities/CssConcatenator/RuleBlock.php
+++ b/src/Utilities/CssConcatenator/RuleBlock.php
@@ -1,0 +1,118 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\Utilities\CssConcatenator;
+
+/**
+ * This class represents a CSS rule block, i.e., one ore multiple selectors (like "p") and a declarations block
+ * (like "margin-top: 2pc").
+ *
+ * @internal
+ *
+ * @author Jake Hotson <jake.github@qzdesign.co.uk>
+ * @author SignpostMarv
+ * @author Oliver Klee <github@oliverklee.de>
+ */
+class RuleBlock
+{
+    /**
+     * selectors for the rule block
+     *
+     * @var string[]
+     */
+    private $selectors = [];
+
+    /**
+     * Array whose keys are selectors for the rule block (values are of no significance).
+     * This is redundant to `$selectors`; the data is stored twice for performance reasons.
+     *
+     * @var array
+     *
+     * @psalm-var array<string, mixed>
+     */
+    private $selectorsAsKeys = [];
+
+    /**
+     * the property declarations, e.g. "margin-top: 0.5em;padding: 0"
+     *
+     * @var string
+     */
+    private $declarationsBlock;
+
+    /**
+     * @param string[] $selectors
+     * @param string $declarationsBlock
+     *
+     * @psalm-param array<string, mixed> $selectorsAsKeys
+     */
+    public function __construct(array $selectors, string $declarationsBlock)
+    {
+        $this->selectors = $selectors;
+        $this->selectorsAsKeys = \array_flip($selectors);
+        $this->declarationsBlock = $declarationsBlock;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getSelectors(): array
+    {
+        return $this->selectors;
+    }
+
+    /**
+     * @return array
+     *
+     * @psalm-return array<string, mixed>
+     */
+    public function getSelectorsAsKeys(): array
+    {
+        return $this->selectorsAsKeys;
+    }
+
+    /**
+     * @return string
+     */
+    public function getDeclarationsBlock(): string
+    {
+        return $this->declarationsBlock;
+    }
+
+    /**
+     * @return string CSS for the rule block
+     */
+    public function getCss(): string
+    {
+        return \implode(',', $this->selectors) . '{' . $this->declarationsBlock . '}';
+    }
+
+    /**
+     * @param RuleBlock $otherRuleBlock
+     *
+     * @return bool
+     */
+    public function hasEquivalentSelectors(RuleBlock $otherRuleBlock): bool
+    {
+        $ownSelectorCount = $this->getSelectorCount();
+        if ($ownSelectorCount !== $otherRuleBlock->getSelectorCount()) {
+            return false;
+        }
+
+        if ($this->getSelectorsAsKeys() === $otherRuleBlock->getSelectorsAsKeys()) {
+            return true;
+        }
+
+        $combinedSelectorKeys = $this->getSelectorsAsKeys() + $otherRuleBlock->getSelectorsAsKeys();
+
+        return $ownSelectorCount === \count($combinedSelectorKeys);
+    }
+
+    /**
+     * @return int
+     */
+    private function getSelectorCount(): int
+    {
+        return \count($this->selectors);
+    }
+}
--- a/tests/Unit/Utilities/CssConcatenator/RuleBlockTest.php
+++ b/tests/Unit/Utilities/CssConcatenator/RuleBlockTest.php
@@ -1,0 +1,200 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\Tests\Unit\Utilities\CssConcatenator;
+
+use Pelago\Emogrifier\Utilities\CssConcatenator\RuleBlock;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Test case.
+ *
+ * @covers \Pelago\Emogrifier\Utilities\CssConcatenator\RuleBlock
+ *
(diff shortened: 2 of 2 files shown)


src/Utilities/CssConcatenator/RuleBlock.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Pelago\Emogrifier\Utilities\CssConcatenator;
+
+/**
+ * This class represents a CSS rule block.
+ *
+ * @internal
+ *
+ * @author Jake Hotson <jake.github@qzdesign.co.uk>
+ * @author SignpostMarv
```
oliverklee: @SignpostMarv I'd like to put your real name here instead of a nickname (and also an email address if that's okay with you). What name would you like to have here, and which email address (if any)?

src/Utilities/CssConcatenator/RuleBlock.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    private $declarationsBlock;
+
+    /**
+     * @param array $selectorsAsKeys
+     * @param string $declarationsBlock
+     *
+     * @psalm-param array<string, mixed> $selectorsAsKeys
+     */
+    public function __construct(array $selectorsAsKeys, string $declarationsBlock)
```
oliverklee: @JakeQZ, @SignpostMarv Does is make sense that an empty array and an empty string can be passed, or would this indicate a problem upstream in the data flow, and an throwing an exception would be in order?

oliverklee: @JakeQZ Does it make sense to expect the selectors as keys here, or can we move the key/value reversal into this class?

oliverklee:
@JakeQZ Is there any functionality in `CssConcatenator` which we currently execute on the array (that now is going to be a class) and which we can move into this class?

src/Utilities/CssConcatenator/RuleBlock.php
SignpostMarv: do variants on `;\n;` create a declaration with an empty string ?

SignpostMarv: tbh, I was a little confused as to why the selectors are retained as keys, as opposed to `array<int, string>`.

src/Utilities/CssConcatenator/RuleBlock.php
SignpostMarv: I'm afraid my "real name" only goes on my day job code, and even the it's not my *legal name* I'm using.

oliverklee: @SignpostMarv Do you have a "preferred name"?

SignpostMarv: "SignpostMarv"

tests/Unit/Utilities/CssConcatenator/RuleBlockTest.php
```diff
@@ -0,0 +1,54 @@
...
+
+use Pelago\Emogrifier\Utilities\CssConcatenator\RuleBlock;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Test case.
+ *
+ * @author Oliver Klee <github@oliverklee.de>
+ *
+ * @covers \Pelago\Emogrifier\Utilities\CssConcatenator\RuleBlock
```
oliverklee: Note: I've already switched positions of the `@author` and `@covers` annotations in my local copy.

src/Utilities/CssConcatenator/RuleBlock.php
oliverklee: :-) Okay, then I'll keep it.

SignpostMarv: you can either view it as a camel-cased mononym, or the use of "Signpost" as a prefixing honorific to a mononym with a non-existent zero-width space separating the two. I tend to lean towards the former these days.

src/Utilities/CssConcatenator/RuleBlock.php
JakeQZ: If we choose to throw an exception in either case we should catch it in `CssInliner` if not in 'debug' mode, convert it to a `trigger_error`, and then carry on (the same as what we do when we get an exception from `CssSelectorConverter`).

I think the selectors array will necessarily be nonempty because I don't think CSS like `/* no selector */ { color: red; }` will get through to here.  But I'm less certain what happens with CSS like `body { /* non declarations */ }`

JakeQZ: The reason for them being stored as keys is to be able to more efficiently (i.e. in O(nlogn) rather than O(n2) time) do the following:
- combine a set of selectors without ending up with duplicates;
- test if two sets of selectors are equivalent, but possibly in a different order.

But looking at `CssConcatenator::append()`, it looks like we could move the bulk of the logic there into this new class.  We could have a method named something like `maybeMerge` which is passed another `RuleBlock` object, and if possible combines `$that` with `$this`, returning `true` if it was able to and has done so.  If it returns `false`, `CssConcatenator::append()` would then add it to the end of `$mediaRule->ruleBlocks`.  (If the new `RuleBlock` was merged, it would only exist for the lifetime of `append()`, being temporarily created to be passed as the parameter to `maybeMerge()`.)

Taking this approach would allow the constructor to be passed the selectors as a normal 'unflipped' array of strings, but store them internally in its class property as the result of `array_flip()`.  I think it would tidy the logic up quite nicely...

JakeQZ:
> 
> 
> @JakeQZ Is there any functionality in `CssConcatenator` which we currently execute on the array (that now is going to be a class) and which we can move into this class?

Yes - see https://github.com/MyIntervals/emogrifier/pull/810#discussion_r332700721

src/Utilities/CssConcatenator/RuleBlock.php
JakeQZ: > convert it to a `trigger_error`

Checking this, we actually silently ignore the exceptions if not in debug mode.  And with good reason I think - `trigger_error` could clutter up error logs if there's an upstream issue in a third-party component that can't be easily resolved.  (There might be a justification for three 'debug' options, with an additional one that does a `trigger_error`, but that's another matter.)

src/Utilities/CssConcatenator/RuleBlock.php
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * @return array
+     *
+     * @psalm-return array<string, mixed>
+     */
+    public function getSelectorsAsKeys(): array
+    {
+        return $this->selectorsAsKeys;
+    }
```
JakeQZ: With the suggested refactoring, we'll probably prefer instead a method `getSelectors` which returns `\array_keys($this->selectorsAsKeys)`.

src/Utilities/CssConcatenator/RuleBlock.php
SignpostMarv: Make the constructor non-public and validate before it hits the constructor?

src/Utilities/CssConcatenator/RuleBlock.php
```diff
@@ -0,0 +1,118 @@
...
+ * @author Oliver Klee <github@oliverklee.de>
+ */
+class RuleBlock
+{
+    /**
+     * selectors for the rule block
+     *
+     * @var string[]
+     */
+    private $selectors = [];
```
JakeQZ: I'm wondering whether trying to also maintain this property would cause efficiency issues, e.g. updating it after every merge from an increasingly growing list of selectors that have come from something like a CSS reset (having been split up in original parsing).  Would it be better to 'lazily' obtain this (via `array_keys`) when it's finally needed, i.e. when outputting the final CSS?

tests/Unit/Utilities/CssConcatenator/RuleBlockTest.php
```diff
@@ -0,0 +1,200 @@
...
+    public function hasEquivalentSelectorsForNotEquivalentSelectorsReturnsFalse(array $ours, array $theirs)
+    {
+        $ourRuleBlock = new RuleBlock($ours, 'color: #000;');
+        $theirsRuleBlock = new RuleBlock($theirs, 'color: #000;');
+
+        self::assertFalse($ourRuleBlock->hasEquivalentSelectors($theirsRuleBlock));
+    }
+
+    // TODO:
+    // - convertDeclarationsBlocks to array internally
```
JakeQZ: We're currently not doing this because we don't need to for how this is currently used (i.e. putting the applicable uninlinable rules into a `<style>` element).  It could be useful if we're also going to use this class during CSS parsing.  But we might want to alternatively take a look at e.g. [`sabberworm/php-css-parser`](https://github.com/sabberworm/PHP-CSS-Parser) (see https://github.com/MyIntervals/emogrifier/issues/544#issuecomment-526967555) sooner rather than later to avoid duplicating effort...

oliverklee:

PR #812, opened 2019-10-08
[TASK] Add @covers annotations to the test cases

### LINKED ISSUE #811: Add @covers annotations to the test cases

oliverklee:
This makes sure that only dedicated tests for a class will mark that
class as covered.

Fixes #811

### DIFF
--- a/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
+++ b/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
@@ -11,7 +11,7 @@ use PHPUnit\Framework\TestCase;
 /**
  * Test case.
  *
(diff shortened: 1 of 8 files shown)


tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -11,6 +11,8 @@
 /**
  * Test case.
  *
+ * @covers \Pelago\Emogrifier\HtmlProcessor\AbstractHtmlProcesso
```
JakeQZ: missing "r" at end of class name.

oliverklee: Good catch! Fixed and repushed.

PR #817, opened 2019-10-27
[BUGFIX] Restore PHP 7.0 compatibility in the dev dependencies

oliverklee:
The package "slevomat/coding-standard" in version >= 4 requires
PHP >= 7.1. Due to a stale Composer cache in Travis CI, the build for
PHP 7.0 still passed with that requirement.

Now Emogrifier uses version 3.3 of that package, allowing the dev
dependencies to install again.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         "phpmd/phpmd": "^2.7.0",
         "phpunit/phpunit": "^6.5.14",
         "psalm/plugin-phpunit": "^0.5.8",
-        "slevomat/coding-standard": "^4.8.7",
+        "slevomat/coding-standard": "^4.0.0",
         "squizlabs/php_codesniffer": "^3.5.1",
         "vimeo/psalm": "^3.2.12"
     },


composer.json
```diff
@@ -48,7 +48,7 @@
         "phpmd/phpmd": "^2.7.0",
         "phpunit/phpunit": "^6.5.14",
         "psalm/plugin-phpunit": "^0.5.8",
-        "slevomat/coding-standard": "^4.8.7",
+        "slevomat/coding-standard": "^3.3.0",
```
JakeQZ: I think we should actually have `"^4.0.0"` (i.e. as it was before #814) - it seems that `slevomat/coding-standard` has not strictly adhered to semantic versioning - [version 4.0](https://packagist.org/packages/slevomat/coding-standard#4.0.0) is in fact compatible with PHP 7.0 whereas [version 4.1](https://packagist.org/packages/slevomat/coding-standard#4.1.0) requires PHP 7.1.  This would also allow a later version (which may have useful improvements) to be installed when running CI scripts against PHP 7.1+.

oliverklee: Oh, good research! Updated and repushed.

PR #821, opened 2019-12-03
[FEATURE] Add support for PHP 7.4

### LINKED ISSUE #819: Test with PHP 7.4
We probably need to disable the platform requirements constrains for Composer for this.

oliverklee:
Fixes #819

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ php:
 - "7.1"
 - "7.2"
 - "7.3"
+- "7.4"
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


oliverklee:
PHP-parallel-lint issue:
https://github.com/JakubOnderka/PHP-Parallel-Lint/issues/154

PHPMD issue:
https://github.com/phpmd/phpmd/issues/658

oliverklee:
We need a way to have the newer version of Psalm on PHP 7.4.

oliverklee:
This PR now is ready for review.

oliverklee:
I have just released version 3.1.0 (including support for PHP 7.4 and Symfony 5) and posted about it on social media.

JakeQZ:
Assuming you want to prevent install on PHP 7.5 when it becomes available, until explicitly supported by Emogrifier, the `composer.json` change was incorrect.

I have found that [tilde version constraints](https://getcomposer.org/doc/articles/versions.md#tilde-version-range-) require the patch release version number to be supplied otherwise they behave like caret version constraints - i.e. `~7.4` means >=7.4 and <8.0 whereas `~7.4.0` means >=7.4 and <7.5.

oliverklee:
Oh, I wasn't aware of this.

JakeQZ:
Neither was I until playing around with Psalm versions earlier today for #792.

oliverklee:
As there will be no PHP 7.5, this does not matter in this case. For the future, we know better now.

Issue #825, opened 2019-12-12
Pass CssInliner as a dependency


SpazzMarticus:
Since `CssInliner` has to be created by passing html to `::fromHtml` it can not be **passed as a dependency**.

Current version 3.0.0 in a PSR-7/PSR-15 example:

```php
use Psr\Http\Message\ResponseInterface;
use Psr\Http\Message\ServerRequestInterface;
use Psr\Http\Server\MiddlewareInterface;
use Pelago\Emogrifier\CssInliner;

class CssInlineMiddleware implements Psr\Http\Server\MiddlewareInterface{
    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface {
        $response = $handler->handle($request);
        $inlineHtml = CssInliner->fromHtml($response->getBody()->getContents())->inlineCss()->render();
        $response->withBody($inlineHtml);
    }
}
```

One simple solution I can think of would be a factory a la:
```php
namespace Pelago\Emogrifier;

class CssInlinerFactory
{
    public function generate($unprocessedHtml): CssInliner {
        return CssInliner::fromHtml($unprocessedHtml);
    }
}
```

Any chance you could make passing `CssInliner` as a dependency as simple as it was with the now deprecated `Emogrifier`.

Cheers!

oliverklee:
@SpazzMarticus Thanks for the feature proposal!

Would you be willing to create a PR for this?

SpazzMarticus:
Of course! PR coming soon.

JakeQZ:
Continuing the discussion from #826...

> (An interface would help if there were multiple packages with different APIs that provide the same functionality as this repo does.)

Or if one wanted to allow functionality to be provided via an optional package (e.g. "suggested" in `composer.json`)?  I have no idea how prevelant a use case that might be.

> If, instead of calling `CssInliner::fromHtml` , one wants to avoid the static method call and/or the thight coupling, they might want to implement their own service, helper-class, or what-ever-class, with an API that fit's their use case.

> Simple use case: A global stylesheet should be applied to user-generated html code.
> 
> ```php
> class StylesheetService
> {
>     protected string $stylesheet;
> 
>     public function inlineStylesheet(string $html): string
>     {
>         return CssInliner::fromHtml($html)->inlineCss($this->stylesheet)->render();
>     }
> }
> ```

I was coming round to the idea that this would be a good general solution to the problem, and could be included in the package.

[...]

PR #826, opened 2019-12-12
[FEATURE] Add Factory to pass as dependency

SpazzMarticus:
CssInliner has to be created by calling the static fromHtml or fromDomDocument methods. This means CssInliner can not be passed as a dependency. Passing the factory as a dependency allows using CssInliner (more) like the now deprecated Emogrifier.

Psalm - which I haven't worked with before - is currently nagging about: `ERROR: UnusedClass - src/CssInlinerFactory.php:15:1 - Class Pelago\Emogrifier\CssInlinerFactory is never used` 

I created a unit test to test for the returned instance to be of type `CssInliner`, which Psalm thinks of as redunant, because this behaviour is already enforced by the defined return type.

```
ERROR: RedundantCondition - tests\Unit\CssInlinerFactoryTest.php:42:15 - Found a redundant condition when evaluating $cssInliner and trying to reconcile type 'Pelago\Emogrifier\CssInliner' to Pelago\Emogrifier\CssInliner
        self::assertInstanceOf(CssInliner::class, $cssInliner);
```

Would you like a redundant test, or no test at all?

### DIFF
--- a/config/phpmd.xml
+++ b/config/phpmd.xml
@@ -7,7 +7,7 @@
     <!-- The commented-out rules will be enabled once the code does not generate any warnings anymore. -->
 
     <rule ref="rulesets/cleancode.xml/BooleanArgumentFlag"/>
-    <!--<rule ref="rulesets/cleancode.xml/StaticAccess"/>-->
+    <rule ref="rulesets/cleancode.xml/StaticAccess"/>
 
     <rule ref="rulesets/codesize.xml/CyclomaticComplexity"/>
     <rule ref="rulesets/codesize.xml/NPathComplexity"/>
--- a/config/phpmd.xml
+++ b/config/phpmd.xml
@@ -7,7 +7,7 @@
     <!-- The commented-out rules will be enabled once the code does not generate any warnings anymore. -->
 
     <rule ref="rulesets/cleancode.xml/BooleanArgumentFlag"/>
-    <rule ref="rulesets/cleancode.xml/StaticAccess"/>
+    <!--<rule ref="rulesets/cleancode.xml/StaticAccess"/>-->
 
     <rule ref="rulesets/codesize.xml/CyclomaticComplexity"/>
     <rule ref="rulesets/codesize.xml/NPathComplexity"/>
--- a/src/CssInlinerFactory.php
+++ b/src/CssInlinerFactory.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier;
+
+use DOMDocument;
+
+/**
+ * This class provides a factory for CssInliner that can be passed as a dependency
+ * or be injected via dependency injection
+ *
+ * @author SpazzMarticus <SpazzMarticus@users.noreply.github.com>
+ */
+class CssInlinerFactory
+{
+    /**
+     * Creates a new instance of CssInliner from a string
+     *
+     * @param string $unprocessedHtml
+     *
+     * @return CssInliner
+     *
+     * @throws \InvalidArgumentException if $unprocessedHtml is anything other than a non-empty string
+     */
+    public function createFromHtml(string $unprocessedHtml): CssInliner
+    {
+        return CssInliner::fromHtml($unprocessedHtml);
+    }
+
+    /**
+     * Creates a new instance of CssInliner from a DomDocument
+     *
+     * @param DOMDocument $document
+     *
+     * @return CssInliner
+     */
+    public function createFromDomDocument(DOMDocument $document): CssInliner
+    {
+        return CssInliner::fromDomDocument($document);
+    }
+}
--- a/src/Factories/CssInlinerFactory.php
+++ b/src/Factories/CssInlinerFactory.php
@@ -2,9 +2,10 @@
 
 declare(strict_types=1);
 
-namespace Pelago\Emogrifier;
+namespace Pelago\Emogrifier\Factories;
 
 use DOMDocument;
+use Pelago\Emogrifier\CssInliner;
 
 /**
  * This class provides a factory for CssInliner that can be passed as a dependency
@@ -22,6 +23,8 @@ class CssInlinerFactory
      * @return CssInliner
      *
      * @throws \InvalidArgumentException if $unprocessedHtml is anything other than a non-empty string
+     *
+     * @SuppressWarnings(PHPMD.StaticAccess)
      */
     public function createFromHtml(string $unprocessedHtml): CssInliner
     {
@@ -29,11 +32,13 @@ class CssInlinerFactory
     }
 
     /**
-     * Creates a new instance of CssInliner from a DomDocument
+     * Creates a new instance of CssInliner from a DOM document
      *
-     * @param DOMDocument $document
+     * @param \DOMDocument $document a DOM document returned by getDomDocument() of another instance
      *
      * @return CssInliner
+     *
+     * @SuppressWarnings(PHPMD.StaticAccess)
      */
     public function createFromDomDocument(DOMDocument $document): CssInliner
     {
--- a/tests/Unit/CssInlinerFactoryTest.php
+++ b/tests/Unit/CssInlinerFactoryTest.php
@@ -34,6 +34,8 @@ class CssInlinerFactoryTest extends TestCase
 
     /**
      * @test
+     *
+     * @psalm-suppress RedundantCondition
      */
     public function createsInstanceFromHtmlString()
     {
@@ -43,6 +45,8 @@ class CssInlinerFactoryTest extends TestCase
 
     /**
      * @test
+     *
+     * @psalm-suppress RedundantCondition
      */
     public function createsInstanceFromDomDocument()
     {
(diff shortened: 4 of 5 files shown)


SpazzMarticus:
Intended test:
```php
<?php

declare(strict_types=1);

namespace Pelago\Emogrifer\Tests\Unit;

use DOMDocument;
use Pelago\Emogrifier\CssInliner;
use Pelago\Emogrifier\CssInlinerFactory;
use PHPUnit\Framework\TestCase;

/**
 * Test case.
 *
 * @covers \Pelago\Emogrifier\CssInlinerFactory
 *
 * @author SpazzMarticus <SpazzMarticus@users.noreply.github.com>
 */
class CssInlinerFactoryTest extends TestCase
{
    /**
     * @var CssInlinerFactory
     */
    protected $factory;

    /**
     * Setup factory for tests
     */
    protected function setUp()
    {
        parent::setUp();
        $this->factory = new CssInlinerFactory();
    }

    /**
     * @test
     */
    public function createsInstanceFromHtmlString()
    {
        $cssInliner = $this->factory->createFromHtml(CssInlinerTest::COMMON_TEST_HTML);
        self::assertInstanceOf(CssInliner::class, $cssInliner);
    }

    /**
     * @test
     */
    public function createsInstanceFromDomDocument()
    {
        $document = new DOMDocument();
        $document->loadHTML(CssInlinerTest::COMMON_TEST_HTML);

        $cssInliner = $this->factory->createFromDomDocument($document);

        self::assertInstanceOf(CssInliner::class, $cssInliner);
    }
}

```

oliverklee:
> Would you like a redundant test, or no test at all?

Yes, please let's have the test even if it's redundant with the code.

JakeQZ:
Sorry I am late in on this.

Of course it is a good design pattern.

But do we @oliverklee really want to have to maintain this?

It is after all pretty much two lines of code to make the adaptor to fit your circumstance.

IMO we can provide an example class in the readme.

Sorry to be disparaging but we are trying to reduce the maintenance burden and focus only on core functionality.  Do one thing.  And do it well.

thomasfw:
Hi, I see this idea is so far rejected but this is a +1 for adding a Factory class.

> "It is after all pretty much two lines of code to make the adaptor to fit your circumstance."

This is arguably not a rare use-case or one-off circumstance. Dependency injection is so commonplace that this should be possible without requiring the implementor to create their own.

Please re-consider :)

oliverklee:
Yes, please let's have a factory so Emogrifier can be injected in some way. And yes, I think this is worth adding to the code and maintaining.

config/phpmd.xml
```diff
@@ -7,7 +7,7 @@
     <!-- The commented-out rules will be enabled once the code does not generate any warnings anymore. -->
 
     <rule ref="rulesets/cleancode.xml/BooleanArgumentFlag"/>
-    <rule ref="rulesets/cleancode.xml/StaticAccess"/>
+    <!--<rule ref="rulesets/cleancode.xml/StaticAccess"/>-->
```
oliverklee, line 10: Please let's not deactive any Psalm checks. Instead, either let's please fix the issues (which I'd prefer) or add annotations to the code to deactivate the respective check for a single line.

src/CssInlinerFactory.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier;
```
oliverklee, line 5: Please let's move this to a dedicated namespace, e.g., `Pelago\Emogrifier\Factories`. (Suggestions and feedback concerning the namespace name welcome.)

src/CssInlinerFactory.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Creates a new instance of CssInliner from a string
+     *
+     * @param string $unprocessedHtml
+     *
+     * @return CssInliner
+     *
+     * @throws \InvalidArgumentException if $unprocessedHtml is anything other than a non-empty string
+     */
+    public function createFromHtml(string $unprocessedHtml): CssInliner
```
oliverklee, line 26: This now is limited to only `CssInliner`. Should we make this more flexible?

src/CssInlinerFactory.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /**
+     * Creates a new instance of CssInliner from a DomDocument
+     *
+     * @param DOMDocument $document
+     *
+     * @return CssInliner
+     */
+    public function createFromDomDocument(DOMDocument $document): CssInliner
```
oliverklee, line 38: Do we need this? Usually, `fromDomDocument` should not be called by users, but only internally (or with data from other `HtmlProcessor` classes).

SpazzMarticus: Since the [more complex example](https://github.com/MyIntervals/emogrifier#more-complex-example) chains together different classes using `fromDomDocument` I would keep it in.

src/CssInlinerFactory.php
SpazzMarticus: I see multiple options here:

- Factory per class `CssInlinerFactory`, `HtmlPrunerFactory`, ...
- Factory with a method per class `createCssInlinerFromHtml($html)`, `createHtmlPrunerFromHtml($html)`
- Dynamic factory à la `createFromHtml($class,$html)`

I am not in favor of the third option, since checking if `$class` is a child of `AbstractHtmlProcessor` is necessary and returning a `AbstractHtmlProcessor` breaks typehinting to a specific class.

oliverklee:
@SpazzMarticus I'm still trying to picture how the factory would be used in real-world code. Would you be willing to provide an example? That would help me. Thanks!

SpazzMarticus:
Since @JakeQZ turned the factory down I implemented a simple service instead of trying to inject `CssInliner`, which works fine. Instead of injecting a factories to create instances of `CssInliner`, `HtmlPruner`, ... into controllers which might need them, I inject a service which uses `CssInliner`, ... directly. 

Simple use case: A global stylesheet should be applied to user-generated html code.

```php
class StylesheetService{

    protected string $stylesheet;

    public function inlineStylesheet(string $html): string {
        return CssInliner::fromHtml($html)->inlineCss($this->stylesheet)->render();
    }
}
``` 

The service is tightly coupled with your package, but the coupling is only happening in one specific place.

Maybe @thomasfw can chime in and show his use case?

I get why @JakeQZ does not want to polute the package with factories and would be fine with closing the pull request.

JakeQZ:
> Since @JakeQZ turned the factory down 

My concern was more that this would be use-case specific, and that by supporting one particular use case with additional classes within the package, we would be obliged to extend that support to cover other cases (i.e. feature creep).

The behaviour of the deprecated `Emogrifier` class was controlled by various option-setting methods, with the functionality invoked through a single `emogrify` method.  That made is possible to have a method to which an `Emogrifier` object (with desired options set) could optionally be passed:

```php
public function doSomething(?Emogrifier $emogrifier)
{
    // ...
    if ($emogrifier !== null) {
        $emogrifier->setHtml($html);
        $html = $emogrifier->emogrify();
    }
    // ...
}
```

However, with `CssInliner`, non-core functionality (some of which was present in `Emogrifier` via options, such as creating legacy attributes like `align` or removing elements with `display:none`) has been moved to separate classes, and is invoked by specific method calls.  Even if factory objects were provided to instantiate those classes, how would it be determined which of their specific methods should be called?

With just a factory for `CssInliner`, only the core functionality provided by `inlineCss()` would be available, unless it was (ab)used to create something other than a `CssInliner` but which nonetheless has an `inlineCss` method to instantiate and invoke the various classes as desired.

> I implemented a simple service instead of trying to inject `CssInliner`, which works fine. Instead of injecting factories to create instances of `CssInliner`, `HtmlPruner`, ... into controllers which might need them, I inject a service which uses `CssInliner`, ... directly.
> 
> ```php
> class StylesheetService
> ```

This looks closer to something that would have a more general-purpose application – i.e. would be not so use-case specific, and allow greater flexibility.

In the very general sense, the use case is “optionally provide an object which can be used to transform some HTML” (and this doesn’t necessarily need to be Emogrifier-based, it could be anything).  I found [something like this in Drupal](https://api.drupal.org/api/drupal/core!lib!Drupal!Component!Render!OutputStrategyInterface.php/interface/OutputStrategyInterface), with the exception that their method is `static`.

This leads to questions: in which package should the interface be defined (noting that a nullable type hint _can_ be specified for a type in an unavailable package)? if Emogrifier, should any concrete implementations be provided?

SpazzMarticus:
I dont't think providing an interface will help much and potentially too become topic of discussion. As of my knowledge your implementation is not _competing_ with other packgages. (A interface would help if there were multiple packages with different APIs that provide the same functionality as this repo does.)

> Do one thing. And do it well.

You already do that - very well - and my proposed factories don't add much value to your code.

> It is after all pretty much two lines of code to make the adaptor to fit your circumstance.

If, instead of calling `CssInliner::fromHtml` , one wants to avoid the static method call and/or the thight coupling, they might want to implement their own service, helper-class, or what-ever-class, with an API that fit's their use case.

(Since I was the one that suggested adding (syntax sugar) factories: Thank you for your time and explanations!)

JakeQZ:
Many thanks for your input so far, much appreciated.  I think there clearly is an issue at stake, and it's good that you have raised it and proposed various solutions.  I've moved the discussion back to #825 - we should certainly consider doing something...

PR #831, opened 2019-12-28
Added masterminds/html5

### LINKED ISSUE #828: Use HTML5 parser instead of DOMDocument
ref; https://github.com/Masterminds/html5-php

bytestream:
Resolves #828


tests/Unit/HtmlProcessor/CssToAttributeConverterTest.php
```diff
@@ -171,8 +171,8 @@ public function notMatchingCssToHtmlMappingDataProvider(): array
             'span.float' => ['<span style="float: right;">Hi</span>'],
             'float: none' => ['<table style="float: none;"></table>'],
             'p.border-spacing' => ['<p style="border-spacing: 5px;">Hi</p>'],
-            'height: auto' => ['<img src="logo.png" alt="" style="height: auto;">'],
-            'width: auto' => ['<img src="logo.png" alt="" style="width: auto;">'],
+            'height: auto' => ['<img src="logo.png" alt style="height: auto;">'],
+            'width: auto' => ['<img src="logo.png" alt style="width: auto;">'],
```
oliverklee: Is this valid HTML?

oliverklee:
Hi @bytestream, thanks for the PR! :heart:

Could you please fix the PHPMD warning? Thanks!

tests/Unit/HtmlProcessor/CssToAttributeConverterTest.php
bytestream: Depends on the DOCTYPE. It's not valid XHTML, it's valid HTML:
> Attributes are placed inside the start tag, and consist of a name and a value, separated by an "=" character. The attribute value can remain unquoted if it doesn't contain ASCII whitespace or any of " ' ` = < or >. Otherwise, it has to be quoted using either single or double quotes. The value, along with the "=" character, can be omitted altogether if the value is the empty string.

```html
<!-- empty attributes -->
<input name=address disabled>
<input name=address disabled="">

<!-- attributes with a value -->
<input name=address maxlength=200>
<input name=address maxlength='200'>
<input name=address maxlength="200">
```

https://html.spec.whatwg.org/multipage/introduction.html#intro-early-example

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -86,10 +86,10 @@ public function reformatsHtml()
...
             '<body></body>' .
             '</html>';
-        $formattedHtml = "<!DOCTYPE HTML>\n" .
-            "<html>\n" .
-            '<head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head>' . "\n" .
-            "<body></body>\n" .
+        $formattedHtml = "<!DOCTYPE html>\n" .
+            "<html>" .
+            '<head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head>' .
+            "<body></body>" .
```
bytestream: masterminds/html5 doesn't support reformatting (`$dom->formatOutput`) so that's why this has changed

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -629,9 +629,14 @@ public function getDomDocumentWithNormalizedHtmlRepresentsTheGivenHtml()
...
-        $domDocument = $subject->getDomDocument();
+        self::assertSame("<!DOCTYPE html>
+<html><head><meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\"></head>
+<body>
+<br>
+</body>
 
-        self::assertSame($html, $domDocument->saveHTML());
+</html>
+", $subject->getHtml5()->saveHTML($subject->getDomDocument()));
```
bytestream: Struggled to debug this, I think there's a rogue `\r` which `masterminds/html5` is converting to `\n` so you get `\n\n` instead of `\n` in between the `</body></html>`

bytestream:
> 
> 
> Hi @bytestream, thanks for the PR! ❤️
> 
> Could you please fix the PHPMD warning? Thanks!

Thanks for checking! I'll look at this soon

src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -228,7 +265,8 @@ private function createRawDomDocument(string $html)
     private function prepareHtmlForDomConversion(string $html): string
     {
         $htmlWithSelfClosingSlashes = $this->ensurePhpUnrecognizedSelfClosingTagsAreXml($html);
-        $htmlWithDocumentType = $this->ensureDocumentType($htmlWithSelfClosingSlashes);
+        $htmlWithRootElms = $this->addMissingRootElements($htmlWithSelfClosingSlashes);
```
JakeQZ: Nitpicky, but we would prefer `Elements` to `Elms` - if it's good enough for a method name it's good enough for a variable :)

src/HtmlProcessor/HtmlParser.php
```diff
@@ -0,0 +1,208 @@
...
+     * @return void
+     *
+     * @psalm-suppress MixedOperand
+     */
+    public function loadHtml(string $html)
+    {
+        $i = 0;
+        $len = \strlen($html);
+        while ($i < $len) {
+            if ($html[$i] == '<') {
```
JakeQZ: Should be `===`?  Also in a few other places.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -86,10 +86,10 @@ public function reformatsHtml()
...
             '<body></body>' .
             '</html>';
-        $formattedHtml = "<!DOCTYPE HTML>\n" .
-            "<html>\n" .
-            '<head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head>' . "\n" .
-            "<body></body>\n" .
+        $formattedHtml = "<!DOCTYPE html>\n" .
+            '<html>' .
+            '<head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head>' .
+            '<body></body>' .
```
JakeQZ: I'm not satisfied there's a good reason to change the tests here.  We should permit all sorts of slightly malformed HTML just like a browser.  Because historically we have through PHP DOM and we don't want to break people's implementations...

If the new HTML5 component barfs on it, can we fall back to DOMDocument?

src/HtmlProcessor/HtmlParser.php
```diff
@@ -0,0 +1,208 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\HtmlProcessor;
+
+/**
+ * Parse a HTML document fragment and add missing root elements.
```
JakeQZ: Is this a class you've used elsewhere?  I'd like more information on why it is needed, and what deficiencies in the HTML5 library it is covering.

tests/Unit/HtmlProcessor/CssToAttributeConverterTest.php
JakeQZ: I see no reason to change this, the RNIB recommendation is `alt=""`.  We can add a separate test for `alt`.  We should not change the tests to allow the code to pass; we should fix the code to pass the tests - unless the tests are demonstrably wrong ;)

JakeQZ:
Incidentally, PHP8 promises some DOM improvements: https://wiki.php.net/rfc/dom_living_standard_api

oliverklee:
> I also wonder if we need some more test cases - some that are HTML4 or XHTML and some HTML5?

Yes, please. Particularly, I'd strongly prefer we get the new parser class in as a separate PR first (with good and complete tests).

src/HtmlProcessor/HtmlParser.php
bytestream: I wrote it but it's not yet been released to any customers. It's needed to handle https://github.com/Masterminds/html5-php/issues/166. You can see two different implementations here:
* https://github.com/roundcube/roundcubemail/blob/master/program/lib/Roundcube/rcube_washtml.php#L842
* https://github.com/ezyang/htmlpurifier/blob/master/library/HTMLPurifier/Lexer/DOMLex.php#L310

It's not strictly required in the context of this lib if you're happy for it to barf on malformed HTML

src/HtmlProcessor/HtmlParser.php
bytestream: Yeah

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
bytestream: This relates to https://github.com/MyIntervals/emogrifier/pull/831#discussion_r362020058

JakeQZ: Sorry, I was looking at the change out of context.

I'm wondering if there's any requirement that the HTML is reformatted with newlines added - I can't think of any reason.  Isn't the test more about confirming that everything apart from `<body>` (which is tested by the previous test, `renderRendersDocumentProvidedToFromDomDocument`) is functionally preserved (e.g. some superfluous whitespace _may_ be added outside of `<body>`).  @oliverklee ?

Most of this is in fact case insensitive (I think everything except `text/html; charset`).  However, allowing an alternative case representation may cause issues with some email clients that are not fully compliant (as we know, most email clients are not fully compliant).

To avoid issues, the [recommendation](https://www.w3.org/TR/html-polyglot/#doctype) for [robustness](https://www.w3.org/TR/html-polyglot/#dfn-polyglot-markup) is uppercase `DOCTYPE` and lowercase `html`.  I think we should require this, and thus may need convert the `HTML` to `html` in the `render` method.

oliverklee: > I'm wondering if there's any requirement that the HTML is reformatted with newlines added 

No, there is none. That's just the way the HTML comes out of `DOMDocument`, and parsing the HTML in the tests in a formatting-agnostic way just had not been necessary in the tests yet.

src/HtmlProcessor/HtmlParser.php
JakeQZ: > It's needed to handle [Masterminds/html5-php#166](https://github.com/Masterminds/html5-php/issues/166). You can see two different implementations here:
> 
>     * https://github.com/roundcube/roundcubemail/blob/master/program/lib/Roundcube/rcube_washtml.php#L842
> 
>     * https://github.com/ezyang/htmlpurifier/blob/master/library/HTMLPurifier/Lexer/DOMLex.php#L310

Thanks.  I think I understand better now: `masterminds/html5-php` treats input as a document fragment rather than a complete document (even if it has a `DOCTYPE`??) and does not (currently) adhere the W3C spec for `<html>`. `<head>` and perhaps `<body>` elements (linked from the discussion in Masterminds/html5-php#166).

It would be good to put all this in the PHPDoc for the class.  (Now I read again what you have put, with the benefit of having looked at the discussion, I see that it's a nice succinct summary 😊 but it wasn't immediately obvious what it was all about.)

> It's not strictly required in the context of this lib if you're happy for it to barf on malformed HTML

Adding missing `DOCTYPE`, `<html>`, `<head>` and `<body>` elements has for a long time been part of Emogrifier.  I think we would end up breaking people's code without them realizing if we stopped doing this.

(This does make me wonder, however, if `Masterminds/html5-php` actually deals properly with all the rules of HTML5 where closing tags can be completely omitted depending what element is next or what the parent is.  E.g. as a simple example, `<div><p>First pargraph in 1st DIV.<p>Second paragraph in 1st DIV<div>Sibling of 1st DIV` is (I think) perfectly valid HTML [though I suspect `DOMDocument` doesn't deal with this kind of thing very well either, particularly as it predates HTML5 AFAIK].)

JakeQZ:
> I'd strongly prefer we get the new parser class in as a separate PR first (with good and complete tests).

We have some tests already in `AbstractHtmlProcessorTest` for adding missing `DOCTYPE`, `<html>`. `<head>` and `<body>` though I agree we also need to test the new class for this (and anything else it does that is testable).  To avoid duplication, I wonder if those tests (or the common part of them) could be moved to a `trait` or suchlike?

I also agree with breaking this down into smaller PRs, and the new class seems a good place to start.

src/HtmlProcessor/HtmlParser.php
```diff
@@ -0,0 +1,208 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\HtmlProcessor;
```
JakeQZ: I think this should belong in the `Pelago\Emogrifier\Utilities` namespace (and probably marked `@internal`).

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
JakeQZ: This may be partly due to `masterminds/html5` [incorrectly using `PHP_EOL`](https://github.com/Masterminds/html5-php/blob/2.7.0/src/HTML5/Serializer/OutputRules.php#L456).  The line ending used should not depend on whether it happens to be running on Windows or not.

Fortunately this is only used in two places: after the `DOCTYPE`, and at the end.  Until corrected in `masterminds/html5`, we would need to replace `"\r\n"` with `"\n"` in those two places - although it probably doesn't matter, it will make testing easier.

There are some newline differences also, and a formatting-agnostic comparison will help here as elsewhere.  I found that `masterminds/html5` would always produce a `DOMDocument` with one more newline between `</body>` and `</html>` that was originally input (so an input which produces the same output is impossible).

I'm not sure we need to provide a `getHtml5` method (can the `HTML5` object remain internal?) or change this test other than dealing with the newlines issue.

tests/Unit/HtmlProcessor/CssToAttributeConverterTest.php
JakeQZ: I’ve looked into this a bit more.

[Earlier versions of the HTML5 spec](https://dev.w3.org/html5/html-author/#empty-attr) permitted the empty attribute syntax only for Boolean attributes.  [Later versions](https://stackoverflow.com/questions/30748847/is-an-empty-class-attribute-valid-html#comment-49561145) still don’t explicitly state that it is allowed for non-Boolean attributes, but they no longer say it is not (the specific clause, “This syntax is permitted only for boolean attributes,” has been removed).

Given that email clients are some way behind in their support for the specs (if they weren’t there’d be no need for Emogrifier), and many are probably still parsing HTML as HTML 4.01 or earlier, I would be reticent about outputting non-Boolean attributes with the empty attribute syntax, as it may cause issues on some clients.

`Masterminds/html5-php` actually has [rules](https://github.com/Masterminds/html5-php/blob/2.7.0/src/HTML5/Serializer/OutputRules.php#L70) defining non-Boolean attributes to be output with `=""` but these do not kick in when the `disable_html_ns` option is set.  But if that option is not set, an `xmlns` attribute is set on the root `html` element and our XPath expressions fail to match elements by type.

It would be possible to extend `OutputRules` to achieve the desired behaviour (which could also help in other cases, e.g. preventing `"\r"` in the output on Windows); to do so would also require extending `HTML5` to replace the `save` method to instantiate an extended `OutputRules`.

I note that the test that has picked this up is unrelated, which suggests some more tests in this area may be pertinent – although in some ways they would really be testing the behaviour of a third-party component (`ext-dom` or `Masterminds\html5-ptp`), albeit indirectly.

src/HtmlProcessor/HtmlParser.php
bytestream: This class has been released in beta and will soon be released in stable so I'll correct if anything comes up.

> It would be good to put all this in the PHPDoc for the class.

If you have a suggestion of what you'd like to see I'm happy to add it.

> Adding missing DOCTYPE, <html>, <head> and <body> elements has for a long time been part of Emogrifier.

This would suggest you have some code elsewhere in the lib already that achives the same as this class?

JakeQZ: > > It would be good to put all this in the PHPDoc for the class.
> 
> If you have a suggestion of what you'd like to see I'm happy to add it.

Probably to clarify which missing root elements are added (is it `<html>`, `<head>` and `<body>` though not `<!DOCTYPE>`?), and that this is because `masterminds/html5` treats the input HTML as a document fragment rather than complete document (even if it has a DOCTYPE), so these elements do not get created if their tags are not in the provided HTML
.
> > Adding missing DOCTYPE, ,  and  elements has for a long time been part of Emogrifier.
> 
> This would suggest you have some code elsewhere in the lib already that achives the same as this class?

`AbstractHtmlProcessor::ensureDocumentType` and `AbstractHtmlProcessor::addContentTypeMetaTag` ensure a DOCTYPE and `Content-Type` are present.  The creation of the `<html>`, `<head>` and `<body>` elements (whether or not explicitly provided in the input HTML) is done by `DOMDocument` itself, so there is no specific code in Emogrifier to do this, though it is tested that these elements are present in the output HTML.

bytestream:
@oliverklee @JakeQZ what needs to happen to get this moving again?

oliverklee:
> what needs to happen to get this moving again?

@bytestream I propose adding the new class (without using it) and unit tests for it in a separate PR, and then (after that PR is merged) we can make this PR about using the new class.

bytestream:
I've sent it upstream. Lets see if they accept it there, otherwise I'll send another PR here.

JakeQZ:
Hi @bytestream, thanks for following-up on this.

I’ve committed a couple of changes to address code changes that were required to the tests as part of this PR.  The remaining issue pertains to the empty attribute syntax, for which I have created #867 for discussion of the best way forward.

I’m also wondering about testing both the HTML5 code paths and HTML4 code paths (i.e. when the DOCTYPE is explicitly HTML4 or earlier) with the same set of tests (that we already have), as these will diverge with the introduction of `Masterminds/html5-php` and maybe there’s a neat way of doing that…

bytestream:
I'll look at sending the new class your way in a separate PR tomorrow. Maintainers of `masterminds/html5-php` seem to be against the idea of merging my PR on the principle that they've long argued for many years that "the lib is only intended to parse *decent* html"

JakeQZ:
> "the lib is only intended to parse _decent_ html"

As opposed to [well-formed and perfectly valid HTML](https://html.spec.whatwg.org/multipage/syntax.html#optional-tags)?

JakeQZ:
> 
> 
> > "the lib is only intended to parse _decent_ html"
> 
> As opposed to [well-formed and perfectly valid HTML](https://html.spec.whatwg.org/multipage/syntax.html#optional-tags)?

Which does make me wonder - are all (or any) of the other tag ommission rules for conforming HTML respected?

JakeQZ:
> Which does make me wonder - are all (or any) of the other tag ommission rules for conforming HTML respected?

Initial testing shows that they actually are, and better than by `DOMDocument` - e.g. I've only found it failing to close a `<p>` element with a `<main>` element following, whereas `DOMDocument` fails to close it with any of the HTML5 sectioning elements following; and it also handles tag omission within `<dl>` better than `DOMDocument`.  So it's surprising then that they are not keen on handling `<html>`, `<head>` and `<body>` tag omission according to the spec, if what is provided to the parser is intended as a complete HTML document rather than a document fragment.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
bytestream: This is inconsistent behaviour on PHPs side. The bumped the minimum libxml version in PHP 7.4 but I can't see anything for PHP 7.3, but these kind of changes will be coming from the version of libxml bundled in PHP.

The issue is essentially that `masterminds/html5-php` copies over the `\n` character but `libxml` then adds it's own which results in unexpected duplication.

https://3v4l.org/vC5RX

Not much we can do about it other than change the test case.

bytestream:
I've submit another PR with changes to the existing parser rather than my own parser ...

When using https://github.com/Masterminds/html5-php/pull/183 there's only 4 broken tests and those are either whitespace, or the `alt=""` problem discussed above.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
JakeQZ: I had found that passing HTML through `Masterminds/html5-php` resulted in an additional newline between `</body>` and `</html>` however many were there originally, which may be related.

If you rebase with the latest `AbstractHtmlProcessorTest.php` this issue should be resolved through the tests being more forgiving of whitespace differences where they don't matter.

However, I note that I have not dealt with the issue of `Masterminds/html5-php` using `PHP_EOL` (it shouldn't be - the OS it's being run on should have no bearing on the result), so the test will still currently fail with a newline difference if you are running on Windows or Mac.

tests/Unit/HtmlProcessor/CssToAttributeConverterTest.php
bytestream: > Masterminds/html5-php actually has rules defining non-Boolean attributes to be output with ="" but these do not kick in when the disable_html_ns option is set.

How much did you dig into that? Enough to submit a PR to them? Doesn't seem too obvious to me how `disable_html_ns` affects the code from a quick glance.

JakeQZ: [`OutputRules::nonBooleanAttribute`](https://github.com/Masterminds/html5-php/blob/2.7.0/src/HTML5/Serializer/OutputRules.php#L371) decides this.

If there is no namespace, it always returns `false`, which means the attribute is considered Boolean and we get the empty attribute syntax.

However, if there is a namespace, as I'm sure you've found, our XPath expressions always fail because, e.g. `xmlns:p` != `p`...

bytestream:
I fear we're either going to stuck waiting for movement upstream, or going back and forth over some of these points. Namely, the issues surrounding boolean attributes and missing `<head>` / `<body>`.

Perhaps to speed this along, we could make it optional that this parser is used and as @JakeQZ mentioned, split the tests per parser so DOMDocument, `masterminds/html5-php` and any other that may be added in the future.

I don't think we should be adding my "normaliser" to this lib anymore which is why I've removed it from the PR. I'll just run on it my side...

oliverklee:
> Perhaps to speed this along, we could make it optional that this parser is used and as @JakeQZ mentioned, split the tests per parser so `DOMDocument`, `masterminds/html5-php` and any other that may be added in the future.

I'm perfectly fine with having things for testing in a branch or PR.

I'd like to avoid the complexity of having two DOM parsers and having to maintain support for both. So please let's use only one, and let's make sure all new code is covered well with tests.

bytestream:
Could this go the approach symfony took? Rather than adding html5 as a requirement, add it as a suggestion and only use it if the class exists. 

Your point about only using one parser I can appreciate but if the hope is that html5 parser will mimic all the same functionality as libxml then that's not feasible. html5 lib has stated explicitly that it's not a drop-in replacement and likely never will be

JakeQZ:
> Could this go the approach symfony took? Rather than adding html5 as a requirement, add it as a suggestion and only use it if the class exists.

That sounds good to me.  It would give me confidence that we will not break anything for those who are not using the `masterminds` library, and we could say something like "use at your own risk" - i.e. if you think you will get a benefit from it then go for it.  I think the edge case of people who are using the `masterminds` library for some other purpose but weren't intending it to be used with Emogrifier would be miniscule.

I'd still like the tests to be run for both code paths as much as possible, but would be happy for certain tests to be skipped where they would fail for the `masterminds` case.  I have already looked into the possibility of running a `PHPUnit` `TestCase` multiple times with different parameters, and it can be done.  I will try and dig out what I did to test that out; it's squirrelled away somewhere I think and hopefully not deleted; right now I can't remember where (it was literally just overriding a couple of methods in the `TestCase` class with about 3 lines of code each).

@oliverklee, WDYT?

I think I mentioned that there were some improvements in the pipeline for `DOMDocument` etc. with PHP 8.  Offhand, I don't remember what these might be or if even if they will actually be included...

@bytestream, thanks again for continuing to follow up on this issue, and putting the time and effort in.  Much appreciated!

bytestream:
Looks like the PHP 8 RFC was accepted and merged in not too long ago https://wiki.php.net/rfc/dom_living_standard_api

JakeQZ:
> Looks like the PHP 8 RFC was accepted and merged in not too long ago https://wiki.php.net/rfc/dom_living_standard_api

It mentions additional methods and properties, but nothing much about HTML5 - e.g. will it recognize certain elements like `wbr` and `track` as being void/self-closing (not that it matters that much in our case, we already have a workaround to deal with those issues).

JakeQZ:
>  I have already looked into the possibility of running a `PHPUnit` `TestCase` multiple times with different parameters, and it can be done. I will try and dig out what I did to test that out

I may have deleted it, but my starting point was [Stack Overflow](https://stackoverflow.com/questions/18253452/running-a-phpunit-testcase-multiple-times).  The `count` method also needs to be overridden as indicated in the comments, and there was some other issue that made it a not-quite-perfect solution, but it was getting there.

JakeQZ:
> 
> 
> > I have already looked into the possibility of running a `PHPUnit` `TestCase` multiple times with different parameters, and it can be done. I will try and dig out what I did to test that out
> 
> I may have deleted it, but my starting point was [Stack Overflow](https://stackoverflow.com/questions/18253452/running-a-phpunit-testcase-multiple-times). The `count` method also needs to be overridden as indicated in the comments, and there was some other issue (I forget precisely, but there was definitely some other problem) that made it a not-quite-perfect solution, but it was getting there.

It's possible of course that this could be tested via the GitHub actions matrix, but that would not help with testing on the local dev box.

oliverklee:
> Could this go the approach symfony took? Rather than adding html5 as a requirement, add it as a suggestion and only use it if the class exists.

Yes, that sounds feasible, and I would be okay with that.

However, I'd like to change this to a configuration option (or some other deliberate way to trigger this, e.g. by injecting the HTML parser). I wouldn't like Emogrifier automatically change its behavior in a project if some other dependency pulls in `masterminds/html5` during an update.

> I'd still like the tests to be run for both code paths as much as possible, but would be happy for certain tests to be skipped where they would fail for the masterminds case.

Even better: We can have different tests for both cases, or a flexible way to test this (which might be too complex - I haven't fully though this through).

JakeQZ:
> However, I'd like to change this to a configuration option (or some other deliberate way to trigger this, e.g. by injecting the HTML parser). I wouldn't like Emogrifier automatically change its behavior in a project if some other dependency pulls in `masterminds/html5` during an update.

I agree.  I think I alluded to that concern in my previous comment, and this 

> Even better: We can have different tests for both cases, or a flexible way to test this (which might be too complex - I haven't fully though this through).

I have now remembered what the issue with PHPUnit was that meant that the simple change I referenced above (to run the `TestCase` twice with different parameters) would not be quite adequate: the `dataProvider`s are executed beforehand and thus the provided data seems rather fixed and can't reflect the requirements for the particular run.  It might be worth opening this up to a wider audience, or even asking Sebastian himself if he has any suggestions.  Because I think what we basically want to do is run the `TestCase` multiple times (well, twice) with different parameters, ***and*** have those different parameters reflected in the provided data (i.e. we can skip some datasets that we know will fail for a particular configuration).

bytestream:
@JakeQZ it sounds like https://phpunit.readthedocs.io/en/9.2/annotations.html#requires is what you're after? If it needs to be done for individual cases within a `@dataProvider` then I would think that's as simple as not returning the case in the data provider function based on some condition.

> However, I'd like to change this to a configuration option (or some other deliberate way to trigger this, e.g. by injecting the HTML parser). I wouldn't like Emogrifier automatically change its behavior in a project if some other dependency pulls in masterminds/html5 during an update.

Just an `enableHtml5()` method? 

```php
public function enableHtml5()
{
    if (! class_exists(Html5::class)) { throw new Exception("Add masterminds/html5 as a project dependency"); }

    $this->html5 = true;
}
```

bytestream:
Any feedback on the above?

oliverklee:
ping @JakeQZ

JakeQZ:
> 
> 
> Any feedback on the above?



> 
> 
> ping @JakeQZ

Not sure what you are waiting for from me here.

I thought we had agreed for this to be implemented as an option that must be specifically enabled (i.e. not just a `class_exists()` test), and that the tests would be dualled up, so that the same set of tests would run for both options.

There are a few testing-related bits and bobs we still ought to implement, e.g. in the area of valueless attributes, but I don't think the lack of those is necessarily blocking this &ndash; it just could make it slightly more awkward because we would have to disable some tests for the Masterminds case.  But we would have to disable some tests for the Masterminds case anyway.

> @JakeQZ it sounds like https://phpunit.readthedocs.io/en/9.2/annotations.html#requires is what you're after? If it needs to be done for individual cases within a `@dataProvider` then I would think that's as simple as not returning the case in the data provider function based on some condition.

Oh, sorry, were you waiting for me to reply to this?  I kind of read it as a comment, and thought "that sounds good" - to myself :/

Um, that sounds good.  If it works, and solves the dual testing issue with the ability to vary the provided data based on with/without Masterminds (which we need because, AFAICS, we need to drop specific datasets depending on whether Masterminds is used), then, yes please, let's go for it.

I'm prepared to accept that certain tests will fail in the Masterminds case (you could also add tests that would fail conversely), as long as it's clear, and the issues involved are minor if not trivial or non-consequental (as I think most probably are).

Look forward to a PR... :)

JakeQZ:
> Look forward to a PR... :)

PS. You could probably split this into two PRs - the first one being for the dual tests, with the option to enable Masterminds which does nothing in that first PR...

bytestream:
Any comments on what I just pushed? This is the simplest way of doing it.

I've added `$html5` as an argument to `fromHtml()`. The reason for this is that a class level `enableHtml5` method does not make sense. The option is only relevant to `fromHtml()`. If you use `fromDomDocument` and `enableHtml5` then nothing is going to happen and I suspect that will confuse people.

I suspect what you really wanted to see is the below. But I'm struggling to see how this structure fits into the static method calls described above.

**AbstractHtmlProcessor:**
```php
public function __construct(?HTML5 $parser = null)
{
    $this->html5 = $parser;  
}
```

**AbstractTest:**
```php
public function fooTest(): void
{
    $this->instance->doSomething();
}
````

**DomDocumentTest:**
```php
public function setUp(): void
{
    $this->instance = new HtmlProcessor();
}
```

**Html5Test:**
```php
public function setUp(): void
{
    $this->instance = new HtmlProcessor(
        new HTML5([/* opts */])
    );
}
```

JakeQZ:
> Any comments on what I just pushed?

I like the solution to the testing issue, by moving it to the Composer / GitHub Action level to run the tests twice.  I hadn't thought of that, but it should make it much easier to do.

In terms of interface for usage, I was thinking that this would be a configration level option, and could be implemented similarly to [how it is done in PHPWord](https://phpword.readthedocs.io/en/latest/general.html#phpword-settings) for these sort of options, rather than an extra parameter.  @oliverklee, WDYT?

You are right that `HTML5` is not applicable to the `fromDomDocument` case.  This highlights another issue that I have just realized.  Since the `DOMDocument` may be passed between class instances (e.g. `CssInliner` then `CssPruner` and/or `CssToAttributeCoverter`), we would need some way for the `HTML5` object instance to be also passed through as well as the `DOMDocument`.  A neat solution, I think, would be to add an `AbstractHtmlProcessor::fromHtmlProcessor` method that takes care of copying all required object references across, and to recommend that in place of `fromDomDocument`, e.g. in the examples:

```php
$domDocument = CssInliner::fromHtml($html)->inlineCss($css)->getDomDocument();

HtmlPruner::fromDomDocument($domDocument)->removeElementsWithDisplayNone();
$html = CssToAttributeConverter::fromDomDocument($domDocument)
  ->convertCssToVisualAttributes()->render();
```

becomes

```php
$htmlProcessor = CssInliner::fromHtml($html)->inlineCss($css);

HtmlPruner::fromHtmlProcessor($htmlProcessor)->removeElementsWithDisplayNone();
$html = CssToAttributeConverter::fromHtmlProcessor($htmlProcessor)
  ->convertCssToVisualAttributes()->render();
```

Such a change could be implmented as a separate PR ahead of other changes.  When it comes to merging to main (hopefully soon!), it would be good to split this up in some reasonable logical way into several separate changesets, so it is more manageable and any issues can be more easily isolated.

bytestream:
>  A neat solution, I think, would be to add an AbstractHtmlProcessor::fromHtmlProcessor method that takes care of copying all required object references across, and to recommend that in place of fromDomDocument, e.g. in the examples:

This sounds better than

> I was thinking that this would be a configration level option, and could be implemented similarly to [how it is done in PHPWord](https://phpword.readthedocs.io/en/latest/general.html#phpword-settings) for these sort of options, rather than an extra parameter

because

> You are right that HTML5 is not applicable to the fromDomDocument case.

It would mean `fromHtmlProcessor` allows use of `masterminds/html5` and `fromDomDocument` / `fromHtml` stay as they are (`DOMDocument`).

JakeQZ:
> > A neat solution, I think, would be to add an `AbstractHtmlProcessor::fromHtmlProcessor` method that takes care of copying all required object references across, and to recommend that in place of `fromDomDocument`, e.g. in the examples:
> 
> This sounds better than
> 
> > I was thinking that this would be a configration level option, and could be implemented similarly to [how it is done in PHPWord](https://phpword.readthedocs.io/en/latest/general.html#phpword-settings) for these sort of options, rather than an extra parameter

I don't follow.  The two are independent.  There still needs to be a means of deciding whether to use `HTML5` in the first place (i.e. when `fromHtml` is initially invoked).

bytestream:
As I mentioned I think `setHtmlProcessor(HtmlProcessor::HTML5);` would be confusing when used with `static::fromDomDocument` ?

JakeQZ:
> 
> 
> As I mentioned I think `setHtmlProcessor(HtmlProcessor::HTML5);` would be confusing when used with `static::fromDomDocument` ?

I see your point, though if only using `fromDomDocument` you would not  call `setHtmlProcessor` as it would have no effect.

I can also see that having an optional Boolean parameter to `fromHtml` would allow greater flexibility more easily (should one want to sometimes but not always use `HTML5`).  Is this the only (public) method that would gain a parameter?

bytestream:
> Is this the only (public) method that would gain a parameter?

I think so

JakeQZ:
> 
> 
> > Is this the only (public) method that would gain a parameter?
> 
> I think so

In that case I would be generally in favour of this approach.

I have thus far only glanced through the changes, and not used a fine-toothed comb.

If there is a way of breaking things down into separate PRs, that would be helpful, though my previous suggestion in this department would be no longer valid (if we are not going to have a global configuration option).

As mentioned, the only thing I think needs to be added is an `AbstractHtmlProcessor::fromHtmlProcessor` method, and this could be done as a separate PR first (including updates to the README to use this rather than `fromDomDocument` where applicable) - then with the 2nd PR this would be extended to copy the `HTML5` reference as well as the `DOMDocument` reference.

@oliverklee WDYT?

oliverklee:
I'm really sorry for being so late for the party. :-(

My thoughts on this:

- I really like the general approach.
- I don't think Emogrifier should be aware of environment variables. Instead, we should have some other way to configure our classes for one of the two modes - maybe via two different subclasss of each class, or maybe by injecting some HTML parsing strategy class. We also can have parametrized tests.
- I dislike having a boolean switch for switching between two (more or less equal) modes. Instead, maybe we can have two HTML parsers: one `NativeHtmlParser` (which would be a wrapper for the native HTML parsing, and which could be the default if no other parser is passed) and a `NewHtmlParser` (or some other name) so that we work with both parsers the same way.

What do you think?

bytestream:
I agree in priciniple but fear the changes you're suggesting require a lot of work.

> I don't think Emogrifier should be aware of environment variables

The env var is only really there to minimise the numbers of changes required to get the test suite to run on masterminds/html5

> I dislike having a boolean switch for switching between two (more or less equal) modes. Instead, maybe we can have two HTML parsers: one NativeHtmlParser (which would be a wrapper for the native HTML parsing, and which could be the default if no other parser is passed) and a NewHtmlParser (or some other name) so that we work with both parsers the same way.

I may be overcomplicating this but I don't understand the concept of an abstract parser in this case. The entire library is based around `DOMDocument` and `XPath` so any future parser would need support for these

oliverklee:

Issue #818, opened 2019-10-30
Move the CI checks from TravisCI to GitHub workflows


oliverklee:
… like PHPUnit: https://github.com/sebastianbergmann/phpunit/tree/master/.github/workflows

PR #832, opened 2020-01-03
[TASK] Move the PHP code sniffing to GitHub actions

oliverklee:
This should speed up the build and reduce our hassle with Travis CI.

Part of #818.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,36 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
+
+on:
+  - pull_request
+  - push
+
+name: CI
+
+jobs:
+  php-code-sniffer:
+    name: PHP Code Sniffer
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        php-version:
+          - 7.3
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: "Cache dependencies installed with composer"
+        uses: actions/cache@v1
+        with:
+          path: ~/.composer/cache
+          key: php${{ matrix.php-version }}-composer-${{ hashFiles('**/composer.json') }}
+          restore-keys: |
+            php${{ matrix.php-version }}-composer-
+
+      - name: Install Composer dependencies
+        run: composer install --no-progress
+
+      - name: Run PHP Code Sniffer
+        run: composer ci:php:sniff
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,15 +72,6 @@ script:
 - >
   if is_php_73; then
     echo;
-    echo "Running PHP_CodeSniffer";
-    composer ci:php:sniff;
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: "Cache dependencies installed with composer"
+        uses: actions/cache@v1
+        with:
+          path: ~/.composer/cache
+          key: php${{ matrix.php-version }}-composer-${{ hashFiles('**/composer.json') }}
```
JakeQZ: `matrix.php-version` seems to be resolving to an empty string, from the logs, e.g. https://github.com/MyIntervals/emogrifier/runs/372326499

Also, an [example I found](https://github.com/actions/cache/blob/master/examples.md#php---composer) is using `composer.lock` for the hash rather than `composer.json`.

oliverklee: This jobs does not use a PHP version matrix. So adding it PHP matrix version to the cache key indeed did not make any sense. (It will do, though, for jobs with a version matrix, e.g., the PHP linting or the unit tests.) Good catch!

oliverklee: We cannot hash `composer.lock` because we don't have one, and it gets generated only in the next step (`composer install`).

oliverklee: I have added the explicit PHP version using a matrix now.

oliverklee: I'll look into the `restore-keys` to find out what's the difference to `key`.

oliverklee: This seems to be the official way AFAICT.

oliverklee:
I have repushed.

JakeQZ:
> I have added the explicit PHP version using a matrix now.

> I have repushed.

I presume this was a GitHub config change, as I don't spot anything different except that the PHP version now appears in the job commands.

Thinking about this, I'm not sure that the cache key needs to reflect either PHP version or the contents of `composer.json` - isn't it just the packages that have previously been downloaded, which could be equally useful to multiple PHP versions or composer configurations?

However, I'd be happy to approve this as it is (as you say it seems to be the recommended way), and we can revisit it later if we discover it is posisble to make caching more effective... - let me know what you think.

oliverklee:
> I presume this was a GitHub config change, as I don't spot anything different except that the PHP version now appears in the job commands.

I added the PHP version matrix.

> Thinking about this, I'm not sure that the cache key needs to reflect either PHP version or the contents of `composer.json` - isn't it just the packages that have previously been downloaded, which could be equally useful to multiple PHP versions or composer configurations?

I honestly don't know. I've read that some people had problems with the composer cache.

> However, I'd be happy to approve this as it is (as you say it seems to be the recommended way), and we can revisit it later if we discover it is posisble to make caching more effective... - let me know what you think.

Yes, please.

PR #837, opened 2020-01-23
[CLEANUP] Provide static analysis commands via GitHub actions matrix

oliverklee:
This way, we can avoid code duplication in the GitHub actions
configuration file several commands that use the same PH Pversion,
run `composer install`, cache the packages and execute a command.

More static analysis commands will be added later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,13 +7,15 @@ on:
 name: CI
 
 jobs:
-  php-code-sniffer:
-    name: PHP Code Sniffer
+  static-analysis:
+    name: Static Analysis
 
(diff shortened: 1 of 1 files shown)


oliverklee:
> I don't know whether it would be possible or useful for "Run Command" to include the actual command 

Yes, that would be possible.

The matrix items are truncated in the actions page, though: We need to make sure the relevant parts are at the beginning so we know which job is what.

JakeQZ:
> The matrix items are truncated in the actions page, though: We need to make sure the relevant parts are at the beginning so we know which job is what.

Ah, yes, I see now.

PR #838, opened 2020-01-25
[TASK] Move more static analysis from TravisCI to GitHub actions

oliverklee:
Part of #818

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,10 @@ jobs:
     strategy:
       matrix:
         command:
-          - php:sniff
+          - php:fixer
           - php:md
+          - php:psalm
+          - php:sniff
         php-version:
           - 7.3
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,11 +32,6 @@ install:
 
 script:
 - >
-  is_php_73() {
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -15,8 +15,10 @@ jobs:
     strategy:
       matrix:
         command:
-          - php:sniff
```
JakeQZ: In `composer.json` we have the order
- `lint`
- `sniff`
- `fixer`
- `md`
- `psalm`

I thought this was already the most logical order?  Though it may be a moot point as the actions run in parallel.

oliverklee: The order in the `composer.json` was somewhat arbitrary/random/organically-grown. The order in the `.travis.yml` followed a rough "more functional checks before more cosmetic checks, and faster checks before slower checks" order. For GitHub actions (which indeed get executed almost in parallel), I have used an alphabetical order.

JakeQZ: > I have used an alphabetical order.

OK, fair enough.

oliverklee:
> Does this close #818, or do you also plan to move the unit tests over?

I plan to move everything over and drop Travis CI here.

PR #844, opened 2020-02-24
[BUGFIX] Always ensure there is a head element present

WebsiteDeveloper:
Always ensure there is a head element present because the CSSInliner Expects it to be there.

### DIFF
--- a/src/HtmlProcessor/AbstractHtmlProcessor.php
+++ b/src/HtmlProcessor/AbstractHtmlProcessor.php
@@ -194,6 +194,7 @@ abstract class AbstractHtmlProcessor
     private function createUnifiedDomDocument(string $html)
     {
         $this->createRawDomDocument($html);
+        $this->ensureExistenceOfHeadElement();
         $this->ensureExistenceOfBodyElement();
     }
 
@@ -301,6 +302,26 @@ abstract class AbstractHtmlProcessor
             '$0/',
             $html
         );
+    }
+
+    /**
+     * Checks that $this->domDocument has a HEAD element and adds it if it is missing.
+     *
+     * @return void
+     *
+     * @throws \UnexpectedValueException
+     */
+    private function ensureExistenceOfHeadElement()
+    {
+        if ($this->getDomDocument()->getElementsByTagName('head')->item(0) !== null) {
+            return;
+        }
+
+        $htmlElement = $this->getDomDocument()->getElementsByTagName('html')->item(0);
+        if ($htmlElement === null) {
+            throw new \UnexpectedValueException('There is no HTML element although there should be one.', 1569930853);
+        }
+        $htmlElement->appendChild($this->getDomDocument()->createElement('head'));
     }
 
     /**
--- a/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
+++ b/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Pelago\Emogrifier\Tests\Unit\HtmlProcessor;
 
+use Pelago\Emogrifier\CssInliner;
 use Pelago\Emogrifier\HtmlProcessor\AbstractHtmlProcessor;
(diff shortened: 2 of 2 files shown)


oliverklee:
Hi @WebsiteDeveloper, thanks for the bug fix! :heart: Could you please add a unit test (or multiple tests) that fails without the fix and that passes with it? Thanks!

WebsiteDeveloper:
@oliverklee i added a test with data i had on hand
Maybe the test could be simplified i bit more i'm not sure

oliverklee:
@WebsiteDeveloper Thanks! Could you also please fix the code style warnings? You can find the results of the CI run in the tab "Checks" of this PR. Thanks!

tests/Unit/CssInlinerTest.php
```diff
@@ -2982,6 +2982,64 @@ public function inlineCssNotInDebugModeKeepsInvalidOrUnrecognizedSelectorsInMedi
         self::assertContainsCss($css, $subject->render());
     }
 
+    /**
+     * @test
+     */
+    public function htmlWithoutHeadOrBodyElementsAndStylesCopiedToStyleBlocksIsProperlyInlined()
```
oliverklee: This test should go into the tests for the base class: `AbstractHtmlProcessorTest`.

tests/Unit/CssInlinerTest.php
```diff
@@ -2982,6 +2982,64 @@ public function inlineCssNotInDebugModeKeepsInvalidOrUnrecognizedSelectorsInMedi
...
+        </p>
+        <p>
+            <a href="/fileadmin/_processed_/c/a/test.jpg" class="lightbox" rel="lightbox[922]" data-fancybox="images"><img height="198" src="fileadmin/_processed_/c/a/test.jpg" width="300" /></a>
+            &nbsp;
+            <a href="/fileadmin/_processed_/c/2/test.jpg" class="lightbox" rel="lightbox[922]" data-fancybox="images"><img height="198" src="fileadmin/_processed_/c/2/test.jpg" width="300" /></a>
+            &nbsp;
+            <a href="/fileadmin/_processed_/3/0/test.jpg" class="lightbox" rel="lightbox[922]" data-fancybox="images"><img height="198" src="fileadmin/_processed_/3/0/test.jpg" width="300" /></a>
+        </p>
+    </div>
+</div>
```
oliverklee: Could you please reduce the HTML to the bare minimum needed for this test to be meaningful? This will help readers see what's relevant (and also make the test shorter).

WebsiteDeveloper: I did reduce both the css and the html to a bare minimum now

tests/Unit/CssInlinerTest.php
WebsiteDeveloper: Done

oliverklee:
I fully agree (to all @JakeQZ's points).

(`AbstractHtmlProcessor` will (and cannot) be used stand-alone, though. The "bare-bone" usage would be the `HtmlNormalizer`.)

garvinhicking:
I ran into this bug too, I've fixed it simply by replacing the code referenced in https://github.com/MyIntervals/emogrifier/pull/844#pullrequestreview-363553242 from this:

```
$hasHeadTag = \stripos($html, '<head') !== false;
```

to this:

```
$hasHeadTag = \preg_match('@<head\s*/?>@i', $html);
```

I left the $hasHtmlTag because there's no valid HTML5 element that starts with `<html`.

HTH - I've not created a pull request because you seem to be on top of it, and just wanted to share my patch.

JakeQZ:
> ```
> $hasHeadTag = \preg_match('@<head\s*/?>@i', $html);
> ```

As the `head` element may have attributes, I would use something like `<head(?=[\s/>])[^>]*+>` (disregarding that attribute values in quotes _could_ have an unescaped `>`).

This would also, however, match a `head` tag with an invalid self-closing slash (e.g. `<head/>`) which I'm not sure is what we want to do.  Perhaps an exception should be thrown if such a case is presented.  And for this instance, `<head[\s>]` would be sufficient for the regular expression.

I may have a bit of time to pick this up in the next few days, unless @WebsiteDeveloper you have something in the pipeline?

PR #845, opened 2020-02-26
Fix mapping width/height when decimal is used

uzegonemad:
This fixes an issue where if you have a width like 33.33%, the mapWidthOrHeightProperty will not correctly detect it and the style cannot be converted into an attribute.


oliverklee:
Hi @uzegonemad, thanks for the bug fix. I appreciate that you've already provided the regression test. :+1:

Could you please fix the style checker warning, add a changelog line (newest on top) and repush? Thanks! :heart:

uzegonemad:
@oliverklee Sure - CI checks are running now.

PR #846, opened 2020-03-27
[TASK] Move the unit tests from Travis CI to GitHub actions

oliverklee:
Removing the Travis CI integration will follow in a separate commit.

Part of #818

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,3 +97,51 @@ jobs:
 
       - name: Run Command
         run: composer ci:php:${{ matrix.command }}
+
+  unit-tests:
+    name: Unit tests
+
+    runs-on: ubuntu-latest
+
+    needs: [php-lint, composer-validate]
+
+    strategy:
+      matrix:
+        php-version:
+          - 7.0
+          - 7.1
+          - 7.2
+          - 7.3
+          - 7.4
+        dependencies:
+          - lowest
+          - highest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v1
+        with:
+          php-version: ${{ matrix.php-version }}
+          ini-values: xdebug.max_nesting_level=500
+
+      - name: Cache dependencies installed with composer
+        uses: actions/cache@v1
+        with:
+          path: ~/.composer/cache
+          key: php${{ matrix.php-version }}-${{ matrix.dependencies }}-composer-${{ hashFiles('**/composer.json') }}
+          restore-keys: |
+            php${{ matrix.php-version }}-${{ matrix.dependencies }}-composer-
+
+      - name: Install Composer dependencies
+        run: |
+          if [[ "${{ matrix.dependencies }}" == 'lowest' ]]; then
+            DEPENDENCIES='--prefer-lowest';
+          fi;
+          composer update --no-progress "${DEPENDENCIES}";
+          composer show;
+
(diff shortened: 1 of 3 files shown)


oliverklee:
Currently, the Travis build status is not automatically visible in this PR.
Here's the direct link to the build:
https://travis-ci.org/github/MyIntervals/emogrifier/builds/667714204

.github/workflows/ci.yml
```diff
@@ -97,3 +97,51 @@ jobs:
...
+      - name: Cache dependencies installed with composer
+        uses: actions/cache@v1
+        with:
+          path: ~/.composer/cache
+          key: php${{ matrix.php-version }}-${{ matrix.dependencies }}-composer-${{ hashFiles('**/composer.json') }}
+          restore-keys: |
+            php${{ matrix.php-version }}-${{ matrix.dependencies }}-composer-
+
+      - name: Install Composer dependencies
+        run: |
```
JakeQZ: Might a `composer install` step be needed first, before `composer update`?

.github/workflows/ci.yml
```diff
@@ -97,3 +97,51 @@ jobs:
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v1
+        with:
+          php-version: ${{ matrix.php-version }}
+          ini-values: xdebug.max_nesting_level=500
```
JakeQZ: Do we need this (we did not previously change the setting)?

.github/workflows/ci.yml
```diff
@@ -97,3 +97,51 @@ jobs:
...
+          path: ~/.composer/cache
+          key: php${{ matrix.php-version }}-${{ matrix.dependencies }}-composer-${{ hashFiles('**/composer.json') }}
+          restore-keys: |
+            php${{ matrix.php-version }}-${{ matrix.dependencies }}-composer-
+
+      - name: Install Composer dependencies
+        run: |
+          if [[ "${{ matrix.dependencies }}" == 'lowest' ]]; then
+            DEPENDENCIES='--prefer-lowest';
+          fi;
```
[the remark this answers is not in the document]
oliverklee: Yes, that would be cleaner. I'll update the PR accordingly.

.github/workflows/ci.yml
```diff
@@ -97,3 +97,51 @@ jobs:
...
+          composer update --no-progress "${DEPENDENCIES}";
```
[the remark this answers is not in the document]
oliverklee: Exactly. (This kind of problem will go away once we've moved the tools from Composer dependencies to Phive.)

.github/workflows/ci.yml
oliverklee: No, this is not required (and would slow down the process as we'd then potentially download two versions of some packages). `update` always includes the `install` (and for calculating the updates, Composer does not need any packages installed.)

.github/workflows/ci.yml
oliverklee: Yes. Without this setting, we'd get a "maximum nesting level reached" error in `CssInlinerTest::inlineCssAppliesMoreSpecificCssSelectorToMatchingElements`:
https://github.com/MyIntervals/emogrifier/runs/539483776?check_suite_focus=true

.github/workflows/ci.yml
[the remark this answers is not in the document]
oliverklee: Done.

PR #847, opened 2020-03-27
[TASK] Drop support for obsolete Symfony versions

oliverklee:
Now only Symfony versions are supported that are either actively
maintained or at least receive security fixes.

This reduces the risk of being impacted by bugs that are long fixed.

For a list of supported Symfony versions, see:
https://symfony.com/releases


oliverklee:
This is blocked by #846 and #848. I'll rebase after both PRs are merged.

PR #848, opened 2020-03-27
[BUGFIX] Stop caching the vendor/ directory on Travis CI

oliverklee:
Having this directory cached between builds can result in incorrect
versions of dependencies getting installed.

For a good build performance, caching the Composer cache directory
should suffice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ php:
 
 cache:
   directories:
-  - vendor
   - $HOME/.composer/cache
 
 env:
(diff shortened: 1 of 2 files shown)


oliverklee:
Currently, the Travis build status is not automatically visible in this PR.
Here's the direct link to the build:
https://travis-ci.org/github/MyIntervals/emogrifier/builds/667723113

oliverklee:
Closing. After #846, this will not be needed anymore.

Issue #822, opened 2019-12-05
Move our CI tools from Composer dev dependencies to PHARs via Phive


oliverklee:
The way https://github.com/sebastianbergmann/phpunit/ does it.

This will avoid problems with conflicting dependencies between the tools (or with the non-dev dependencies of Emogrifier itself).

oliverklee:
Comment by @burzum moved from https://github.com/MyIntervals/emogrifier/issues/819#issuecomment-562283209:

@oliverklee how do you want to do the setup with Phive? Add the files in their according version to each branch / version? Or set it up when running Travis? I'm working on getting Phive set up only on Travis right now.

I've made some progress but it's trial and error with Travis... Is there no way to change the config online like we can do with Scrutinize and run builds using that config without committing the config file?

oliverklee:
> how do you want to do the setup with Phive? Add the files in their according version to each branch / version? Or set it up when running Travis?

@burzum The way https://github.com/sebastianbergmann/phpunit/ does it, i.e., the usage via Phive should replace the current dev dependencies and should be available both for Travis CI as well as local development (via Composer scripts).

oliverklee:
Note to myself: We can autoload from PHARs (e.g., so that Psalm can find PHPUnit's classes) this way: https://github.com/composer/composer/issues/5036#issuecomment-467761062

oliverklee:
And this is how to register an autoloader with Psalm:
https://psalm.dev/docs/running_psalm/configuration/#autoloader

PR #851, opened 2020-03-28
[TASK] Move php-cs-fixer to phive

oliverklee:
This reduces conflicts between our dependencies.

Part of #822


oliverklee:
> Should https://github.com/MyIntervals/emogrifier/blob/master/.github/CONTRIBUTING.md#install-the-development-dependencies also be updated?

Yes, when we move the corresponding tools (i.e., not in this PR).

PR #852, opened 2020-03-28
[TASK] Move PHPMD to Phive

oliverklee:
Also regenerate `phive.xml` to use the default formatting.

Part of #822


phive.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phive xmlns="https://phar.io/phive">
-    <phar name="php-cs-fixer" version="^2.16" installed="2.16.1" location="./tools/php-cs-fixer" copy="true"/>
+  <phar name="php-cs-fixer" version="^2.16" installed="2.16.1" location="./tools/php-cs-fixer" copy="true"/>
+  <phar name="phpmd" version="^2.7" installed="2.7.0" location="./tools/phpmd" copy="true"/>
```
JakeQZ: I've just noticed the `installed` attribute.  It looks to me like something specific to a local copy and set up, and thus should be excluded from version control?

PR #853, opened 2020-03-29
[BUGFIX] Add `<head>` element despite `<header>`

JakeQZ:
Also test around the issue.

Resolves #844.

### DIFF
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -2253,6 +2253,20 @@ class CssInlinerTest extends TestCase
     }
 
     /**
+     * #844
+     *
+     * @test
+     */
+    public function inlineCssKeepsRuleWithPseudoComponentInMatchingSelectorForHtmlWithHeader()
+    {
+        $subject = $this->buildDebugSubject('<html><header><a>foo</a></header></html>');
+
+        $subject->inlineCss('a:hover { color: green; }');
+
+        self::assertContainsCss('<style type="text/css">a:hover { color: green; }</style>', $subject->render());
+    }
+
+    /**
      * @return string[][]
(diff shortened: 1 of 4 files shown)


tests/Unit/CssInlinerTest.php
```diff
@@ -2252,6 +2252,20 @@ public function inlineCssAppliesCssToMatchingElementsAndKeepsRuleWithPseudoCompo
         self::assertContainsCss('<style type="text/css">a:hover { color: green; }</style>', $result);
     }
 
+    /**
+     * #844
```
oliverklee: To keep the code succinct, please let's not have issue numbers in the code. Having them in the changelog and the git history should suffice.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -175,6 +175,9 @@ public function contentWithoutHeadTagDataProvider(): array
             'doctype only' => ['<!DOCTYPE html>'],
             'body content only' => ['<p>Hello</p>'],
             'BODY element' => ['<body></body>'],
+            // #844
```
oliverklee: Please let's remove the issue number here as well.

JakeQZ:
> To keep the code succinct, please let's not have issue numbers in the code. Having them in the changelog and the git history should suffice.

OK, I've removed these.

Perhaps I am abusing the `needs unit tests` label?  I added it because the tests added went slightly beyond the specific issue.  But perhaps it is intended as a temporary label like `waiting for feedback`?

oliverklee:
@JakeQZ I intended the `unit tests needed`label to mean "there still are some tests missing".

PR #855, opened 2020-03-31
[CLEANUP] Whitespace-agnostic HTML test comparison

JakeQZ:
In tests for `AbstractHtmlProcessor`, ignore whitespace differences in the outer
`<html>` element.

This will help with #831.

Also added test that the `<body>` content passed to `fromHtml()` is preserved
and returned by `render()`.

### DIFF
--- a/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
+++ b/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
@@ -757,8 +757,8 @@ class AbstractHtmlProcessorTest extends TestCase
      */
     private static function assertEqualsHtml(string $expected, string $actual, string $message = '')
     {
-        $normalizedExpected = self::normalizeHtml($expected);
-        $normalizedActual = self::normalizeHtml($actual);
+        $normalizedExpected = self::normalizeHtmlElement($expected);
+        $normalizedActual = self::normalizeHtmlElement($actual);
 
         self::assertSame($normalizedExpected, $normalizedActual, $message);
     }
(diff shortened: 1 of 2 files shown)


tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -732,4 +746,42 @@ public function getDomDocumentVoidElementNotHasChildNodes(string $htmlWithNonXml
...
+
+    /**
+     * Normalizes whitespace in the HTML element itself (but not its descendants) and after its closing tag, with a
+     * single newline inserted or replacing whitespace at positions where whitespace may occur but is superfluous.
+     *
+     * @param string $html
+     *
+     * @return string
+     */
+    private static function normalizeHtml(string $html)
```
oliverklee, line 774: This method can have a `string` return type annotation.

oliverklee, line 774: I propose renaming this to `normalizeHtmlElement` to make it clear it's not about the string `$html`, but about the specific element. WDYT?

JakeQZ: Good catch.  I've added #857 to see if missing return type anntations could be automatically caught.

JakeQZ: OK.  Have changed thus.  The original name was because I wasn't sure whether some whitespace within the `<head>` element might also need normalizing, but testing against changes in #831 I found this wasn't required.

oliverklee: I think the callers still are using the old name.

JakeQZ: > I think the callers still are using the old name.

Sorry, I'd fixed this in a local copy where I ran the tests, but forgot to merge with the `git` folder from which I did the `push`.

PR #861, opened 2020-04-09
[TASK] Upgrade to V2 of the PHP setup GitHub action

oliverklee:


oliverklee:
We might need to fix #862 first. Let's see.

oliverklee:
Rebased and repushed.

oliverklee:
Yay, everything is green now! Merging.

PR #865, opened 2020-04-20
[BUGFIX] Enquote development tool paths for Windows

JakeQZ:
If the path-to-executable part of a command contains Unix-style path separators
(which are supported on Windows), it must be enclosed in double quotation marks
to work on Windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 - Make sure to use the Composer-installed development tools
-  ([#862](https://github.com/MyIntervals/emogrifier/pull/862))
+  ([#862](https://github.com/MyIntervals/emogrifier/pull/862),
+  [#865](https://github.com/MyIntervals/emogrifier/pull/865))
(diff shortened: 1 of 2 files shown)


oliverklee:
I'm not sure that "enquote" is a word. :-) Should it be "encode" or "quote" (or did you mash both together on purpose)?

JakeQZ:
> 
> 
> I'm not sure that "enquote" is a word. :-) Should it be "encode" or "quote" (or did you mash both together on purpose)?

It's [listed](http://www.engyes.com/en/dic-content/enquote) [in](https://en.wiktionary.org/wiki/enquote) [some](https://glosbe.com/en/en/enquote) [dictionaries](https://www.wordsense.eu/enquote/), though perhaps not the [most](https://www.collinsdictionary.com/spellcheck/english?q=enquote) [authoritative](https://chambers.co.uk/search/?query=enquote&title=21st).

The change could be committed with a different message if you prefer?

oliverklee:
> It's listed in some dictionaries, though perhaps not the most authoritative.

Ah, thanks, I've just learned something new! :-)

Issue #869, opened 2020-05-22
[Parse CSS Rules] @font-face should be uninlineable


aarongerig:
The `parseCssRules` method should not mark `@font-face` as inlineable. Otherwise, the font declaration simply gets removed since there is no DOM element found for the `@font-face` selector.

`@font-face`should be put into `<head><style type="text/css">...</style></head>` so fonts can be loaded and rendered.

WDYT?

JakeQZ:
I agree, we should support this, thanks @aarongerig for raising it.

It could probably be handled similarly to how `@import` and `@charset` rules are - [these are filtered out of the input CSS](https://github.com/MyIntervals/emogrifier/blob/v3.1.0/src/Emogrifier/CssInliner.php#L476) and [put back in at the end](https://github.com/MyIntervals/emogrifier/blob/v3.1.0/src/Emogrifier/CssInliner.php#L1065).

We are considering using a 3rd party library for parsing CSS (#544), which may help, but TBH I don't expect this to happen any time soon.

In the meantime, if you would like to submit a PR for this specific issue, it would be very much appreciated (and hopefully the links above will be pointers for where to start)... the code in this area looks a bit messy (I wrote some of it) so any other improvements you could suggest would also be welcome...

aarongerig:
@JakeQZ Thanks for the quick response! Sure, I'll give it a try with the PR. 👍🏻

PR #870, opened 2020-05-25
Extract font-face rules and inject into head <styles> element

aarongerig:
This is the follow up to issue https://github.com/MyIntervals/emogrifier/issues/869.

The code was tested on one of my current projects and there all `@font-face` rules were successfully extracted and injected into the `<styles>` element.

Let me know if I can do something else to get this into master.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -172,13 +172,13 @@ class CssInliner extends AbstractHtmlProcessor
         $cssWithoutComments = $this->removeCssComments($combinedCss);
         list($cssWithoutCommentsCharsetOrImport, $cssImportRules)
             = $this->extractImportAndCharsetRules($cssWithoutComments);
-        list($cssWithoutCommentsCharsetOrImportAndFontFaces, $cssFontFaces)
-            = $this->extractFontFaces($cssWithoutCommentsCharsetOrImport);
+        list($cssWithoutCommentsCharsetImportOrFontFace, $cssFontFaces)
+            = $this->extractFontFaceRules($cssWithoutCommentsCharsetOrImport);
 
         $uninlinableCss = $cssImportRules . $cssFontFaces;
 
         $excludedNodes = $this->getNodesToExclude();
-        $cssRules = $this->parseCssRules($cssWithoutCommentsCharsetOrImportAndFontFaces);
+        $cssRules = $this->parseCssRules($cssWithoutCommentsCharsetImportOrFontFace);
         $cssSelectorConverter = $this->getCssSelectorConverter();
         foreach ($cssRules['inlinable'] as $cssRule) {
             try {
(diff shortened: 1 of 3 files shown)


src/CssInliner.php
```diff
@@ -172,9 +172,13 @@ public function inlineCss(string $css = ''): self
...
         list($cssWithoutCommentsCharsetOrImport, $cssImportRules)
             = $this->extractImportAndCharsetRules($cssWithoutComments);
+        list($cssWithoutCommentsCharsetOrImportAndFontFaces, $cssFontFaces)
+            = $this->extractFontFaces($cssWithoutCommentsCharsetOrImport);
+
+        $uninlinableCss = $cssImportRules . $cssFontFaces;
 
         $excludedNodes = $this->getNodesToExclude();
-        $cssRules = $this->parseCssRules($cssWithoutCommentsCharsetOrImport);
+        $cssRules = $this->parseCssRules($cssWithoutCommentsCharsetOrImportAndFontFaces);
```
JakeQZ, line 181: I don't think the variable name should contain both "or" and "and" - perhaps `$cssWithoutCommentsCharsetImportOrFontFace`?

tests/Unit/CssInlinerTest.php
```diff
@@ -3206,6 +3206,121 @@ public function inlineCssNotCopiesInlinableRuleAfterImportRuleToStyleElement()
...
+                'before' => 'p:hover { color: green; }' . "\n",
+                '@font-face' => '@font-face { font-family: "Foo Sans"; src: url("/foo-sans.woff2") format("woff2"); }',
+                'after' => '',
+            ],
+            '@font-face preceded by matching @media rule' => [
+                'before' => '@media (max-width: 640px) { p { color: green; } }' . "\n",
+                '@font-face' => '@font-face { font-family: "Foo Sans"; src: url("/foo-sans.woff2") format("woff2"); }',
+                'after' => '',
+            ],
+        ];
```
JakeQZ, line 3272: Good set of tests 👍

src/CssInliner.php
aarongerig: I agree. I renamed it as well as the name of the extraction method, just to be consistent.

aarongerig:
Great – thanks for reviewing @JakeQZ !

tests/Unit/CssInlinerTest.php
```diff
@@ -3206,6 +3206,121 @@ public function inlineCssNotCopiesInlinableRuleAfterImportRuleToStyleElement()
...
+                '@font-face' => '@font-face { font-family: "Foo Sans"; src: url("/foo-sans.woff2") format("woff2"); }',
+                'after' => "\n" . '@media (max-width: 640px) { p { color: green; } }',
+            ],
+            '@font-face preceded by matching inlinable rule' => [
+                'before' => 'p { color: green; }' . "\n",
+                '@font-face' => '@font-face { font-family: "Foo Sans"; src: url("/foo-sans.woff2") format("woff2"); }',
+                'after' => '',
+            ],
+            '@font-face preceded by matching uninlinable rule' => [
+                'before' => 'p:hover { color: green; }' . "\n",
```
oliverklee, line 3263: As there are not double quotes in the string, we can use a single string encased in double quotes here, allowing us to avoid the concatenation altogether here. Ditto for the next data set.

aarongerig: Done.

aarongerig:
Awesome, thanks so much guys! 🙏🏻

PR #875, opened 2020-06-11
[BUGFIX] Avoid exception with :last-of-type etc.

### LINKED ISSUE #872: ExpressionErrorException: "*:last-of-type" is not implemented.
I'm trying to change over from 2.0 to 3.1

old code:
```
$emogrifier = new Emogrifier();
$emogrifier->setHtml($html);
$emogrifier->setCss($this->createCss());
return $emogrifier->emogrify();
```

new code attempted:

`return CssInliner::fromHtml($html)->inlineCss($this->createCss())->render();`

or

```
$domDocument = CssInliner::fromHtml($html)->inlineCss($this->createCss())->getDomDocument();
Emogrifier\HtmlProcessor\HtmlPruner::fromDomDocument($domDocument)->removeElementsWithDisplayNone();
$html = Emogrifier\HtmlProcessor\CssToAttributeConverter::fromDomDocument($domDocument)->convertCssToVisualAttributes()->render();
return $html;
```

Both give me

Symfony\Component\CssSelector\Exception\ExpressionErrorException: "*:last-of-type" is not implemented.

JakeQZ:
This simply broadens the range of exceptions thrown by `symfony/css-selector` (for unsupported selector expressions) which will be caught and discarded in non-debug mode.

It does not make anything magically work which didn't previously; however, it does allow the rest of the CSS to achieve full functionality, with only the affected CSS rule and selector suffering.

The tests and changes to the README clarify and correct-as-documented the current behaviour for these edge cases, which could be improved upon in future.

These are currently rarely-used CSS pseudo-classes; their level of support in mainstream browsers has only [recently reached 95% of audiences](https://caniuse.com/#search=last-of-type).  Their level of support in email clients is likely to be drasitcally lower per se for some time to come, but may be something Emogrifier can help with moving forwards, as their use becomes more widespread&hellip;

Closes #872.  New issue https://github.com/MyIntervals/emogrifier/issues/876 for follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#773](https://github.com/MyIntervals/emogrifier/pull/773))
 
 ### Fixed
+- Allow `:last-of-type` etc. without type, without causing exception
+  ([#875](https://github.com/MyIntervals/emogrifier/pull/875))
 - Make sure to use the Composer-installed development tools
   ([#862](https://github.com/MyIntervals/emogrifier/pull/862),
   [#865](https://github.com/MyIntervals/emogrifier/pull/865))
--- a/README.md
+++ b/README.md
@@ -308,33 +308,35 @@ Emogrifier currently supports the following
    * [first-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child)
    * [first-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:first-of-type)
      (with a type, e.g. `p:first-of-type` but not `*:first-of-type` which will
-     behave as `*:first-child`)
+     currently be treated as `*:not(*)`)
    * [last-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child)
    * [last-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:last-of-type)
-     (with a type)
+     (with a type &ndash; without a type, it will be treated as `:not(*)`)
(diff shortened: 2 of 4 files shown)


JakeQZ:
PS. In the README I am trying to clarify with `:not(*)` that the selector will not match anything but will not be discarded completely if in a comma-separated list.  E.g. those in-the-know will know that

```css
input::-webkit-input-placeholder, input:-ms-input-placeholder, 
  input:-moz-placeholder, input::-moz-placeholder 
{
  font-size: 1.25rem;
}
```

will not work.  Each specifically-targeted browser will throw away the entire rule because it doesn't recognize the component targeting a rival browser, and the above must be written as

```css
input::-webkit-input-placeholder {
  font-size: 1.25rem;
}
input:-ms-input-placeholder {
  font-size: 1.25rem;
}
input:-moz-placeholder {
  font-size: 1.25rem;
}
input::-moz-placeholder {
  font-size: 1.25rem;
}
```

Or, for a non-browser-specific example lifted from a popular WordPress theme which does not work as well as it could:

```css
.home.blog .site-header,
.home.page:not(.page-template-template-homepage) .site-header,
.home.post-type-archive-product .site-header {
    margin-bottom: 4.235801032em
}
```

IE8 (or any other parser which does not support `:not`) would, perfectly justifiably according to the spec, discard the entire rule, so it should be broken up as follows:

```css
.home.blog .site-header,
.home.post-type-archive-product .site-header {
    margin-bottom: 4.235801032em
}
.home.page:not(.page-template-template-homepage) .site-header {
    margin-bottom: 4.235801032em
}
```

I want to be clear that this splitting up of rules is not required, even if it is slightly at odds with the spec.

PR #877, opened 2020-06-11
[TASK] Prepare release of version 4.0.0

### LINKED ISSUE #874: Get version 4.0.0 out of the door
@zoliszabo @JakeQZ @jjriv I'd like to get version 4.0.0 out of the door soon.

For this, we'll should fix those bugs first:
- [x] #873 build failure on master
- [x] #872 catch exception thrown in the Symfony CSS component

Do we need to fix/change anything else before we can release version 4.0.0? (I'll move all other issues currently targeted at 4.0.0 to the next milestone.)

oliverklee:
Closes #874

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,6 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Configuration setting methods now all return `$this` to allow chaining
   ([#824](https://github.com/MyIntervals/emogrifier/pull/824),
   [#854](https://github.com/MyIntervals/emogrifier/pull/854))
-- Add support for PHP 7.4
-  ([#821](https://github.com/MyIntervals/emogrifier/pull/821))
 - Disable php-cs-fixer Yoda conditions
   ([#791](https://github.com/MyIntervals/emogrifier/issues/791),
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ This project adheres to [Semantic Versioning](https://semver.org/).
...
+
+### Deprecated
+- Support for PHP 7.1 will be removed in Emogrifier 6.0.
+
+### Removed
+
+### Fixed
+
+## 4.0.0
+
```
JakeQZ, line 20: Should there be an entry for 3.1.0 as well?  We've previously had entries for e.g. 2.1.0, 2.1.1 and 2.2.0.

oliverklee: 2.1.0 etc. ended up in the 3.0.0 changelog because we released the 2.1.0 versions etc. from master (which then became version 3.0.0), whereas 3.1.0 comes from a branch off the 3.0.0 tag because we already had some breaking changes in master at that time (if I remember correctly).

To make things more straightforward (so people can see the changes for 4.0.0 over 3.1.0), I'll update the changelog accordingly.

oliverklee: Done and pushed.

PR #880, opened 2020-06-12
[TASK] Drop support for PHP 7.0

oliverklee:


composer.json
```diff
@@ -37,7 +37,7 @@
         "source": "https://github.com/MyIntervals/emogrifier"
     },
     "require": {
-        "php": "~7.0 || ~7.1 || ~7.2 || ~7.3 || ~7.4",
+        "php": "~7.1 || ~7.2 || ~7.3 || ~7.4",
```
JakeQZ, line 40: Noting that [the `.0` point version suffix](https://getcomposer.org/doc/articles/versions.md#tilde-version-range-) should be restored at some point so the behaviour is actually as intended (though there will not be a PHP 7.5).

oliverklee: Agreed.

JakeQZ:
The CI tests seemed to get stuck waiting for Lint and unit tests for PHP 7.0, but that's now resolved - ?

composer.json
oliverklee: #881

oliverklee:
They still seem to get executed in #881. I have no idea why.

JakeQZ:
> They still seem to get executed in #881. I have no idea why.

Something being cached?  Or is the any relevant GitHub configuratoin for the project that is external to the files visible within the source tree?

JakeQZ:
Maybe this: https://github.community/t/github-actions-pull-requests-always-include-some-outdated-checks/16157

PR #885, opened 2020-06-13
[TASK] Stop checking tests/ with Psalm

### LINKED ISSUE #879: Stop checking tests/ with Psalm
This will pave the way towards [moving Psalm to Phive](#822) until we find a better solution.

oliverklee:
This will pave the way towards moving Psalm to Phive
until we find a better solution.

As the tests call the production code, Psalm would mark the called
code as unused. So we need to disable some checks for unused code.

Fixes #879


oliverklee:
This needs more changes for the Psalm configuration.

PR #888, opened 2020-06-14
[TASK] Upgrade to PHPUnit 7.5

oliverklee:
Also pin `vimeo/psalm` to avoid automatic updates that are now
possible due to the changed PHPUnit dependencies.


oliverklee:
This probably needs to #885 to be merged first. I'll rebase this then.

oliverklee:
It looks like we need to upgrade and configure Psalm (and fix some issues) first.

PR #894, opened 2020-06-14
[TASK] Move PHPUnit to Phive

oliverklee:
Part of #822


composer.json
```diff
@@ -74,8 +73,8 @@
         "ci:php:fixer": "\"./tools/php-cs-fixer\" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff-format=udiff config/ src/ tests/",
         "ci:php:md": "\"./tools/phpmd\" src text config/phpmd.xml",
         "ci:php:psalm": "\"vendor/bin/psalm\" --show-info=false",
-        "ci:tests:unit": "\"vendor/bin/phpunit\" tests/",
-        "ci:tests:sof": "\"vendor/bin/phpunit\" tests/ --stop-on-failure",
+        "ci:tests:unit": "\"./tools/phpunit\" tests/",
+        "ci:tests:sof": "\"./tools/phpunit\" tests/ --stop-on-failure",
```
JakeQZ, line 77: If the directory is now defined in the XML file, is it also required on the command line?

Also, does the 'stop-on-failure' variant still work as intended now the conifguration is in the XML file?

oliverklee: > If the directory is now defined in the XML file, is it also required on the command line?

Good catch! No, it's not required anymore. I've removed it now.

> Also, does the `stop-on-failure` variant still work as intended now the configuration is in the XML file?

I've just tested it: Yes, it does.

PR #895, opened 2020-06-14
[TASK] Upgrade Psalm

oliverklee:
Also regenerate the baseline and fix the issues that cannot be covered
by the baseline.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -464,10 +464,12 @@ class CssInliner extends AbstractHtmlProcessor
         }
 
         $css = '';
-        /** @var \DOMNode $styleNode */
         foreach ($styleNodes as $styleNode) {
             $css .= "\n\n" . $styleNode->nodeValue;
-            $styleNode->parentNode->removeChild($styleNode);
+            $parentNode = $styleNode->parentNode;
+            if ($parentNode instanceof \DOMNode) {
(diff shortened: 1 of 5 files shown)


oliverklee:
The PHP CS warning confuses me.

JakeQZ:
> 
> 
> The PHP CS warning confuses me.

My guess is that the sniffer does not understand `array<array-key, array<string, string>>` and sees it as simply `array`.  I think previously for these situations we have added a separate `@psalm-return` annotation with that syntax whilst retaining the `@return` annotation with the original syntax.

oliverklee:
Yay, green!

src/CssInliner.php
```diff
@@ -713,6 +713,8 @@ private function getCssSelectorPrecedence(string $selector): int
      *
      * @param string $css CSS with comments removed
      *
+     * @psalm-return array<array-key, array<string, string>> Array of string sub-arrays with the keys
```
JakeQZ, line 716: I think there should be no description here - the description belongs with `@return` - if the description were to change, these could be liable to become out-of-step.

oliverklee: Oh, you're right. I'll fix it.

oliverklee:
Fixed and repushed.

PR #899, opened 2020-06-14
[BUGFIX] Preserve all uninlinable CSS at-rules

### LINKED ISSUE #886: @-webkit-keyframes animationName not supported
created a branch with a test that fails:
https://github.com/janmyszkier/emogrifier/tree/jan/2_2_0_css_animation_test

done based on 2.2.0 (which seems to be currently used by magento/magento2 repository):
https://github.com/magento/magento2/blob/2.3/composer.lock#L3670

please provide a fix or doc information on how to avoid this error with emogrifier, so we can alter our code. (mentioning the animation in the readme would also be helpful if not going to be supported)

JakeQZ:
All at-rules that have no special handling by Emogrifier will now by copied to a
`<style>` element.  Previously, Emogrifier would attempt to process them as
regular CSS rules, causing an exception in `CssSelectorConverter::toXPath()`,
which in 'debug' mode would be propagated; in either case they would be
discarded.

The rationale is that if any such rules are in the supplied CSS, they are
desired in the resulting HTML.  Emogrifier can treat them as a black box it
knows nothing about, simply to be passed on.

For at-rules which are containers for regular CSS rules (e.g. `@supports`), the
ordering could be important.  For now, these rules will be placed at the start
of the `<style>` element, so it is possible that they will be overridden by
other uninlinable rules that are placed after them, but were originally placed
before.  However, there should be no regression because, by being placed first
(and in their own original order), they would still never override a rule in the
`<style>` element that they would not have in the original CSS.

The ordering improvement can be addressed separately, e.g. by handling container
at-rules, like `@supports`, more generally, in a similarly way to how `@media`
rules are currently processed.  (Addressing #544 first may also help.)

Fixes #886.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -46,6 +46,18 @@ class CssInliner extends AbstractHtmlProcessor
     private const CACHE_KEY_COMBINED_STYLES = 3;
 
     /**
+     * This regular expression pattern will match any uninlinable at-rule with nested statements, along with any
+     * whitespace immediately following.  Currently, any at-rule apart from `@media` is considered uninlinable.  The
+     * first capturing group matches the at sign and identifier (e.g. `@font-face`).  The second capturing group matches
+     * the nested statements along with their enclosing curly brackets (i.e. `{...}`), and via `(?2)` will match deeper
+     * nested blocks recursively.
+     *
+     * @var string
+     */
+    private const UNINLINABLE_AT_RULE_MATCHER
+        = '/(@(?!media\\b)[\\w\\-]++)[^\\{]*+(\\{[^\\{\\}]*+(?:(?2)[^\\{\\}]*+)*+\\})\\s*+/i';
+
+    /**
      * Regular expression component matching a static pseudo class in a selector, without the preceding ":",
      * for which the applicable elements can be determined (by converting the selector to an XPath expression).
      * (Contains alternation without a group and is intended to be placed within a capturing, non-capturing or lookahead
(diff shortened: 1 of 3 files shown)


src/CssInliner.php
```diff
@@ -523,38 +523,44 @@ private function extractImportAndCharsetRules(string $css): array
...
     {
         $possiblyModifiedCss = $css;
-        $fontFaces = '';
+        $atRules = '';
 
+        // `(\{[^\{\}]*+(?:(?2)[^\{\}]*+)*+\})` matches possibly nested `{...}` as 2nd capturing group, via `(?2)`
         while (
             \preg_match(
-                '/(@font-face[^}]++}\\s*+)/i',
+                '/(@(?!media\\b)[\\w\\-]++)[^\\{]*+(\\{[^\\{\\}]*+(?:(?2)[^\\{\\}]*+)*+\\})\\s*+/i',
```
oliverklee, line 546: Please let's move this to a private const so that the method has only the code, but no data.

src/CssInliner.php
```diff
@@ -523,38 +523,44 @@ private function extractImportAndCharsetRules(string $css): array
...
             )
         ) {
-            [$fullMatch, $atRuleAndFollowingWhitespace] = $matches;
+            [$fullMatch, $atRuleName] = $matches;
 
-            if (\stripos($fullMatch, 'font-family') !== false && \stripos($fullMatch, 'src') !== false) {
-                $fontFaces .= $atRuleAndFollowingWhitespace;
+            if (
+                \strcasecmp($atRuleName, '@font-face') !== 0
+                || \stripos($fullMatch, 'font-family') !== false && \stripos($fullMatch, 'src') !== false
```
oliverklee, line 555: Can we move this to a small, well-named method?

tests/Unit/CssInlinerTest.php
```diff
@@ -52,6 +52,100 @@ final class CssInlinerTest extends TestCase
         </html>
     ';
 
+    /**
+     * @var string[]
+     *
+     * selectoin of at-rules which have no special handling by `CssInliner` but should be passed through and placed in a
```
oliverklee, line 58: typo: selectoin -> selection

src/CssInliner.php
JakeQZ: Done.  This also allows the preceding comment to be moved to the DocBlock for the `const`.

src/CssInliner.php
JakeQZ: Done.

tests/Unit/CssInlinerTest.php
JakeQZ: Fixed.  Well spotted.

JakeQZ:
> 
> 
> Looks mostly very good. I've added some comments. This PR also needs a rebase.

I've made changes to address the comments and rebased.

I forgot to use the `-force` when rebasing then followed the incorrect suggestion to execute `git pull` - [déjà vu](https://github.com/MyIntervals/emogrifier/issues/373#issuecomment-536631092) - but all should hopefully be sorted now.

PR #906, opened 2020-06-19
[CLEANUP] Move AssertCss logic to Constraint classes

JakeQZ:
This is the pattern used by PHPUnit, and provides a couple of benefits:
- Constraints can be combined with, e.g., LogicalOr (will be useful for #544);
- The message provided upon a test failure will be clearer.

Also corrected the namespace for `AssertCssTest` for consistency with other
`TestCase` classes.

### DIFF
--- a/tests/Support/Constraint/CssConstraint.php
+++ b/tests/Support/Constraint/CssConstraint.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\Tests\Support\Constraint;
+
+/**
+ * Provides a method for use in constraints making assertions about CSS content where whitespace may vary.
+ *
+ * @author Jake Hotson <jake.github@qzdesign.co.uk>
+ */
+trait CssConstraint
+{
+    /**
+     * Processing of @media rules may involve removal of some unnecessary whitespace from the CSS placed in the <style>
+     * element added to the document, due to the way that certain parts are `trim`med.  Notably, whitespace either side
+     * of "{", "}" and "," or at the beginning of the CSS may be removed.
+     *
+     * This method helps takes care of that, by converting a search needle for an exact match into a regular expression
+     * that allows for such whitespace removal, so that the tests themselves do not need to be written less humanly
+     * readable and can use inputs containing extra whitespace.
+     *
+     * @param string $needle Needle that would be used with `assertContains` or `assertNotContains`.
+     *
+     * @return string Needle to use with `assertRegExp` or `assertNotRegExp` instead.
+     */
+    private static function getCssNeedleRegExp(string $needle): string
+    {
+        $needleMatcher = \preg_replace_callback(
+            '/\\s*+([{},])\\s*+|(^\\s++)|(>)\\s*+|(?:(?!\\s*+[{},]|^\\s)[^>])++/',
+            static function (array $matches): string {
+                if (isset($matches[1]) && $matches[1] !== '') {
+                    // matched possibly some whitespace, followed by "{", "}" or ",", then possibly more whitespace
+                    return '\\s*+' . \preg_quote($matches[1], '/') . '\\s*+';
+                }
+                if (isset($matches[2]) && $matches[2] !== '') {
+                    // matched whitespace at start
+                    return '\\s*+';
+                }
+                if (isset($matches[3]) && $matches[3] !== '') {
+                    // matched ">" (e.g. end of <style> tag) followed by possibly some whitespace
+                    return \preg_quote($matches[3], '/') . '\\s*+';
+                }
+                // matched any other sequence which could not overlap with the above
+                return \preg_quote($matches[0], '/');
+            },
+            $needle
+        );
+        return '/' . $needleMatcher . '/';
+    }
+}
--- a/tests/Support/Constraint/CssConstraint.php
+++ b/tests/Support/Constraint/CssConstraint.php
@@ -4,13 +4,35 @@ declare(strict_types=1);
 
 namespace Pelago\Emogrifier\Tests\Support\Constraint;
 
+use PHPUnit\Framework\Constraint\Constraint;
+
 /**
  * Provides a method for use in constraints making assertions about CSS content where whitespace may vary.
  *
+ * Due to {@link https://bugs.php.net/bug.php?id=75060 PHP 'bug' #75060}, traits cannot have constants, so, as a
+ * workaround, this is currently implemented as a base class (but ideally would be a `trait`).
+ *
  * @author Jake Hotson <jake.github@qzdesign.co.uk>
  */
-trait CssConstraint
+abstract class CssConstraint extends Constraint
 {
+    /**
+     * This regular expression pattern will match various parts of a string of CSS, and uses capturing groups to clarify
+     * what, actually, has been matched.
+     *
+     * @var string
+     */
+    private const CSS_REGEX_PATTERN = '/
+        \\s*+([{},])\\s*+               # `{`, `}` or `,` captured in group 1, with possible whitespace either side
+        |(^\\s++)                       # whitespace at the very start, captured in group 2
+        |(>)\\s*+                       # `>` (e.g. closing a `<style>` element opening tag) with optional whitespace
+                                        # following, captured in group 3
+        |(?:(?!\\s*+[{},]|^\\s)[^>])++  # Anything else is captured in group 4.  There is a reason for this, which is
+                                        # currently forgotten.  It has something to do with the way alternation works,
+                                        # in terms of matching the longest pattern, or, if equal, the first, and that
+                                        # the implementation in PHP does not quite behave as described on the tin.
+    /x';
+
     /**
      * Processing of @media rules may involve removal of some unnecessary whitespace from the CSS placed in the <style>
      * element added to the document, due to the way that certain parts are `trim`med.  Notably, whitespace either side
@@ -24,25 +46,21 @@ trait CssConstraint
      *
      * @return string Needle to use with `assertRegExp` or `assertNotRegExp` instead.
      */
-    private static function getCssNeedleRegExp(string $needle): string
+    protected static function getCssNeedleRegularExpressionPattern(string $needle): string
     {
         $needleMatcher = \preg_replace_callback(
-            '/\\s*+([{},])\\s*+|(^\\s++)|(>)\\s*+|(?:(?!\\s*+[{},]|^\\s)[^>])++/',
+            self::CSS_REGEX_PATTERN,
             static function (array $matches): string {
-                if (isset($matches[1]) && $matches[1] !== '') {
-                    // matched possibly some whitespace, followed by "{", "}" or ",", then possibly more whitespace
-                    return '\\s*+' . \preg_quote($matches[1], '/') . '\\s*+';
+                switch (true) {
+                    case isset($matches[1]) && $matches[1] !== '':
+                        return '\\s*+' . \preg_quote($matches[1], '/') . '\\s*+';
+                    case isset($matches[2]) && $matches[2] !== '':
+                        return '\\s*+';
+                    case isset($matches[3]) && $matches[3] !== '':
+                        return \preg_quote($matches[3], '/') . '\\s*+';
+                    default:
+                        return \preg_quote($matches[0], '/');
                 }
-                if (isset($matches[2]) && $matches[2] !== '') {
-                    // matched whitespace at start
-                    return '\\s*+';
-                }
-                if (isset($matches[3]) && $matches[3] !== '') {
-                    // matched ">" (e.g. end of <style> tag) followed by possibly some whitespace
-                    return \preg_quote($matches[3], '/') . '\\s*+';
-                }
-                // matched any other sequence which could not overlap with the above
-                return \preg_quote($matches[0], '/');
             },
             $needle
         );
--- a/tests/Support/Constraint/CssConstraint.php
+++ b/tests/Support/Constraint/CssConstraint.php
@@ -62,15 +62,16 @@ abstract class CssConstraint extends Constraint
      */
     private static function getCssNeedleRegularExpressionReplacement(array $matches): string
     {
-        switch (true) {
-            case isset($matches[1]) && $matches[1] !== '':
-                return '\\s*+' . \preg_quote($matches[1], '/') . '\\s*+';
-            case isset($matches[2]) && $matches[2] !== '':
-                return '\\s*+';
-            case isset($matches[3]) && $matches[3] !== '':
-                return \preg_quote($matches[3], '/') . '\\s*+';
-            default:
-                return \preg_quote($matches[0], '/');
+        if (($matches[1] ?? '') !== '') {
+            $regularExpressionEquivalent = '\\s*+' . \preg_quote($matches[1], '/') . '\\s*+';
+        } elseif (($matches[2] ?? '') !== '') {
+            $regularExpressionEquivalent = '\\s*+';
+        } elseif (($matches[3] ?? '') !== '') {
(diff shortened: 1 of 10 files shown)


tests/Support/Constraint/CssConstraint.php
```diff
@@ -0,0 +1,51 @@
...
+     * readable and can use inputs containing extra whitespace.
+     *
+     * @param string $needle Needle that would be used with `assertContains` or `assertNotContains`.
+     *
+     * @return string Needle to use with `assertRegExp` or `assertNotRegExp` instead.
+     */
+    private static function getCssNeedleRegExp(string $needle): string
+    {
+        $needleMatcher = \preg_replace_callback(
+            '/\\s*+([{},])\\s*+|(^\\s++)|(>)\\s*+|(?:(?!\\s*+[{},]|^\\s)[^>])++/',
```
oliverklee, line 30: Can we move this string into a constant?

tests/Support/Constraint/CssConstraint.php
```diff
@@ -0,0 +1,51 @@
...
+            static function (array $matches): string {
```
oliverklee, line 31: Please let's move this into a non-anonymous function with a proper description and a more specific parameter type annotation

tests/Support/Constraint/CssConstraint.php
```diff
@@ -0,0 +1,51 @@
...
+                    // matched possibly some whitespace, followed by "{", "}" or ",", then possibly more whitespace
+                    return '\\s*+' . \preg_quote($matches[1], '/') . '\\s*+';
+                }
+                if (isset($matches[2]) && $matches[2] !== '') {
+                    // matched whitespace at start
+                    return '\\s*+';
+                }
+                if (isset($matches[3]) && $matches[3] !== '') {
+                    // matched ">" (e.g. end of <style> tag) followed by possibly some whitespace
+                    return \preg_quote($matches[3], '/') . '\\s*+';
```
oliverklee, line 42: Generally, please let's not have those early returns. I'd like to have the return point either at the end of a method or (for guard clauses that cover error or empty cases) at the beginning.

oliverklee: Maybe we can either use if/elseif chains? Or move this to a table stored in an array?

tests/Support/Constraint/StringContainsCss.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * @inheritdoc
+     */
+    public function toString(): string
+    {
+        return 'contains CSS `' . $this->css . '`';
+    }
+
+    /**
+     * @inheritdoc
```
oliverklee, line 45: Please no `@inheritdoc`. It's good for machines, but not very useful for the human reader.

tests/Support/Constraint/StringContainsCssCount.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->count = $count;
+        $this->css = $css;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function toString(): string
+    {
+        return 'contains exactly ' . (string)$this->count . ' occurrence(s) of CSS `' . $this->css . '`';
```
oliverklee, line 48: Any particular reason why you are using backticks here and no double quotes?

tests/Support/Constraint/StringContainsCssCount.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * @inheritdoc
```
oliverklee, line 52: No `@inhertidoc`, please (see above).

tests/Support/Constraint/StringContainsCssCount.php
JakeQZ: Because it's code (it's CSS, but still code 😉).  I note that in the PHPUnit tests they use `"`, but those are for generic strings.  Quotes are for what people said.  Backticks are for what people coded.  I think.  YMMV...

tests/Support/Constraint/StringContainsCss.php
JakeQZ: Fair enough.  It works fine when sent though something like phpDocumentor, for online documentation, but I guess that's not the most common use case (particularly for an internal class)!

JakeQZ:
The method `getCssNeedleRegExp` was simply copied from the other class.  I think it just about passed minimum standard for acceptance at the time.  I agree, now would be a good time to tidy it up.

Particularly, as you might have imagined, there will be changes coming in that area for `Sabberworm`...

Offhand, I think these will be:
- allow variation in whitespace around more symbollic characters (`:`, `;`);
- allow variation in whitespace of whitespace itself;
- allow `;` to be inserted at end of (minified) declarations block<sup>[1]</sup>;
- allow numbers to have leading or (after the decimal point) trailing zeros stripped off;
- allow `url(foo.scam)` to be converted to `url("foo.scam")`.

[1] This would allow some unminification of supplied minified CSS, but I note that the current implementation allows arbitraray whitespace to be added in certain places, and no-one's ever complained about spaces in their inlined style attributes (surely `style="color: green; font-size: 18px;"` would be better as `style="color:green;font-size:18px"`).  Someone might rightfully say that it would all compress just as well under GZIP.  I would laugh, and say: "Are you sure email is compressed during transmission?"  (I don't know.)  Sabberworm has a kind-of solution - it allows you to [specify an output formatter](https://github.com/sabberworm/PHP-CSS-Parser#output)...  (BTW, I am liking this, it looks 99.99% promising, and has 16M downloads, which can't be a bad thing.)

tests/Support/Constraint/StringContainsCssCount.php
oliverklee: The backtick thing comes from Markdown as it is used on GitHub, and then spilled over to some PHPDoc comments. I've never seen it applied to strings in code before. (I'm not opposed to this, either.)

JakeQZ: > The backtick thing comes from Markdown

Actually I think its use in Markdown comes from historical usage for quoting code.  Note that PHP has a [backtick operator](https://www.php.net/manual/en/language.operators.execution.php) that, I'm sure, predates Markdown.

tests/Support/Constraint/CssConstraint.php
JakeQZ: Done.

Unfortunately due to a current limitation of PHP ([#75060](https://bugs.php.net/bug.php?id=75060)) this means that the `trait` must now be implemented as an abstract base class (in C++ this wouldn't be such an issue, because it allows multiple inheritance - however, I think traits are cleaner, and I have pushed, with the backing of 72 others, for this to be addressed in PHP 8.0).

Given that this is used only in testing, and thus extreme optimization is not required, I have also made use of the `x` modifier so that the regex can be self-documented inline.

tests/Support/Constraint/CssConstraint.php
JakeQZ: Oops.  Forgot about this.  Will address...

tests/Support/Constraint/CssConstraint.php
JakeQZ: Methinks this calls for the `switch (true)` tool.  I hope you find it makes it much more readable, whilst being functionally equivalent...

tests/Support/Constraint/StringContainsCss.php
JakeQZ: Still todo :/

tests/Support/Constraint/CssConstraint.php
JakeQZ: Now done.

tests/Support/Constraint/StringContainsCss.php
JakeQZ: Now done.

JakeQZ:
> I like the approach. A lot.

Kudos there goes to @sebastianbergmann, I think.  I am merely following the pattern.

> I've added some remarks.

I've addressed these.  I also renamed a method to avoid an abbreviation - with `opcache` now as standard in PHP 7.0+, I don't see much value any more in shorter variable and method names - human understanding I think should be the driving factor moving forwards.

tests/Support/Constraint/CssConstraint.php
```diff
@@ -50,20 +49,28 @@ protected static function getCssNeedleRegularExpressionPattern(string $needle):
...
+    {
+        switch (true) {
+            case isset($matches[1]) && $matches[1] !== '':
+                return '\\s*+' . \preg_quote($matches[1], '/') . '\\s*+';
+            case isset($matches[2]) && $matches[2] !== '':
+                return '\\s*+';
+            case isset($matches[3]) && $matches[3] !== '':
+                return \preg_quote($matches[3], '/') . '\\s*+';
+            default:
+                return \preg_quote($matches[0], '/');
```
oliverklee, line 73: We still have the early returns here. Please let's change this to put the result in a local variable in the different cases and return that local variable at the end of the function.

tests/Support/Constraint/CssConstraint.php
```diff
@@ -50,20 +49,28 @@ protected static function getCssNeedleRegularExpressionPattern(string $needle):
...
     }
+
+    /**
+     * @param string[] $matches array of matches for {@see CSS_REGEX_PATTERN}
+     *
+     * @return string replacement string, which may be `$matches[0]` if no alteration is needed
+     */
+    private static function getCssNeedleRegularExpressionReplacement(array $matches): string
+    {
+        switch (true) {
```
oliverklee, line 65: This looks like a misuse of `switch` to me (although it syntactically is valid), and I find it confusing to read. Please let's change this to an `if`/`elseif`/`else` chain that communicates the intent more clearly.

tests/Support/Constraint/CssConstraint.php
```diff
@@ -50,20 +49,28 @@ protected static function getCssNeedleRegularExpressionPattern(string $needle):
...
+            case isset($matches[1]) && $matches[1] !== '':
+                return '\\s*+' . \preg_quote($matches[1], '/') . '\\s*+';
+            case isset($matches[2]) && $matches[2] !== '':
+                return '\\s*+';
+            case isset($matches[3]) && $matches[3] !== '':
```
oliverklee, line 70: We have this `isset(foo) && foo` construct multiple times here (and also some matches are used multiple times). Would it work to
1. move the matches into local, well-named variables
1. also use the [null coalescing operator](https://www.php.net/manual/de/migration70.new-features.php#migration70.new-features.null-coalesce-op) instead: `$foo = ($matches[1] ?? '') !== ''`;

tests/Support/Constraint/CssConstraint.php
JakeQZ: Hmm.  It's arguable that they are not early returns because there is no more code on any execution branch.  But I will use a temporary local variable...

tests/Support/Constraint/CssConstraint.php
JakeQZ: OK.  To me it makes it clear, but I appreciate other people's mileage may vary.  Will change.

tests/Support/Constraint/CssConstraint.php
JakeQZ: Good point about `isset … &&`…

oliverklee:
@JakeQZ I think there has been no new push after your latest comments. Are you still working on the updates, or is there a new version for me to review?

JakeQZ:
> @JakeQZ I think there has been no new push after your latest comments. Are you still working on the updates, or is there a new version for me to review?

Sorry, I have not had time (as indicated on another thread).  Responding to comments is one thing, but actually making code changes is more demanding.  I'll try and get this wrapped up in the next couple of days (I still don't have time right now, as it needs a bit of thinking :/).

JakeQZ:
Finally found some time, and have addressed the points (not sure why I thought I needed to think about it significantly).

However, there's an issue with the GitHub actions - it seems `composer update` is being run without `composer install`...:

```
Run if [[ "highest" == 'lowest' ]]; then
  if [[ "highest" == 'lowest' ]]; then
    DEPENDENCIES='--prefer-lowest';
  else
    DEPENDENCIES='';
  fi;
  composer update --no-progress "${DEPENDENCIES}";
  composer show;
  shell: /bin/bash -e {0}
Cannot update only a partial set of packages without a lock file present.
Error: Process completed with exit code 1.
```

JakeQZ:
Rebased with #924 included.  The build is passing now.

tests/Support/Constraint/StringContainsCss.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Evaluates the constraint for the parameter `$other`.
+     *
+     * @param mixed $other value or object to evaluate
+     *
+     * @return bool `true` if the constraint is met, `false` otherwise
+     */
+    protected function matches($other): bool
+    {
+        return \preg_match(self::getCssNeedleRegularExpressionPattern($this->css), $other) > 0;
```
oliverklee, line 49: The second parameter of `preg_match` needs to be a string. So we either should add a `string` type declaration to `$other` or return an early `false` if `$other` is no string.

tests/Support/Constraint/StringContainsCssCount.php
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * Evaluates the constraint for the parameter `$other`.
+     *
+     * @param mixed $other value or object to evaluate
+     *
+     * @return bool `true` if the constraint is met, `false` otherwise
+     */
+    protected function matches($other): bool
+    {
+        return \preg_match_all(self::getCssNeedleRegularExpressionPattern($this->css), $other) === $this->count;
```
oliverklee, line 56: Ditto about `$other` needing to be a string.

tests/Unit/Support/Constraint/Fixtures/TestingCssConstraint.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Pelago\Emogrifier\Tests\Unit\Support\Constraint\Fixtures;
+
+use Pelago\Emogrifier\Tests\Support\Constraint\CssConstraint;
+
+/**
+ * Extends the `CssConstraint` class to provide indirect access to protected method(s) for testing.
+ *
+ * @author Jake Hotson <jake.github@qzdesign.co.uk>
+ */
+abstract class TestingCssConstraint extends CssConstraint
```
oliverklee, line 14: If I read this correctly, this class is intended to be neither sublassed nor instantiated. In this case, it should be `abstract final` or `final abstract` (I don't know if PHP allows this, though).

oliverklee (changes requested):
Looks very good already. I've added my comments for the finishing touches.

tests/Support/Constraint/StringContainsCss.php
JakeQZ: It's an implementation of an abstract method, so the parameter type can't be changed.  Thus I've added a check that it's a string, along with tests for that.

tests/Support/Constraint/StringContainsCssCount.php
JakeQZ: Ditto

tests/Unit/Support/Constraint/Fixtures/TestingCssConstraint.php
JakeQZ: A class cannot be both `abstract` and `final`.  So I have not changed anything here.  If you want, I could expand on the commentary, but OTOH, who knows what the future holds?  It's not intended to be subclassed or instantiated, but there might be some reason to do so that we haven't thought of yet.

JakeQZ:
Ready for re-review...

oliverklee (approved):
LGTM.

PR #909, opened 2020-06-19
[TASK] Automatically add a backslash for global functions

oliverklee:
Using a leading backslash when calling a function in the global namespace
provides a (small) performance benefit.

Adjust the PHP CS Fixer configuration to add any missing backslashes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   [#907](https://github.com/MyIntervals/emogrifier/pull/907))
 
 ### Changed
+- Automatically add a backslash for global functions
+  ([#909](https://github.com/MyIntervals/emogrifier/pull/909))
 - Update the development tools
   ([#898](https://github.com/MyIntervals/emogrifier/pull/898),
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#894](https://github.com/MyIntervals/emogrifier/pull/894))
 
 ### Changed
+- Upgrade to PHPUnit 7.5
+  ([#908](https://github.com/MyIntervals/emogrifier/pull/908))
```
JakeQZ, line 14: Forget to edit after copy and paste? ;))

oliverklee: Good catch. Obviously very much so. :-)

oliverklee:
> I'm sure I added something to PHPCS (the other one, not the fixer) to catch this, shortly after you added the FQN prefix throughout. 

Indeed. (At least we seem to have a working PHPCS rule for this now - I've just checked.)

oliverklee:
Fixed the changelog (also the issue number there) and repushed.

PR #911, opened 2020-06-21
[BUGFIX] Remove uppercase !important from attributes

JakeQZ:


tests/Unit/CssInlinerTest.php
```diff
@@ -2785,7 +2785,7 @@ public function importantIsCaseInsensitive(): void
 
         $subject->inlineCss('p { margin: 1px !ImPorTant; }');
 
-        self::assertStringContainsString('<p style="margin: 1px !ImPorTant;">', $subject->renderBodyContent());
+        self::assertStringContainsString('<p style="margin: 1px;">', $subject->renderBodyContent());
```
oliverklee, line 2788: I think we then also should rename the test.

oliverklee: And maybe we can use a data provider for the the `important` variants (lowercase, all-uppercase, mixed case).

JakeQZ: > And maybe we can use a data provider for the the `important` variants (lowercase, all-uppercase, mixed case).

Good idea.  I have done that, and adjusted the tests.  I also then found another bug - `! important` (with a space after the `!`) was not being removed as intended (though it was being handled correctly elsewhere).

The final fix involves changing the `str_replace` not to a `str_ireplace` but to a `preg_replace`.  Now Psalm complains because `preg_replace` _might_ return `null`.  So I have had to implement a wrapper to ensure that is not the case.

Do some of the other Psalm issues included in the `baseline` relate to this (we use `preg_replace` elsewhere)?  Maybe this wrapper will allow that specific issue to be addressed in those other cases.

JakeQZ: Seems I also need to disable a PHPCS rule, because I declared the extra exception that could be thrown.

JakeQZ:
This turns out to be not quite as straightforward as I'd hoped.  The PR is ready for re-review.  However, if you'd like it split into separate PRs, I'd be both happy and loathed to do so ;)

On the subject of whether `@throws` should be added for exceptions thrown by methods called (I think it should), [I found this](https://discuss.neos.io/t/to-throws-or-not-to-throws-indirect-exceptions/3018).

src/CssInliner.php
```diff
@@ -1044,7 +1049,7 @@ private function removeImportantAnnotationFromNodeInlineStyle(\DOMElement $node)
         $importantStyleDeclarations = [];
         foreach ($inlineStyleDeclarations as $property => $value) {
             if ($this->attributeValueIsImportant($value)) {
-                $importantStyleDeclarations[$property] = \trim(\str_replace('!important', '', $value));
+                $importantStyleDeclarations[$property] = $this->pregReplace('/\\s*+!\\s*+important$/i', '', $value);
```
JakeQZ, line 1052: Psalm says that `preg_replace` might return `null`.  I think it's more likely that the server will crash and it won't return anything at all.

Please reject this change and either remove Psalm from the project or tell me how I'm supposed to actually make code changes that satisfy Psalm in some kind-of semi-positive way.

oliverklee: In this case, Psalm is right. [`preg_replace` will return `null` if there is an error](https://www.php.net/manual/en/function.preg-replace.php).

For doing this in a cleaner way, we could do something like this:
`… = \preg_replace(…) ?? ''`
This will return an empty string in case of an error.
Or we could throw an exception if `null` is returned to make the problem visible.

PR #914, opened 2020-07-05
[CLEANUP] Handle preg_replace returning null in a simpler way

oliverklee:
In this specific case, `preg_replace` will never return `null`,
and we do not need to handle that case. Instead, a cast
will suffice to make sure static type analysis will not
complain.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -1049,7 +1049,7 @@ class CssInliner extends AbstractHtmlProcessor
         $importantStyleDeclarations = [];
         foreach ($inlineStyleDeclarations as $property => $value) {
             if ($this->attributeValueIsImportant($value)) {
-                $importantStyleDeclarations[$property] = $this->pregReplace('/\\s*+!\\s*+important$/i', '', $value);
+                $importantStyleDeclarations[$property] = (string)\preg_replace('/\\s*+!\\s*+important$/i', '', $value);
             } else {
                 $regularStyleDeclarations[$property] = $value;
             }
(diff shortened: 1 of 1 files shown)


oliverklee:
>  IMO the best course of action in such cases is to log an error and return the original string

Do we then still need any changes over the current code, or can we close this PR?

JakeQZ:
> Do we then still need any changes over the current code, or can we close this PR?

It may boil down to a matter of opinion.

I suspect that you may have approved #911 without spotting the chunk of code that was added to deal with `preg_replace` possibly returning `null`, and which could have been put in as a separate commit.

However, I think this is the approach to take, and we can, moving forward, substitute calls to `preg_replace` with the wrapper method.  Perhaps some of the current Psalm warnings relate to this. 

It is the approach I've taken on other projects (WordPress-based), where a regular expression replacement of something in some HTML (like the entire page content) has turned out to be the most viable solution (the other would be reparsing the HTML into DOM, of course, which seems like overkill), and occasionally I have found `preg_*` to fail.  This is why I am super-keen on always using the possessive `+` in regex patterns; in pretty much all cases this fixes the problem.  But as a 'double-bagged' fallback for `preg_replace` I would say that in most situations you are more likely to want the original string returned rather than an empty one - in the WordPress application it would mean the user gets a page of content which is not quite as it should be, instead of a completely unhelpful blank page.  Much better UX IMO.

But do think about it before closing.

oliverklee:
> I suspect that you may have approved #911 without spotting the chunk of code that was added to deal with `preg_replace` possibly returning `null`, and which could have been put in as a separate commit.

Actually, I noticed it, and I understood your comments in that PR that you'd like to find a different way to deal with this. It looks like I misunderstood (or the comment was outdated).

JakeQZ:
> Actually, I noticed it, and I understood your comments in that PR that you'd like to find a different way to deal with this. It looks like I misunderstood (or the comment was outdated).

The comment was based on realizing that it perhaps ought to have been split into two separate PRs: one for the `preg_replace`; and one for the actual bugfix.  However, I was trying to wrap a few things up quickly and getting frustrated by a constant trail of Psalm errors (there's one error; you fix that; then there's another identical error at the next stage downstream; ad infinitum), and did not really have time at the time.  That was the only 'different way' (i.e. one of process, not code).

PR #915, opened 2020-07-05
[TASK] Convert psalm-specific type annotations

### LINKED ISSUE #913: Switch Psalm-specific array annotations to the non-Psalm-specific version
`@param`/`@return`/@var` annotations like `array<int, string>` are now recognized by PhpStorm's code inspection.

This will make our type annotations more concise and allows us to switch to another static type analysis tool (e.g., PHPStan) more easily.

We will need to modify the PHP_CodeSniffer settings for this.

oliverklee:
`@param`/`@return`/`@varannotations` like `array<int, string>`
are now recognized by PhpStorm's code inspection.

This will make our type annotations more concise and allows us
to switch to another static type analysis tool (e.g., PHPStan) more easily.

Fixes #913

### DIFF
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -1129,16 +1129,12 @@ final class CssInlinerTest extends TestCase
      * Specificity ordering for selectors involving pseudo-classes, attributes and `:not` is covered through the
      * combination of these tests and the equal specificity tests and thus does not require explicit separate testing.
      *
-     * @return string[][]
-     *
-     * @psalm-return array<string, array<int, string>>
+     * @return array<string, array<int, string>>
      */
     public function differentCssSelectorSpecificityDataProvider(): array
     {
         /**
-         * @var string[] Selectors targeting `<span id="text">` with increasing specificity
-         *
-         * @psalm-var array<string, string>
+         * @var array<string, string> Selectors targeting `<span id="text">` with increasing specificity
          */
         $selectors = [
             'universal' => '*',
@@ -2185,15 +2181,11 @@ final class CssInlinerTest extends TestCase
     }
 
     /**
-     * @param string[] $precedingSelectorComponents Array of selectors to which each type of pseudo-component is
-     *                                              appended to create a selector for a CSS rule.
-     *                                              Keys are human-readable descriptions.
+     * @param array<string, string> $precedingSelectorComponents selectors to which each type of pseudo-component is
+     *                                                           appended to create a selector for a CSS rule.
+     *                                                           Keys are human-readable descriptions.
      *
-     * @psalm-param array<string, string> $precedingSelectorComponents
-     *
-     * @return string[][]
-     *
-     * @psalm-return array<string, array<int, string>>
+     * @return array<string, array<int, string>>
      */
     private function getCssRuleDatasetsWithSelectorPseudoComponents(array $precedingSelectorComponents): array
     {
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -2182,8 +2182,7 @@ final class CssInlinerTest extends TestCase
 
     /**
      * @param array<string, string> $precedingSelectorComponents selectors to which each type of pseudo-component is
-     *                                                           appended to create a selector for a CSS rule.
-     *                                                           Keys are human-readable descriptions.
+     *        appended to create a selector for a CSS rule. Keys are human-readable descriptions.
      *
      * @return array<string, array<int, string>>
      */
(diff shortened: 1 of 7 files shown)


oliverklee:
@zoliszabo @JakeQZ ping

oliverklee:
> However, I think the definitive would perhaps be phpDocumentor. Have you tested this against that?

No, not yet. I have not created HTML API documentation for my PHP code for years.

> So I'd prefer to retain the type annotation checking in PHPCS if it doesn't present a particular problem (other than purity of tool use :]).

In this case, there is another rationale for me to move away from using the additional sniffs from `slevomat/coding-standard`: I have not been able to find any sane way to use PHPCS via Phive while having additional packages with sniffs, i.e., using `slevomat/coding-standard` currently is a blocker for moving PHPCS to Phive.

Would it be okay for you if we dropped `slevomat/coding-standard` so we can move PHPCS to Phive?

JakeQZ:
It was originally added for #576 to ensure global function calls used FQNs.  Possible we added some other sniffs it provides.  It's possible that by now we are also using another tool that can or does check these.  Or if not we can find a way or revisit this.

So I don't mind if we remove this for the time being, but if we do, I think we should ensure that the code checks are reinstated in future if not covered elsewhere (e.g. by creating a separate issue to revisit this), otherwise they will inevitably be occasionally missed.

tests/Unit/CssInlinerTest.php
```diff
@@ -2185,15 +2181,11 @@ public function inlineCssNotKeepsUnneededMediaRuleAfterEmptyMediaRule(string $em
...
-     * @param string[] $precedingSelectorComponents Array of selectors to which each type of pseudo-component is
-     *                                              appended to create a selector for a CSS rule.
-     *                                              Keys are human-readable descriptions.
-     *
-     * @psalm-param array<string, string> $precedingSelectorComponents
-     *
-     * @return string[][]
+     * @param array<string, string> $precedingSelectorComponents selectors to which each type of pseudo-component is
+     *                                                           appended to create a selector for a CSS rule.
+     *                                                           Keys are human-readable descriptions.
```
JakeQZ, line 2186: I think we should indent like this:

```php
/**
 * @param bool $needsExplanation This parameter requires some explanaining which spans onto
 *        multiple lines.
 * @param bool $selfExplanatory
 */
```

We have this style a fair bit (thought not entirely consistently) in the code base, and looking at other projects, it seems to be a happy middle ground.  (I doubt there's a sniff or suchlike for it, though you never know.)

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -712,9 +710,8 @@ public function notAddsSecondContentTypeMetaTag(string $htmlBefore, string $html
     }
 
     /**
-     * @return string[][]
-     *
-     * @psalm-return array<string, array{0:string, 1:string}>
+     * @phpcsSuppress SlevomatCodingStandard.TypeHints.ReturnTypeHint
+     * @array<string, array{0:string, 1:string}>
```
JakeQZ, line 714: Should be `@return array...` rather than `@array...`.  Perhaps then the `@phpcsSuppress` would not be needed.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -760,25 +755,23 @@ static function (array $dataset) {
     }
 
     /**
-     * @return string[][] Each dataset has three elements in the following order:
+     * @return array<string, array{0:string, 1:string, 2:string}>
+     *         Each dataset has three elements in the following order:
      *         - HTML with non-XML self-closing tags (e.g. "...<br>...");
      *         - The equivalent HTML with XML self-closing tags (e.g. "...<br/>...");
      *         - The name of a self-closing tag contained in the HTML (e.g. "br").
```
JakeQZ, line 762: Multiline `@return` DocBlock entries are generally not indented from what I've seen, as there is only ever one `@return` and there will be blank lines above and below if necessary.  Though I accept this is how it was already formatted; it's a case of 'since you're here you could change it'.

JakeQZ (changes requested):
Looks good, but there appears to be a mistake in one of the DocBlocks, and I've made suggestions about indentation on a couple of others.

Since we've updated `slevomat/coding-standard` to 6.x, has that resolved the previous issue regarding this style of DocBlock type specification?

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
oliverklee: Ah, that's why PHPCS was complaining. Now everything makes sense. :-) Will change.

oliverklee: Fixed.

tests/Unit/CssInlinerTest.php
oliverklee: Fixed.

oliverklee:
Updated, rebased and repushed.

JakeQZ (approved):

Issue #916, opened 2020-07-10
Allow PHP 8 installs


oliverklee:
… a bit like this (but with GitHub actions): https://github.com/tijsverkoyen/CssToInlineStyles/pull/198/files

JakeQZ:
Is the PHP 8 alpha available via GitHub actions now?

Yes please, let's be on the ball about this and future PHP versions, so we are not having to make patch releases to support a new PHP version :)

oliverklee:
> Is the PHP 8 alpha available via GitHub actions now?

Yes, the setup PHP action [provides the PHP 8 nightly](https://github.com/marketplace/actions/setup-php-action#tada-php-support).

PR #926, opened 2020-11-20
[TASK] Support PHP 8.0

### LINKED ISSUE #925: Support PHP 8
Will need `composer.json` updating, and changes to the GitHub Action testing matrix.  Hopefully that will be all apart from actually making a new release.  I believe the beta/RC version can already be used with GitHub Actions.

JakeQZ:
Resolves #925.


JakeQZ:
Blocked by #924 at present, but will hopefully enough here to give insight as to whether there is anything else we need to do or fix.

JakeQZ:
Lint passes, at least, and the checks seem to run fine against whatever RC/beta version is currently deployed by GitHub.

JakeQZ:
Ah, but we need to update `slevomat/coding-standard` (even though it is only used by the Sniffer against PHP 7.3, Composer requires all development dependencies to be compatible with all development tools), and this requires finding whatever replaced `SlevomatCodingStandard.TypeHints.TypeHintDeclaration`.

JakeQZ:
That's now resolved, but the `phpunit.phar` file needs updating.  However, that won't work.  There is no version of PHPUnit compatible with both PHP 8 and PHP 7.1.  We need 7.x for PHP 7.1, but 9.x for PHP 8.  Even dropping PHP 7.1 support won't help.

I think we need to revert back to Composer for PHPUnit (i.e. revert #894), unless there's a way for Phive to handle this situation?

oliverklee:
@JakeQZ I've tested whether Phive allows parallel installation of multiple PHPUnit versions: It doesn't. However, we could manually add a `tools/phpunit-9.4.phar` use that conditionally on the used PHP version. What do you think?

JakeQZ:
> 
> 
> @JakeQZ I've tested whether Phive allows parallel installation of multiple PHPUnit versions: It doesn't. However, we could manually add a `tools/phpunit-9.4.phar` use that conditionally on the used PHP version. What do you think?

Can we not run `phive update phpunit` prior to running the tests, and have `<phar name="phpunit" version="^7.5 || ^9.4" installed="7.5.20" location="./tools/phpunit.phar" copy="true"/>` in `phive.xml`.  Do we even need the `.phar` files in the repository at all (other than to speed up setting up the development environment)?  Will PHIVE be installed and available to GitHub Actions or would there need to be a separate step to install it?

JakeQZ:
Now have a green build :))

JakeQZ:
Will make a separate PR for the changes for PHPUnit & PHIVE...

JakeQZ:
This can be reviewed and merged now.  Note that the linked issue to be resolved is #916 not #925 (the latter which I closed as a duplicate).

oliverklee (approved):

PR #927, opened 2020-11-20
[TASK] Update `slevomat/coding-standard` to 6.x

JakeQZ:
This is required for the dev tools to run with PHP 8.0.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
     },
     "require-dev": {
         "php-parallel-lint/php-parallel-lint": "^1.2.0",
-        "slevomat/coding-standard": "^4.0.0",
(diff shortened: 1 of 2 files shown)


JakeQZ:
Will probably need #924 for the build to pass.

oliverklee:
I have merged #924 and rebased this PR. Let's see how the build goes.

oliverklee (approved):

JakeQZ:
Sorry, looks like I forgot to `git add composer.json`.  However, we need to require version 6, as version 4 is not compatible with the changes.  (The build passes because we don't run `sniff` with `--prefer-lowest`.)

PR #928, opened 2020-11-20
[TASK] Run the static analysis tools with PHP 7.4, not 7.3

oliverklee:
All the tools which we are using should properly support PHP 7.4 by
now, and we can benefit from the improved performance.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - 7.3
+          - 7.4
 
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


oliverklee:
Ah, it looks like we need to tweak our style rules a bit.

oliverklee:
Please ignore the checks that don't run. I'll need to update the CI settings once this is merged.

JakeQZ (approved):
LGTM.  I had thought about upping these versions when adding PHP 8 support, but that it should be a separate PR.

JakeQZ:
> I'll need to update the CI settings once this is merged.

I've not merged yet for this reason, but you can do so when you update the CI settings.

PR #929, opened 2020-11-20
[CLEANUP] Require `slevomat/coding-standard` 6.x

JakeQZ:
The changes in #927 are not compatible with version 4.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
     },
     "require-dev": {
         "php-parallel-lint/php-parallel-lint": "^1.2.0",
-        "slevomat/coding-standard": "^6.4",
+        "slevomat/coding-standard": "^6.4.1",
         "squizlabs/php_codesniffer": "^3.5.1"
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
     },
     "require-dev": {
         "php-parallel-lint/php-parallel-lint": "^1.2.0",
-        "slevomat/coding-standard": "^4.0.0 || ^6.4",
+        "slevomat/coding-standard": "^6.4",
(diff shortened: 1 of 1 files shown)


JakeQZ:
The only way I can think of to detect this type of issue in the CI scripts would be to run the static analysis tools against all supported PHP versions, and also with `--prefer-lowest`, which seems like overkill?

composer.json
```diff
@@ -44,7 +44,7 @@
     },
     "require-dev": {
         "php-parallel-lint/php-parallel-lint": "^1.2.0",
-        "slevomat/coding-standard": "^4.0.0 || ^6.4",
+        "slevomat/coding-standard": "^6.4",
```
oliverklee, line 47: Please let's be as specific as possible here so we are sure to have the latest bug fixes: `^6.4.1`

(We don't need to to care about incompatibilities when Emogrifier is used as a library as this is a dev dependency.)

oliverklee (changes requested):

oliverklee:
Yes, this would be overkill. The only static analysis that can have different results depending on the PHP version is the PHP linting; code sniffing, type checking etc. should use only one PHP version and one version of the tool.

composer.json
JakeQZ: I originally did have that in the `composer.json` file I failed to commit.  This time I tried to be more consistent with the change you committed in its place - but I see that was taken from the [change I committed](https://github.com/MyIntervals/emogrifier/pull/926/commits/32ddd12c352f5b36f24bd407731fd79d057f2659) to #926 which was in its infancy at the time.

JakeQZ:
Upped the minimum and this is ready for re-review.

oliverklee (approved):

PR #930, opened 2020-11-21
[TASK] Use PHPUnit 9.x with PHP 7.3+

JakeQZ:
This is required to be able to test against PHP 8.  However, PHPUnit 9.x
requires PHP 7.3, so a different version of PHPUnit is required for different
testing environments.

The main change here is a step in the GitHub Action to conditionally update
PHPUnit via PHIVE for PHP >=7.3.  PHIVE does not yet have the ability for
conditional installs (see
https://github.com/phar-io/phive/issues/295#issuecomment-731466678) so the
script must check the PHP version before running the update.

PHIVE has also been added to the tools (self-referencing) as it is not available
by default to GitHub Actions.

Note: There are warnings from PHPUnit 9.x about use of deprecated `assert`
methods (which will be removed in PHPUnit 10.x).  However, these don't cause the
tests to fail, and the replacement methods are not available in PHPUnit 7.x
which is still required to test against PHP 7.1 and 7.2.

Part of #925/#926.


composer.json
```diff
@@ -90,7 +90,8 @@
         "ci": [
             "@ci:static",
             "@ci:dynamic"
-        ]
+        ],
+        "phive:update:phpunit": "echo y | \"./tools/phive.phar\" update phpunit"
```
oliverklee, line 94: I propose we use the phive option `--no-progress` here as the progress bar does not really work when the output is logged.

And maybe we can explicitly trust Sebastian's GPG key D8406D0D82947747293778314AA394086372C20A by using the `--trust-gpg-keys` option instead of consenting to the first question phive asks.

.github/workflows/ci.yml
```diff
@@ -144,5 +144,11 @@ jobs:
           composer update --no-progress "${DEPENDENCIES}";
           composer show;
 
+      - name: Update PHIVE dependencies
+        run: |
+          if ! composer php:version | grep -q '^7\.[012]'; then
+            composer phive:update:phpunit
+          fi
```
oliverklee, line 151: I like this!

oliverklee (changes requested):

composer.json
JakeQZ: I could not find the documentation for PHIVE so did not know about these, but have now added the command line options.

composer.json
```diff
@@ -91,7 +91,7 @@
             "@ci:static",
             "@ci:dynamic"
         ],
-        "phive:update:phpunit": "echo y | \"./tools/phive.phar\" update phpunit"
+        "phive:update:phpunit": "\"./tools/phive.phar\" update phpunit --trust-gpg-keys --no-progress"
```
oliverklee: If I read the Phive documentation correctly, we'll probably need to specify a comma-separated list of key IDs:

```
--trust-gpg-keys         Silently import these keys when required (separated by comma)
```

Let's see whether this works without the key IDs as well in the build.

composer.json
JakeQZ: I could not find any documentation for PHIVE so did not know about these options.  However, the `--trust-gpg-keys` option appears to be unknown - I don't know if it requires PHIVE version 0.14 - I selected version 0.13.5 because that is compatible with PHP 7.1 (although we are only actually running it during the build for PHP >= 7.3).

oliverklee: Executing ```phive``` without any options should display its built-in help.

JakeQZ:
Managed to add `--no-progress` (it seems the options must come first on the command line), but it seems `--trust-gpg-keys` is not supported in PHIVE version 0.13.5 (I get 'unknown option' whether it's first or last on the command line)?

Where is the documentation for PHIVE?

composer.json
oliverklee: Then I'm all for using the latest version of PHIVE so we can use the current features and bug fixes.

oliverklee:
> Where is the documentation for PHIVE?

It's built-in: https://github.com/MyIntervals/emogrifier/pull/930#discussion_r528224884

oliverklee (changes requested):
Please let's use the latest PHIVE version and explicitly allow Sebastian's GPG key.

JakeQZ:
> 
> 
> explicitly allow Sebastian's GPG key.

This does not appear to be possible.  The `--trust-gpg-keys` option is only available with the `install` command, not with the `update` command.  With the `install` command, it does not seem possible to specify that the PHAR file be copied with a `.phar` extension (it is copied with no file extension, as simply e.g. `phpunit`), nor any version constraints - any pre-existing entry in `phive.xml` is simply ignored and overwritten.

oliverklee:
Okay, then it seems we need to keep this workaround for trusting the GPG key.

Please let's still use the latest PHIVE version.

JakeQZ:
I've added https://github.com/phar-io/phive/issues/296.

JakeQZ:
> Please let's still use the latest PHIVE version.

Done.

oliverklee:
I think #906 has introduced some changes that are not compatible with PHPUnit 9.

oliverklee:
PHPUnit's `Constraint` parent class does not have a constructor in PHPUnit 9.x anymore. So maybe we can copy the old version's constructor to our `CssConstraint` class:

```php
    public function __construct()
    {
        $this->exporter = new Exporter();
    }
```

(And create an issue that we should rework this once we're PHPUnit-9-only.)

JakeQZ:
> (And create an issue that we should rework this once we're PHPUnit-9-only.)

I've added a constructor which will conditionally call the parent constructor only if it exists.  This should cover all bases, including if some future version of PHPUnit reintroduces a constructor.  So I don't think any future rework would be required.

I can't help thinking that PHP should provide an implicit default constructor, as C++ does, to avoid this type of issue.

oliverklee (approved):
<3

JakeQZ:
There are some suggestions for how to handle PHIVE in relation to this in the responses to https://github.com/phar-io/phive/issues/296.  I don't know if this is something we want to address in the short term; it will become a non-issue when we drop PHP 7.2 support.

PR #566, opened 2018-04-06
[TASK] Add sniff for type hinting

JakeQZ:
Note:
`<exclude name="Squiz.Commenting.FunctionComment.ScalarTypeHintMissing"/>`
would not be required if `<config name="php_version" value="50500"/>` was
present, but the latter would probably have further-reaching consquences and
PHP 5 support will be deprecated soon.

Closes #563.

PR #933, opened 2020-11-21
[FEATURE] Also run the CI build once a week

### LINKED ISSUE #932: Run the CI once a week
This allows us to catch problems that are related to changes in the build environment or the dependencies instead of in our code, but that probably need fixing from our side.

oliverklee:
This allows us to catch problems that are related to changes in the
build environment or the dependencies instead of in our code,
but that probably need fixing from our side.

Fixes #932

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 - Run the CI build once a week
-  ([#932](https://github.com/MyIntervals/emogrifier/pull/932),
-  [#933](https://github.com/MyIntervals/emogrifier/pull/933))
+  ([#933](https://github.com/MyIntervals/emogrifier/pull/933))
 - Move more development tools to Phive
   ([#894](https://github.com/MyIntervals/emogrifier/pull/894),
   [#907](https://github.com/MyIntervals/emogrifier/pull/907))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+- Run the CI build once a week
+  ([#932](https://github.com/MyIntervals/emogrifier/pull/932),
```
JakeQZ, line 10: The URL should contain `issues` rather than `pull` - if needed at all (I think often but not always, historically, we have just linked the PR, as that should be sufficient).

JakeQZ (changes requested):
Mistake in the changelog URL, but otherwise looks good.

JakeQZ:
Regarding whether the issue as well as the PR should be referenced in the changelog, [you previously commented](https://github.com/MyIntervals/emogrifier/pull/566#discussion_r179775180) thus:

> 
> 
> The link to the PR should be enough in the changelog. You can add "FIxes #xy" as a single line in the commit message to relate to the ticket (and to autoclose it on merge).

CHANGELOG.md
oliverklee: Thank. Changed, rebased and repushed.

JakeQZ (approved):

PR #934, opened 2020-11-21
[BUGFIX] Explicitly set the Ubuntu version for GitHub Actions

### LINKED ISSUE #931: Update the OS version on GitHub actions to resolve a warning
This warning:
Ubuntu-latest workflows will use Ubuntu-20.04 soon. For more details, see https://github.com/actions/virtual-environments/issues/1816

oliverklee:
This fixes a warning when running the CI build.

Fixes #931

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
   php-lint:
     name: PHP Lint
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


JakeQZ (approved):
Approving, though I'm wondering if it's a warning we actually need to worry about.  Also, it means we may need to update the `runs-on` from time to time.  OTOH, at least we know where we are, and it's consistent with specifying version constraints for the tools (e.g. so we don't suddenly end with a broken build because, say, PHPUnit 10.x is now being used).

PR #939, opened 2020-11-21
[BUGFIX] Fix misspelling of "omission" in the tests

oliverklee:


JakeQZ (approved):
As a native English speaker, bizarrely (bazaarly?), I never even noticed :S

PR #940, opened 2020-11-22
[FEATURE] Add an .editorconfig file

### LINKED ISSUE #781: Add an editorconfig file
Helps keep a consistent edit style for line endings & indentation across editors (i.e. your php files appear to have a tab size of 4, whereas the markdown files have a tab size of 2, and yaml is always spaces etc.)

oliverklee:
This allows IDEs to automatically set the basic code formatting options.

Fixes #781

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,24 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# PHP files
+[*.php]
+indent_style = space
+indent_size = 4
+
+# MD files
+[*.md]
+indent_style = space
+indent_size = 4
+
+# XML files
+[*.xml]
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,24 @@
...
+
+# PHP files
+[*.php]
+indent_style = space
+indent_size = 4
+
+# MD files
+[*.md]
+indent_style = space
+indent_size = 4
```
JakeQZ, line 19: Do we not have 2 spaces for indenting in `.md` files - e.g. in bulleted lists - or is the indentation for lists something different?

JakeQZ (commented):
I just have a query about the indentation for `.md` files.

.editorconfig
oliverklee: According to these style guides, it's 4 spaces (for code and lists where we do not use use hanging indent determined by the number of digits):

- https://google.github.io/styleguide/docguide/style.html#nested-list-spacing
- https://guides.github.com/features/mastering-markdown/
- https://cirosantilli.com/markdown-style-guide/#indentation-of-content-inside-lists

This setting in the `.editorconfig` only determine what the tab key does, not what a linter might expect.

JakeQZ: OK.

JakeQZ (approved):

PR #944, opened 2020-11-22
[BUGFIX] Use the most specific exception for @throws annotations

### LINKED ISSUE #787: Use ExpectationFailedException rather than AssertionFailedError in HtmlPrunerTest

oliverklee:
`ExpectationFailedException` is more specific than `AssertionFailedError`,
and we should use that for `@throws` annotations so that consumers of these
exceptions can react in a more specific way.

Fixes #787


JakeQZ (approved):
Possibly `AssertionFailedError` was correct with earlier versions of PHPUnit but `ExpectationFailedException` is correct with PHPUnit 7, 8 and 9.

PR #947, opened 2020-11-23
[TASK] Allow addition/removal of optional `;` in CSS tests

JakeQZ:
Also allow variation in whitespace around `;`.

Would be needed to switch to an alternative CSS parser like
`sabberworm/php-css-parser` which may result in the addition or removal the
optional semicolon, or whitespace around semicolons, in CSS declarations -
see #544.


oliverklee (approved):
LTGM.

PR #948, opened 2020-11-23
[BUGFIX] Also update the dependencies when running Composer update

oliverklee:
This allows us to test compatibility with the lowest/highest versions
of all dependencies, not only the direct ones.


JakeQZ (approved):
I've previously seen Composer not updating indirect dependencies and wondered if it was a bug in Composer.  I see from this that it's a (default) option.

PR #949, opened 2020-11-23
[BUGFIX] Fix typo in a private constant name

oliverklee:
"expressiom" is not a valid expression. And it's also not an
abbreviation of "expressionism".


JakeQZ (approved):
Good spot.

PR #950, opened 2020-11-23
[TASK] Prepare release of version 5.0.0

### LINKED ISSUE #878: Get version 5.0.0 out of the door
Our [yet-to-be-released version 4.0.0](#874) will still support PHP 7.1. However, [PHP 7.1 has alreay reached its end of life as well](https://www.php.net/supported-versions.php).

So I'd like to release a version 5.0.0 shortly after the release of 4.0.0, allowing us to drop support for PHP 7.1 as well (and maybe add support for PHP 8, if we can do that in any way).

What do you think, @JakeQZ, @zoliszabo, @jjriv?

oliverklee:
Also fix the casing of "PHIVE" in the changelog.

Closes #878

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,25 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+
+### Changed
+
+### Deprecated
+
+### Removed
+
+### Fixed
+
+## 5.0.0
+
+### Added
 - Add an `.editorconfig` file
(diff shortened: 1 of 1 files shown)


oliverklee:
I have not change the master branch alias for two reasons:
1. We will need to change this anyway when we change the default branch to main.
2. I'd like to release version 5.0.1 before we start with development for version 6.0.

JakeQZ (approved):
> I have not change the master branch alias for two reasons

OK, I was about to ask about that.

PR #955, opened 2020-12-29
[BUGFIX] Disable Xdebug in the CI build

### LINKED ISSUE #954: Build Failure
https://github.com/MyIntervals/emogrifier/actions/runs/434894580

PHIVE now gives the following error when run to update PHPUnit:

> Call to undefined function PharIo\Phive\xdebug_disable()

A bit of research suggests [the problem is introduced](https://github.com/magento/magento2/issues/28961#issuecomment-734340591) [by XDebug 3.0](https://xdebug.org/docs/upgrade_guide).

As a temporary solution, would we be able to run `pecl install -f xdebug-2.9.8` in the GitHub action?

oliverklee:
This avoid problems with Xdebug 3 (the version that now automatically
gets installed).

It also should speed up the build a bit.

Fixes #954

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php-version }}
+          coverage: none
 
       - name: PHP Lint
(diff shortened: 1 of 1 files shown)


JakeQZ (approved):
LGTM.  We do not currently use XDebug for any tests.

PR #956, opened 2020-12-29
[TASK] Explicitly use Composer 2 in the CI builds 

oliverklee:
This makes sure we can reap the performance and traffic benefits
of Composer 2.


JakeQZ (approved):
I'm presuming the cache path change is because this is where Composer 2.0 now caches the downloaded packages?

JakeQZ:
> 
> 
> I'm presuming the cache path change is because this is where Composer 2.0 now caches the downloaded packages?

Just noticed the commit comment about this.

Issue #923, opened 2020-09-24
CssInliner getHeadElement Error (missing head)


dyanakiev:
Hello,

I have this error, i did a little reasearch and i know that its because of missing head tag, i cannot asure the existence of head tag aways because users generate the templates.

Im actually using `HtmlNormalizer::fromHtml(...)->render()` before running `CssInliner::fromHtml(..)->inlineCss()->render()`
Why HtmlNormalizer didn't added the head tag?

```
Exception: Return value of Pelago\Emogrifier\CssInliner::getHeadElement() must be an instance of DOMElement, null returned
/emogrifier/pelago/emogrifier/src/CssInliner.php [ 1169 ]
```

Im using 5.0.x-dev, PHP 7.3

dyanakiev:
As a note; I never had this issue with the old emogrifier, i upgraded to the new one because i had to update from php 7.0 to php 7.3...

oliverklee:
Hi @dyanakiev, could you please post the HTML that triggers the problem (either the full HTML or preferably a minimal HTML testcase) and the exact HTML you're using? Thanks!

dyanakiev:
Hi @oliverklee as requested im posting minimal HTML that triggers the error

```
<p>
   <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="format-detection" content="telephone=no">
   <meta name="format-detection" content="date=no">
   <meta name="format-detection" content="address=no">
   <meta name="format-detection" content="email=no">
</p>
<link href="http://fonts.googleapis.com/css?family=Noto+Sans:400,700" rel="stylesheet" type="text/css" />
[...]

PR #959, opened 2021-03-21
[BUGFIX] Allow "Content-Type" in content

### LINKED ISSUE #957: False positives for content-type meta element presence check
I've recently updated Emogrifier and when moving from the old style syntax to the new, I receive this error:
`PHP Fatal error:  Uncaught TypeError: Return value of Pelago\Emogrifier\CssInliner::getHeadElement() must be an instance of DOMElement, null returned in /var/www/vhosts/test.system/httpdocs/vendor/pelago/emogrifier/src/CssInliner.php:1293`

Is there a reason there must always be a `<head>` element?
https://github.com/MyIntervals/emogrifier/blob/b43b650880d189b0ada61d95d0729c7424b1752d/src/CssInliner.php#L1291

I'm just inlining CSS in a simple email template that doesn't use a `<head>` element for anything.
I've added a blank one for now but strikes me as a little arbitrary.

Thanks!

JakeQZ:
Use a regular expression to determine whether the provided HTML contains a
`Content-Type` `meta` element instead of just testing for the string
"Content-Type".  Also allow for case insensitivity.

A valid `Content-Type` `meta` element which is invalidly placed (i.e. within the
`body`) will still be determined to be valid.  This will be addressed separately
for #923.

Add the package `rawr/cross-data-providers` to the development dependencies.
This is useful to generate a matrix of combinations for test data providers.
There are a few existing instances where this could be used to replace code
effectively doing that, but which are not addressed in this commit.

Rename affected data provider methods from `xyzDataProvider` to `provideXyx`,
which is more succinct and a more commonly used nomenclature on other projects.

Fixes #957.

### DIFF
--- a/src/HtmlProcessor/AbstractHtmlProcessor.php
+++ b/src/HtmlProcessor/AbstractHtmlProcessor.php
@@ -271,8 +271,9 @@ abstract class AbstractHtmlProcessor
      */
     private function addContentTypeMetaTag(string $html): string
     {
-        $hasContentTypeMetaTag
-            = \preg_match('%<meta(?=\\s)[^>]*\\shttp-equiv=(["\']?+)Content-Type\\g{-1}[\\s/>]%i', $html) === 1;
+        $contentTypeMetaTagMatchCount
+            = \preg_match('%<meta(?=\\s)[^>]*\\shttp-equiv=(["\']?+)Content-Type\\g{-1}[\\s/>]%i', $html);
+        $hasContentTypeMetaTag = \is_int($contentTypeMetaTagMatchCount) && $contentTypeMetaTagMatchCount > 0;
         if ($hasContentTypeMetaTag) {
             return $html;
         }
--- a/src/HtmlProcessor/AbstractHtmlProcessor.php
+++ b/src/HtmlProcessor/AbstractHtmlProcessor.php
@@ -271,7 +271,8 @@ abstract class AbstractHtmlProcessor
      */
     private function addContentTypeMetaTag(string $html): string
     {
-        $hasContentTypeMetaTag = \stripos($html, 'Content-Type') !== false;
+        $hasContentTypeMetaTag
+            = \preg_match('%<meta(?=\\s)[^>]*\\shttp-equiv=(["\']?+)Content-Type\\g{-1}[\\s/>]%i', $html) === 1;
         if ($hasContentTypeMetaTag) {
             return $html;
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
+- Allow "Content-Type" in content
+  ([#959](https://github.com/MyIntervals/emogrifier/pull/959))
 
 ## 5.0.0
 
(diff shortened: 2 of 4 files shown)


src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -271,7 +271,8 @@ private function normalizeDocumentType(string $html): string
      */
     private function addContentTypeMetaTag(string $html): string
     {
-        $hasContentTypeMetaTag = \stripos($html, 'Content-Type') !== false;
+        $hasContentTypeMetaTag
+            = \preg_match('%<meta(?=\\s)[^>]*\\shttp-equiv=(["\']?+)Content-Type\\g{-1}[\\s/>]%i', $html) === 1;
```
oliverklee, line 275: With the `=== 1`,  we've switched the meaning from "at least one content-type meta tag" to "exactly one content-type meta tag".

Even if multiple occurrences do not make much sense, I'd still prefer us to recognize this as "there is one" by keeping using `!== false`, and also cover this edge case with a test.

oliverklee: I don't quite understand the `(?=\\s)` part yet. What does it do?

oliverklee (changes requested):

src/HtmlProcessor/AbstractHtmlProcessor.php
JakeQZ: > 
> 
> With the `=== 1`, we've switched the meaning from "at least one content-type meta tag" to "exactly one content-type meta tag".
> 
> Even if multiple occurrences do not make much sense, I'd still prefer us to recognize this as "there is one" by keeping using `!== false`, and also cover this edge case with a test.

The return value does appear to indicate the number of matches (`0` or `1` because it is limited to one match) with `false` on error.  So to be more explicit we'd probably need something like

```php
$contentTypeMetaTagMatchCount = \preg_match(...);
$hasContentTypeMetaTag = $contentTypeMetaTagMatchCount !== false && $contentTypeMetaTagMatchCount > 0;
```

JakeQZ: > 
> 
> I don't quite understand the `(?=\\s)` part yet. What does it do?

It asserts that at that position the next character is a whitespace but without 'consuming' it ('lookahead').  If it was 'consumed', `\\shttp-equiv` would not match if there was nothing else beetween `meta` and `http-equiv` because the whitespace to be mached there would have already been consumed.  Without asserting `meta` is followed by whitespace, other tag names beginning with `meta` (hitherto as yet not defined) would also be matched.

oliverklee: > So to be more explicit we'd probably need something like
> 
> ```
> $contentTypeMetaTagMatchCount = \preg_match(...);
> $hasContentTypeMetaTag = $contentTypeMetaTagMatchCount !== false && $contentTypeMetaTagMatchCount > 0;
> ```

Or even more explicit:

```php
$hasContentTypeMetaTag = \is_int($contentTypeMetaTagMatchCount) false && $contentTypeMetaTagMatchCount > 0;
```

oliverklee: Thanks for the explanation!

JakeQZ:
I've changed the test for the `preg_match` return value as suggested, and also added tests to confirm the regex does not match unexpected longer keywords which contain the expected keyword as a substring.

oliverklee (approved):
LGTM.

PR #963, opened 2021-03-30
[TASK] Allow whitespace variations in CSS tests

JakeQZ:
Would be needed to switch to an alternative CSS parser like
`sabberworm/php-css-parser` which may result in some whitespace variations -
see #544.

Also added a test missing from #953 and updated a DocBlock to reflect that
change.


tests/Unit/Support/Constraint/CssConstraintTest.php
```diff
@@ -129,4 +130,32 @@ public function getCssNeedleRegularExpressionPatternInsertsOptionalWhitespaceAft
 
         self::assertSame('/\\<style\\>\\s*+a/', $result);
     }
+
+    /**
+     * @return string[][]
```
oliverklee, line 135: We can make this even more specific:
`array<string, array<string>>`

oliverklee (approved):

tests/Unit/Support/Constraint/CssConstraintTest.php
JakeQZ: Done.

PR #964, opened 2021-04-05
[TASK] Upgrade Psalm to version 4.7

### LINKED ISSUE #943: Update Psalm

oliverklee:
Also regenerate the baseline, but do not fix any code issues
(new or pre-existing).

Fixes #943

### DIFF
--- a/psalm.baseline.xml
+++ b/psalm.baseline.xml
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="3.11.5@3c60609c218d4d4b3b257728b8089094e5c6c6c2">
+<files psalm-version="4.x-dev@">
   <file src="src/CssInliner.php">
     <ArgumentTypeCoercion occurrences="1">
       <code>$node</code>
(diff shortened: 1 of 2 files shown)


psalm.baseline.xml
```diff
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="3.11.5@3c60609c218d4d4b3b257728b8089094e5c6c6c2">
+<files psalm-version="4.x-dev@">
```
JakeQZ, line 2: I'm curious as to why this now indicates a development branch rather than a specific release version.

JakeQZ (commented):
Looks fine, though `psalm-version` in `psalm.baseline.xml` is now showing as a development branch rather than a specific release.  Is it just a case of this being how it is now?

psalm.baseline.xml
oliverklee: This line (actually, the complete file) is autogenerated. It looks like the Psalm project has changed their `psalm-version` schema.

JakeQZ (approved):

PR #969, opened 2021-04-06
[CLEANUP] Make more test-related classes final

oliverklee:
This communicates that these classes are not intended to be subclassed.


JakeQZ (approved):
Seems reasonable.

PR #970, opened 2021-04-06
[CLEANUP] Slim down redundant class comments

oliverklee:
- Drop `@author` annotation as the Git history already has this
  information.
- Drop the `Test case.` comments from test cases as their purpose
  already is implicit by their namespace.


JakeQZ (approved):
I remember at least one person being apprehensive about having their name in an `@author` tag.

PR #971, opened 2021-04-06
[TASK] Stop using Slevomat PHPCS rules

oliverklee:
This allows us to use the PHAR version of PHP_CodeSniffer, getting us
one step further from dependency hell.


JakeQZ (approved):
This of course means that some of the issues covered will no longer be picked up by CI testing, though I'm aware some have since become covered elsewhere, e.g. `FullyQualifiedGlobalFunctions` now by PHP-CS-Fixer.

PR #973, opened 2021-04-06
[TASK] Upgrade to PHPUnit 8

### LINKED ISSUE #942: Update to PHPUnit 8 (for PHP 7.2)

oliverklee:
Also update the PHPUnit configuration file to use the recommended defaults.

Fixes #942


phpunit.xml
```diff
@@ -1,12 +1,23 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/7.5/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/8.5/phpunit.xsd"
          bootstrap="vendor/autoload.php"
          colors="true"
+         executionOrder="depends,defects"
```
JakeQZ, line 6: I could not find any documentation on what the eight possible values actually mean.

phpunit.xml
```diff
@@ -1,12 +1,23 @@
...
          verbose="true">
     <testsuites>
         <testsuite name="default">
-            <directory suffix="Test.php">tests</directory>
+            <directory>tests</directory>
         </testsuite>
     </testsuites>
+
+    <filter>
+        <whitelist processUncoveredFilesFromWhitelist="true">
```
JakeQZ, line 19: Again, struggling to understand what this is doing from the docs.

JakeQZ (commented):
I'm finding it hard to have any idea what a couple of the new entries in the XML file actually do.  Are you able to enlighten me?

phpunit.xml
oliverklee: depends: If test A depends on test B (as marked by an annotation), execute B before A

defects: Run those tests first that have failed during the previous run

phpunit.xml
oliverklee: This means that empty lines and comments can also be marked as executed/covered.

JakeQZ (approved):
OK.

PR #974, opened 2021-04-06
[CLEANUP] Make closures static

oliverklee:
This will slightly improve performance.


JakeQZ (approved):
A slight optimization, as well as clarifying that `$this` is not used.

PR #987, opened 2021-04-08
[BUGFIX] Allow line feeds within <html> tag

### LINKED ISSUE #981: Multiline <head> tag not matched?
Something like

```html
<html
  lang="de"
>
```

(with line breaks within the `<html>` tag) looks like it would cause issues, due to lack of `PCRE_DOTALL`/`s` modifer on the regex pattern matching it.

Needs further investigation, and suitable tests.

Fix should probably involve using the same style regex as for `<head>` to also ensure an unknown tag like `<htmler>` is not matched.

A quick search through the source does not turn up any other instances of `.*` in a regex pattern without the required modifier.

JakeQZ:
Also add tests to confirm that the `<html>` tag's attributes are preserved, and
a duplicate `<html>` tag does not arise.

Resolves #981.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Drop support for PHP 7.1 (#967)
 
 ### Fixed
+- Allow line feeds within `<html>` tag (#987)
 
 ## 5.0.1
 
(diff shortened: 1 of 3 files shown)


JakeQZ:
I've not gone as far as attempting to cater for rogue tags like `<htmler>` or `<xhtml>`, which would require more tests, and seem pretty unlikely to arise in any real world scenario, given that no such tag names are defined by W3C.

oliverklee (approved):

PR #988, opened 2021-04-09
[CLEANUP] Fix Psalm warnings in the utility classes

oliverklee:


psalm.xml
```diff
@@ -7,7 +7,7 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
-    errorBaseline="./psalm.baseline.xml"
+    errorBaseline="psalm.baseline.xml"
```
oliverklee, line 10: Note: Re-generating the baseline file also updates `psalm.xml` as seen here. I'd like to keep this change so updating the baseline file will in the future only change the baseline file.

src/Utilities/ArrayIntersector.php
```diff
@@ -22,14 +22,14 @@ class ArrayIntersector
     /**
      * the array with which the object was constructed, with all its keys exchanged with their associated values
      *
-     * @var (int|string)[]
+     * @var array<int|string, int|string>
```
JakeQZ, line 25: `array-key` rather than `int|string` could be used for these, though I don't know if one or the other is any clearer.

src/Utilities/CssConcatenator.php
```diff
@@ -150,7 +159,10 @@ private static function getMediaRuleCss(\stdClass $mediaRule): string
      */
     private static function getRuleBlockCss(\stdClass $ruleBlock): string
     {
-        $selectors = \array_keys($ruleBlock->selectorsAsKeys);
-        return \implode(',', $selectors) . '{' . $ruleBlock->declarationsBlock . '}';
+        /** @var string[] $selectors */
```
JakeQZ, line 162: Any reason why not `array<array-key, string>`?

JakeQZ (commented):
Changes look fine, just a couple of questions.

src/Utilities/CssConcatenator.php
oliverklee: No, no particular reason. Changed (and I've also added another type annotation).

src/Utilities/ArrayIntersector.php
oliverklee: Like this?
`array<array-key, array-key>`

JakeQZ: I was thinking `array<array-key, array-key>` - it's mentioned in the DocBlock for the class that the values have to be integers or strings becuase `array_flip` is used - though put another way the values have to be allowed as array keys for this reason, which in turn means they have to be integers or strings because those are the only types allowed for array keys.

But like I said, I don't know whether one or other is any clearer.

oliverklee: Then let's settle for `<array-key, array-key>` for the time being. We can change it later if anything else seems to make more sense.

src/Utilities/CssConcatenator.php
```diff
@@ -150,7 +159,12 @@ private static function getMediaRuleCss(\stdClass $mediaRule): string
      */
     private static function getRuleBlockCss(\stdClass $ruleBlock): string
     {
-        $selectors = \array_keys($ruleBlock->selectorsAsKeys);
-        return \implode(',', $selectors) . '{' . $ruleBlock->declarationsBlock . '}';
+        /** @var array<string, array-key> $selectorsAsKeys */
+        $selectorsAsKeys = $ruleBlock->selectorsAsKeys;
+        /** @var array<array-key, string> $selectors */
+        $selectors = \array_keys($selectorsAsKeys);
```
JakeQZ, line 165: With the type hint now added for `$selectorsAsKeys`, is the one for `$selectors` necessary any more - is Psalm not able to determine it from the signature or `array_keys`, or is it unable to determine that `array_keys` will necessarily return an array of strings when the input array has string keys?

oliverklee: Usually, Psalm complains if deems a type annotation to be redundant. In this case, it has not complained. :-)

JakeQZ: OK, I guess then it's not smart enough to deduce that `array_keys` will return an array whose values will have the same type as the keys in the input array.

JakeQZ (approved):

PR #991, opened 2021-04-09
[CLEANUP] Fix Psalm warnings in AbstractHtmlProcessor

oliverklee:


src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -151,6 +145,20 @@ private function setDomDocument(\DOMDocument $domDocument): void
...
 
+    /**
+     * @return \DOMXPath
+     *
+     * @throws \BadMethodCallException
+     */
+    protected function getXPath(): \DOMXPath
+    {
+        if (!$this->xPath instanceof \DOMXPath) {
+            throw new \BadMethodCallException('$this->xPath has not been initialized yet.', 1617819086);
```
JakeQZ, line 156: Could we use `self::class` in the message (like for `getDomDocument`) rather than `$this` which might be a bit vague?

src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -60,16 +60,16 @@ abstract class AbstractHtmlProcessor
     protected $domDocument = null;
 
     /**
-     * @var \DOMXPath
+     * @var \DOMXPath|null
```
JakeQZ, line 63: Could we use `@var ?\DOMXPath`?  It's a matter of style but the 'nullable' syntax is consistent with method signatures recognized by the PHP interpreter.

JakeQZ (changes requested):
Code changes look good.  But for best forward compatibility (i.e. avoiding potentially having to make breaking changes in future) I think the constuctor should be `protected` and not `final`.  Also a couple of minor issues/comments.

src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -60,16 +60,16 @@ abstract class AbstractHtmlProcessor
...
-    protected $xPath = null;
+    private $xPath = null;
 
     /**
      * The constructor.
      *
      * Please use `::fromHtml` or `::fromDomDocument` instead.
      */
-    private function __construct()
+    final private function __construct()
```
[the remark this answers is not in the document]
oliverklee: We can make it `protected`, but we cannot drop the `final` or drop the constructor altogether. If we do any of these, Psalm (rightly) complains:

```
ERROR: UnsafeInstantiation - src/HtmlProcessor/AbstractHtmlProcessor.php:106:21 - Cannot safely instantiate class Pelago\Emogrifier\HtmlProcessor\AbstractHtmlProcessor with "new static" as its constructor might change in child classes (see https://psalm.dev/229)
        $instance = new static();
```

src/HtmlProcessor/AbstractHtmlProcessor.php
oliverklee: Yes, this seems to work for Psalm. Changed.

src/HtmlProcessor/AbstractHtmlProcessor.php
oliverklee: Yes, we can! Changed and repushed.

src/HtmlProcessor/AbstractHtmlProcessor.php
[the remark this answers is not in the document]
JakeQZ: In that case I think we should use the `@psalm-consistent-constructor` annotation so that child classes can have a constructor if needed, but, if so, Psalm would enforce that it has the same parameters (i.e. no parameters).

I also think the constructor should be `protected` but that's probably beyond the scope of this PR.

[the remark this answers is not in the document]
oliverklee: Okay, changed accordingly.

JakeQZ (approved):

PR #992, opened 2021-04-09
[TASK] Move PHPUnit from PHIVE back to Composer

### LINKED ISSUE #983: Move PHPUnit back from PHIVE to Composer
This would allow tools like Psalm or Rector to also work on the tests (as they need to be able to autoload the PHPUnit classes).

oliverklee:
This allows us to use Psalm and Rector on the tests as well.

Fixes #983


composer.json
```diff
@@ -40,6 +40,7 @@
     },
     "require-dev": {
         "php-parallel-lint/php-parallel-lint": "^1.2.0",
+        "phpunit/phpunit": "^8.5.15",
```
JakeQZ, line 43: Could we have `^8.5.15 || ^9.5.4`?  Since we have already been using the latest 9.x PHPUnit with supported PHP versions, we know it works for us.  That way, we would get whatever benefits it gives when testing against PHP versions it supports.

JakeQZ (changes requested):
I think we can allow PHPUnit 9.x to be installed for supported PHP versions.

composer.json
oliverklee: Done and repushed.

JakeQZ (approved):

composer.json
JakeQZ: This does actually introduce warnings about deprecated assertion methods (they have effectively been renamed), but if we wanted to eliminate that noise we'd have to be restricted to PHPUnit 8 until we drop PHP 7.2.

oliverklee: I'm fine with staying on one major version of PHPUnit at a time.

JakeQZ: I'm not particularly bothered about the noise (at least so far anyway), so I'm fine either way at the moment.

PR #993, opened 2021-04-09
[CLEANUP] Drop superfluous whitespace in a Composer command

oliverklee:


JakeQZ (approved):
I noticed this after I approved the change, but it wasn't bothering me too much :)

PR #996, opened 2021-04-10
[CLEANUP] Be more explicit about type checks in `CssInliner`

oliverklee:


src/CssInliner.php
```diff
@@ -616,6 +619,9 @@ private function getNodesToExclude(): array
                 continue;
             }
             foreach ($matchingNodes as $node) {
+                if (!$node instanceof \DOMElement) {
+                    throw new \UnexpectedValueException('$node is on DOMElement.', 1617975914);
```
JakeQZ, line 623: There's a typo here.  Should be "is not a DOMElement"?  Also, we could give more info in the message by using `$node->getNodePath() ?? '$node'`.

src/CssInliner.php
```diff
@@ -1276,10 +1283,17 @@ protected function addStyleElementToDocument(string $css): void
...
      * @return \DOMElement
+     *
+     * @throws \UnexpectedValueException
      */
     private function getHeadElement(): \DOMElement
     {
-        return $this->domDocument->getElementsByTagName('head')->item(0);
+        $node = $this->getDomDocument()->getElementsByTagName('head')->item(0);
+        if (!$node instanceof \DOMElement) {
+            throw new \UnexpectedValueException('The head is no DOMElement. This should never happen.', 1617923227);
```
JakeQZ, line 1293: Typo.  Should be "is not a DOMElement"?  Also, this exception could occur if there is no `<head>` element (as we have seen in #923 and #957) so the message should be something like "There is no head element or it is not a DOMElement. ..."

JakeQZ (changes requested):
Gets rid of a few Psalm errors :)

I found a couple of typos in exception messages and have additional suggestions for their improvement.

oliverklee:
Oh, I missed your review comments. On it (after cooking some curry …).

src/CssInliner.php
JakeQZ: Actually I think the if the exception occurs it will always be because there is no `<head>` element.  If there is one, it will necessarily be a `DOMElement`.

oliverklee: I've updated the message accordingly.

src/CssInliner.php
oliverklee: Done.

src/CssInliner.php
```diff
@@ -421,9 +421,11 @@ static function (array $m): string {
...
         $nodePath = $node->getNodePath();
-        if (!isset($this->styleAttributesForNodes[$nodePath])) {
-            $this->styleAttributesForNodes[$nodePath] = $this->parseCssDeclarationsBlock($normalizedOriginalStyle);
-            $this->visitedNodes[$nodePath] = $node;
+        if (\is_string($nodePath)) {
+            if (!isset($this->styleAttributesForNodes[$nodePath])) {
+                $this->styleAttributesForNodes[$nodePath] = $this->parseCssDeclarationsBlock($normalizedOriginalStyle);
+                $this->visitedNodes[$nodePath] = $node;
+            }
         }
```
JakeQZ, line 429: Sorry, I just noticed this which I missed first time round.  We could use `&&` rather than two separate `if` statements.

JakeQZ (changes requested):
Sorry, I've just spotted something that I missed first time round.

src/CssInliner.php
oliverklee: Done.

JakeQZ (approved):

PR #997, opened 2021-04-10
[CLEANUP] Make the preg_replace wrapper more specific

oliverklee:
We currently do not need the functionality of working with
array there.


JakeQZ (approved):
Fair enough.

These multi-purpose internal PHP functions do cause some issues with Psalm.  Or perhaps, rather, Psalm highlights the poor design of the internal PHP functions allowing a variety of different parameter types.

PR #998, opened 2021-04-10
[CLEANUP] Improve the type annotations for some `CssInliner` fields

oliverklee:


src/CssInliner.php
```diff
@@ -93,22 +93,22 @@ class CssInliner extends AbstractHtmlProcessor
     ];
 
     /**
-     * @var CssSelectorConverter
+     * @var CssSelectorConverter|null
```
JakeQZ, line 96: Is `@var ?CssSelectorConverter` possible/preferable?

psalm.baseline.xml
```diff
@@ -51,6 +49,10 @@
       <code>string[]</code>
       <code>string[][][]</code>
     </MixedInferredReturnType>
+    <MixedPropertyTypeCoercion occurrences="2">
+      <code>$this-&gt;styleAttributesForNodes</code>
+      <code>$this-&gt;visitedNodes</code>
+    </MixedPropertyTypeCoercion>
```
JakeQZ, line 55: I don't understand why we now have these errors but didn't before.  Though I note the overall error count has gone down :)

JakeQZ (commented):
Looks OK but I have a couple of questions.

src/CssInliner.php
oliverklee: Yes, changed.

psalm.baseline.xml
oliverklee: Now that the field types are more specific, Psalm can recognize cases where the code (or - more probable - the annotations for that code) does not make sure (yet) that these more specific type restrictions are adhered to.

JakeQZ: OK.

JakeQZ (approved):

PR #1000, opened 2021-04-11
[CLEANUP] Use TRegx\DataProvider\DataProviders::cross()

JakeQZ:
Since the library `rawr/cross-data-providers` has been added as a development
dependency (in #959), make use of it where applicable to combine datasets for
tests.  This makes the test data providers clearer and eliminates duplication
of functionality.

### DIFF
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -10,6 +10,7 @@ use Pelago\Emogrifier\Tests\Support\Traits\AssertCss;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\CssSelector\CssSelectorConverter;
 use Symfony\Component\CssSelector\Exception\SyntaxErrorException;
+use TRegx\DataProvider\DataProviders;
 
 /**
  * @covers \Pelago\Emogrifier\CssInliner
@@ -1752,7 +1753,7 @@ final class CssInlinerTest extends TestCase
                         // test with each possible CSS in all three positions
                         || ($cssBefore === $cssBetween && $cssBetween === $cssAfter)
                     ) {
-                        $description = ' with ' . $descriptionBefore . ' before, '
+                        $description = 'with ' . $descriptionBefore . ' before, '
                             . $descriptionBetween . ' between, '
                             . $descriptionAfter . ' after';
                         $datasetsSurroundingCss[$description] = [$cssBefore, $cssBetween, $cssAfter];
@@ -1761,28 +1762,21 @@ final class CssInlinerTest extends TestCase
             }
         }
 
-        $datasets = [];
-        foreach ($datasetsSurroundingCss as $description => $datasetSurroundingCss) {
-            $datasets += [
-                'two media rules' . $description => \array_merge(
-                    ['@media all { p { color: #333; } }', '@media print { p { color: #000; } }'],
-                    $datasetSurroundingCss
-                ),
-                'two rules involving pseudo-components' . $description => \array_merge(
-                    ['a:hover { color: blue; }', 'a:active { color: green; }'],
-                    $datasetSurroundingCss
-                ),
-                'media rule followed by rule involving pseudo-components' . $description => \array_merge(
-                    ['@media screen { p { color: #000; } }', 'a:hover { color: green; }'],
-                    $datasetSurroundingCss
-                ),
-                'rule involving pseudo-components followed by media rule' . $description => \array_merge(
-                    ['a:hover { color: green; }', '@media screen { p { color: #000; } }'],
-                    $datasetSurroundingCss
-                ),
-            ];
-        }
-        return $datasets;
+        return DataProviders::cross(
+            [
+                'two media rules' => ['@media all { p { color: #333; } }', '@media print { p { color: #000; } }'],
+                'two rules involving pseudo-components' => ['a:hover { color: blue; }', 'a:active { color: green; }'],
+                'media rule followed by rule involving pseudo-components' => [
+                    '@media screen { p { color: #000; } }',
+                    'a:hover { color: green; }',
+                ],
+                'rule involving pseudo-components followed by media rule' => [
+                    'a:hover { color: green; }',
+                    '@media screen { p { color: #000; } }',
+                ],
+            ],
+            $datasetsSurroundingCss
+        );
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tests/Unit/CssInlinerTest.php
```diff
@@ -3467,86 +3461,53 @@ public function inlineCssNotCopiesInlinableRuleAfterImportRuleToStyleElement():
      */
     public function provideValidAtRulesWithSurroundingCss(): array
     {
-        $datasets = [];
-
-        foreach (self::BLACK_BOX_AT_RULES as $name => $rule) {
-            $datasets += [
-                'single ' . $name => [
-                    'before' => '',
```
JakeQZ, line 3475: `DataProviders::cross()` throws an exception if the dataset items have string keys, so unfortunately the descriptive keys can't be retained.

oliverklee (approved):
LGTM.

Danon:
Glad you used it <3 If there's any need for help, support, missing functionalities, feel free to contact me directly.

JakeQZ:
Hi @Danon,

Thanks for touching base.

It helps simplify our `TestCase` code by not having to duplicate functionality someone else has already written :)

One feature that might be useful would be a combination comprising just the first row and column of the matrix (extending to more dimensions, so for example with a 3-dimensional matrix it would be the first row and column of the first slice and just the top-left cell of the other slices).  For example, we have one test with 3 parameters each of which can take about 10 different values, which would lead to 1,000 tests with all combinations, and could be expensive.  OTOH, maybe we should split this into 3 different tests, one for each parameter being varied.

Danon:
@JakeQZ Hi, thanks for the kind words.

Yea, I think i know what you mean. Nice way to safe on performance, and you're right, it does scale well to more dimentions.

So, just to know we're clear, if for performance we cross with the first item from other columns, that would mean for 2 sets of ten, it would do 20 runs, instead of 100. And for sets of 20, 10 and 5, it would do 350 runs, instead of 1000. Are we on the same page, or did I misunderstand it? :) 

That would also mean we can scale it, so it takes just the first item from other columns, two items, etc, and that would mean the current implementation is just a special case of that rule where it takes all the items.

JakeQZ:
I was thinking that for minimal coverage, say in the case of a 3x3x3 matrix, we may just want to test 3 edges.  A Rubik's Cube may help illustrate:

Rubik's Cube

We would test the 3 smaller cubes along the edge between the red and the white face, plus 2 additional cubes along the edge between the red and blue faces, and 2 more along the edge between the white and blue faces.

I guess a more general case would involve specifying along which axes you want to test everything, and which you want to test minimally (with the current `cross` implementation testing all axes maximally), but right now I'm struggling to get my head round exactly how that would work...

Another idea I had in conjunction with this was an option to randomly test a few extra combinations.  This would ensure that over time the additional combinations are all tested, but could be a pain to work with if the tests only sometimes fail, and it's tricky to nail the specific combination(s) which are failing.

Though as I said, maybe splitting the test up is a better approach.  But doing so means duplicating the code running the test (even if it is only a few lines).

@oliverklee may have some thoughts on these testing conundrums...

I'll think about it a bit more...

Danon:
@JakeQZ Oh, the analogy with rubik's cube is great.

Thanks for clarification, at first I thought to run the test on whole faces (ignoring the middle, it makes more sense when there are more items), but in your example, that's running only the edges.

> Another idea I had in conjunction with this was an option to randomly test a few extra combinations. This would ensure that over time the additional combinations are all tested, but could be a pain to work with if the tests only sometimes fail, and it's tricky to nail the specific combination(s) which are failing.

I see what you mean, and I do agree that introducing randomisation to unit tests is a bad idea. So you would like to try each value in each data provider at least once (so it's tested); multiple times with other columns (to see if it works with others); once with every item in other column (that's `O(n)`), but you don't need to check every combination with every other combination (so we don't want `O(n^3)` say, for 3 dimensions).

I also understand the struggle, 10 items in a data provider is a relatively small amount; but them them and you have 1000 unit tests; but probably test something that probably 100 or even less tests could test. Perhaps I can take a look into matrix's maths and get some ideas from there.

JakeQZ:
> So you would like to try each value in each data provider at least once (so it's tested); multiple times with other columns (to see if it works with others); once with every item in other column (that's `O(n)`), but you don't need to check every combination with every other combination (so we don't want `O(n^3)` say, for 3 dimensions).

Basically not having to reinveint this wheel: https://github.com/MyIntervals/emogrifier/blob/v5.0.1/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php#L903-L931

But I don't think there's an easy general case between a full `cross` and whether you want faces or edges or some of each to be included, and in which dimensions.

> missing functionalities

On a slightly different topic, I just used your library to knock up a test for something (see https://github.com/MyIntervals/emogrifier/commit/86e49736aad4e65a8dcb8f3fa32e42940c63cad8).

But I had to implement the `dataProvider` providing the `cross`.  Integration into PHPUnit so I could write something like `@dataProvider provideTransparentTagName x provideFlowOrPhrasingContent` could be nice.

Then there's Psalm support - it might be possible to provide an accurate return value using [`@template`](https://psalm.dev/docs/annotating_code/templated_annotations/) but I haven't looked into this in any detail.

tests/Unit/CssInlinerTest.php
Danon: @JakeQZ I'm on it! <3

Danon:
> > So you would like to try each value in each data provider at least once (so it's tested); multiple times with other columns (to see if it works with others); once with every item in other column (that's `O(n)`), but you don't need to check every combination with every other combination (so we don't want `O(n^3)` say, for 3 dimensions).
> 
> Basically not having to reinveint this wheel: https://github.com/MyIntervals/emogrifier/blob/v5.0.1/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php#L903-L931

Oh, nice! Yea, I might use that. I'll still have to reinvent something, so documentation, unit tests, all that good stuff. But the code is a good starting point.

> But I don't think there's an easy general case between a full `cross` and whether you want faces or edges or some of each to be included, and in which dimensions.

I'll probably go for the faces in 3d or more. Using edges is no different than just `array_merge()` of the data providers, so that's not useful to anyone in terms of testing.

> > missing functionalities
> 
> On a slightly different topic, I just used your library to knock up a test for something (see [86e4973](https://github.com/MyIntervals/emogrifier/commit/86e49736aad4e65a8dcb8f3fa32e42940c63cad8)).

Left a comment ;)

> But I had to implement the `dataProvider` providing the `cross`. Integration into PHPUnit so I could write something like `@dataProvider provideTransparentTagName x provideFlowOrPhrasingContent` could be nice.

Yes, that's a really cool idea! We could also use `x` for full cross, `|` for edges and and `*` for faces. Yes, that's a really good point.

> Then there's Psalm support - it might be possible to provide an accurate return value using [`@template`](https://psalm.dev/docs/annotating_code/templated_annotations/) but I haven't looked into this in any detail.

I don't have much experience with Psalm, so I'd have to rely on your expertise.

JakeQZ:
> Using edges is no different than just `array_merge()` of the data providers, so that's not useful to anyone in terms of testing.

The different edges are a different parameter being changed, so cannot be `array_merge`d.  In the [code link](https://github.com/MyIntervals/emogrifier/blob/v5.0.1/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php#L903-L931) in my previous post, how could you replace the nested `for` loops with an `array_merge`?

> I'll probably go for the faces in 3d or more.

With 4D or higher, it's hard to visualize, but I think the possible edge vs face possibilities may increase like in Pascale's triangle.

Danon:
> > Using edges is no different than just `array_merge()` of the data providers, so that's not useful to anyone in terms of testing.
> 
> The different edges are a different parameter being changed, so cannot be `array_merge`d. In the [code link](https://github.com/MyIntervals/emogrifier/blob/v5.0.1/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php#L903-L931) in my previous post, how could you replace the nested `for` loops with an `array_merge`?

Sorry, pehraps I didn't speak my mind correctly. My point was, if we'd go for edges, then typing 3 arrays into the method; would be very simmilar to just joining those 3 arrays (plus, added one single value).

Instead of 
```
DataProviders::edges(
[
  'a', 
  'b',
  'c'
], 
[
  '1', 
  '2',
  '3'
], 
[
  'A',
  'B',
  'C'
]);
```
so we have 9 lines/9 values, but you could achieve the same by going
```
array_merge(
[
  ['a', $y, $z], 
  ['b', $y, $z],
  ['c', $y, $z]
], 
[
  [$x, '1', $z], 
  [$x, '2', $z],
  [$x, '3', $z]
], 
[
  [$x, $y, 'A'],
  [$x, $y, 'B'],
  [$x, $y, 'C']
]
)
```
Where `$x`, `$y` and `$z` is constant. 
So in case of edge-strategy, what the library would do, is just choose an arbitrary `$x`, `$y`, `$z` values, and append it to each element. Basically a linear data provider `O(n)`, and constant arbitrary added value `O(1)`. So yea, that's useful if you want to decrease the duplication slightly, but you won't lower the number of lines used in your data provider that way.

What I mean is, if you want for example 25 items in your edge-strategy data provider, you would have to type 25 elements into your code anyway. So that's a little less useful, you won't test more values that way. It's only use would be to reduce duplication from the code, which is also a good idea.

As you see, the edges-strategy is in fact very performant, because it's really similar to just typing values as regular data providers.

> > I'll probably go for the faces in 3d or more.
> 
> With 4D or higher, it's hard to visualize, but I think the possible edge vs face possibilities may increase like in Pascale's triangle.

I guess it'd be best to let users choose what type of matrix they want to choose, edges, planes or cubes (for 4d, which is hard to visualize). Maybe `DataProviders::edges()`, `DataProviders::faces()`, `DataProviders::cross()`, etc.

JakeQZ:
> > Then there's Psalm support - it might be possible to provide an accurate return value using [`@template`](https://psalm.dev/docs/annotating_code/templated_annotations/) but I haven't looked into this in any detail.
> 
> I don't have much experience with Psalm, so I'd have to rely on your expertise.

Psalm does not currently define a template for `array_merge`.  I think something similar (to what doesn't yet exist) would be needed to be able to precisely specify the resultant type for `cross` when there are mixed element types.  However, something along the lines of the template it has for `array_merge_recursive` might help to some extent - e.g. if the data was all strings, it could then be asserted that the result of `cross` would consist only of strings.

I'm not sure however about adding this kind of type annotation without adding the Psalm static analysis to the project itself to confirm it is actually valid...

JakeQZ:
> so we have 9 lines/9 values, but you could achieve the same by going

I agree in this example there's not a lot more code.  But there are two extra values on each line, and more potential for typos etc. to be missed.  In a real world scenario, the values won't be just 'a', 'b', etc., so that's more potential for mistakes with copied-and-pasted dataset parameters.  It's also possible that the `$x` or `$y` or `$z` parameter already has a data provider which is being used in another test, and which could be re-used.

JakeQZ:
> Psalm ... the template it has for `array_merge_recursive`

FYI this is it:

```php
/**
 * @psalm-template TKey as array-key
 * @psalm-template TValue
 *
 * @param array<TKey, TValue> $arr
 * @param array<TKey, TValue> ...$arr2
 *
 * @return array<TKey, TValue>
 */
```

So it has the contract that the types of the values in the returned array are the same as those in the input arrays.

Danon:
I'm guessing!

Wouldn't it be something like that for 2d data-providers?

```php
/**
 * @psalm-template TKey as array-key
 * @psalm-template TValue1
 * @psalm-template TValue2
 *
 * @param array<TKey, TValue1> $arr
 * @param array<TKey, TValue2> ...$arr2
 *
 * @return array<array<TKey, TValue1, TValue2>>
 */
```

JakeQZ:
> Wouldn't it be something like that for 2d data-providers?

The following appears to work for the `DataProviders::cross` method:

```php
    /**
     * @template TValue
     *
     * @param array<string, array<int, TValue>> ...$dataProviders
     *
     * @return array<string, array<int, TValue>>
     */
```

I'm assuming that the datasets are expected to have `string` keys (though if `int` is possible the `string` for the parameter key can be changed to `array-key`).  They are always returned with `string` keys.  Currently the parameters are expected to have `int` keys but if this is changed then likewise `int` can be changed to `array-key`.

(To allow multiple types for the array values it would need to be `TValue1|TValue2` but it isn't actually necessary to have the additional template variable - a single template variable would assume the union of all the types it is used for.  I think `@template` rather than `@psalm-template` would allow it to work with other static analysis tools such as PHPStan if they support the same syntax - I had merely copied and pasted the DocBlock from the stub internal to Psalm.)

> I guess it'd be best to let users choose what type of matrix they want to choose, edges, planes or cubes (for 4d, which is hard to visualize). Maybe `DataProviders::edges()`, `DataProviders::faces()`, `DataProviders::cross()`, etc.

Thinking about this a bit more, in the 2D case there is only one realistic alternative to cross, which is two adjacent edges.  Most alternatives for the 3D case can be made by combining 2D cases.  If the alternative 2D operator for 2 adjacent edges was `|` (with `x` for `cross`), then
- `A | B | C` would give 3 adjacent edges
- `(A x B) | C` would give a face plus a perpendicular edge
- `(A | B) x C` would give 2 adjacent faces

The one variation not possible is 3 adjacent faces, so parhaps this should be what the new method provides when called with 3 arguments.  By extension, when called with 4 arguments, it should probably provide 4 adjacent cuboids which are 'faces' of the hypercuboid (and so on into the nth dimension).  I think the other variations for 4D would similarly be possible by combining operators in 2D and/or 3D.  Though reordering of parameters may be needed to cover all bases.

Functionally, the method would provide all dataset combinations except those that do not include at least one input dataset which is the first in its list.  Does that make sense?

JakeQZ:
> Functionally, the method would provide all dataset combinations comprising at least one input dataset which is the first in its list.

But maybe in 4D (with 4 input datasets), yet another variation might be obtained by selecting dataset combinations comprising at least two input datasets which are the first in their list.  Perhaps the general case is with a parameter specifying the minimum number of datasets which must be the first in their list, to filter all of the combinations.  `cross` is then a specific case with this parameter being zero.

PR #1001, opened 2021-04-11
[CLEANUP] Always use anonymous functions for callables

oliverklee:
This makes it possible to have type declarations and annotations,
both helping static code analysis as well as making the code easier
to understand.

Also make the type annotations of the methods called from the
anonymous functions more specific.

### DIFF
--- a/psalm.baseline.xml
+++ b/psalm.baseline.xml
@@ -7,7 +7,8 @@
     <DocblockTypeContradiction occurrences="1">
       <code>$this-&gt;matchingUninlinableCssRules === null</code>
     </DocblockTypeContradiction>
-    <InvalidArgument occurrences="1">
+    <InvalidArgument occurrences="3">
+      <code>$cssRule</code>
       <code>$cssRule</code>
     </InvalidArgument>
     <InvalidOperand occurrences="1">
@@ -16,12 +17,8 @@
     <LessSpecificReturnStatement occurrences="1">
       <code>$excludedNodes</code>
     </LessSpecificReturnStatement>
-    <MixedArgument occurrences="1">
-      <code>$m[0]</code>
-    </MixedArgument>
     <MixedArgumentTypeCoercion occurrences="2">
       <code>$attributeName</code>
-      <code>\usort($cssRules['inlinable'], [$this, 'sortBySelectorPrecedence'])</code>
     </MixedArgumentTypeCoercion>
     <MixedArrayAccess occurrences="4">
       <code>$this-&gt;caches[self::CACHE_KEY_COMBINED_STYLES][$cacheKey]</code>
--- a/psalm.baseline.xml
+++ b/psalm.baseline.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="4.x-dev@">
   <file src="src/CssInliner.php">
-    <ArgumentTypeCoercion occurrences="1">
+    <ArgumentTypeCoercion occurrences="2">
       <code>$node</code>
     </ArgumentTypeCoercion>
     <DocblockTypeContradiction occurrences="1">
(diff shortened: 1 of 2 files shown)


psalm.baseline.xml
```diff
@@ -8,7 +8,8 @@
       <code>$this-&gt;cssSelectorConverter === null</code>
       <code>$this-&gt;matchingUninlinableCssRules === null</code>
     </DocblockTypeContradiction>
-    <InvalidArgument occurrences="1">
+    <InvalidArgument occurrences="3">
+      <code>$cssRule</code>
       <code>$cssRule</code>
     </InvalidArgument>
```
JakeQZ, line 13: `occurrences="3"` but there are only 2 in the list - ?

psalm.baseline.xml
```diff
@@ -16,12 +17,8 @@
     <LessSpecificReturnStatement occurrences="1">
       <code>$excludedNodes</code>
     </LessSpecificReturnStatement>
-    <MixedArgument occurrences="1">
-      <code>$m[0]</code>
-    </MixedArgument>
     <MixedArgumentTypeCoercion occurrences="2">
       <code>$attributeName</code>
-      <code>\usort($cssRules['inlinable'], [$this, 'sortBySelectorPrecedence'])</code>
```
JakeQZ, line 24: The `occurrences="2"` attribute has not changed but now there's only one occurrence.

src/CssInliner.php
```diff
@@ -690,7 +691,16 @@ private function parseCssRules(string $css): array
             }
         }
 
-        \usort($cssRules['inlinable'], [$this, 'sortBySelectorPrecedence']);
+        \usort(
+            $cssRules['inlinable'],
+            /**
+             * @param array<string, string> $a
+             * @param array<string, string> $b
+             */
```
JakeQZ, line 699: I think the existing DocBlock type hint (`string[]`) for `sortBySelectorPrecedence` parameters was incorrect.  In full they should be `array{media: string, selector: string, hasUnmatchablePseudo: bool, declarationsBlock: string, line: int}` but probably only the keys used by the function need to be specified, i.e. `array{selector: string, line: int}`.

src/CssInliner.php
```diff
@@ -745,8 +755,8 @@ private function selectorPartHasUnsupportedOfTypePseudoClass(string $selectorPar
     }
 
     /**
-     * @param string[] $a
-     * @param string[] $b
+     * @param array<string, string> $a
+     * @param array<string, string> $b
```
JakeQZ, line 759: Ditto.

JakeQZ (changes requested):
I think an existing DocBlock was wrong, so should be corrected with this change.  Also some `occurrences` attributes in `psalm.baseline.xml` now seem to mismatch with the actual count of `<code>` elements.

psalm.baseline.xml
oliverklee: Yes, that's strange indeed. Still, this is what Psalm generates (for whatever reason), and I'd like to keep it so we don't need to always change this manually when re-generating the file.

psalm.baseline.xml
oliverklee: Ditto. I've re-generated the file.

Thinking of this: I'll try if deleting if before re-generating it helps.

psalm.baseline.xml
oliverklee: No, this doesn'tt change anything.

psalm.baseline.xml
oliverklee: No, this doesn't change anything.

src/CssInliner.php
oliverklee: Fixed and repushed (both instances).

psalm.baseline.xml
JakeQZ: OK.

psalm.baseline.xml
JakeQZ: Weird.  Looks like on the regeneration this one has been corrected, but different incorrect one has now arisen.

JakeQZ (approved):

PR #1002, opened 2021-04-11
[CLEANUP] Remove workaround for PHPUnit 7

oliverklee:
PHPUnit's `Constraint` class now does not have a constructor anymore,
and hence we do not need to call it anymore in the constructors of
classes inheriting from `Constraint`.

### DIFF
--- a/tests/Support/Constraint/CssConstraint.php
+++ b/tests/Support/Constraint/CssConstraint.php
@@ -95,14 +95,4 @@ abstract class CssConstraint extends Constraint
 
         return $regularExpressionEquivalent;
     }
-
-    /**
-     * Invokes the parent class's constructor if there is one.
-     */
-    public function __construct()
-    {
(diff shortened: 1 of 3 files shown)


JakeQZ (commented):
Actually I would be in favour of keeping this as it is.  If a later version of PHPUnit adds back the constructor and it needs to do something, the behaviour will not be as intended.  Most likely there'll be an obvious error, but if the consequence happened to be that tests could pass when they should fail but there was no other error, we could end up introducing bugs that the tests were designed to catch but did not catch.

I would also argue that the removal of the constructor entirely (rather than changing it to an empty method body) from PHPUnit was a mistake.  Or put another way, this workaround is for an issue introduced with PHPUnit 8.x.

Really, the issue stems from PHP itself: neither are parent constructors implicitly called, nor is there an implicit default constructor.

oliverklee:
Okay, then let's keep it. Closing.

PR #1004, opened 2021-04-11
[CLEANUP] Fix Psalm warnings in `CssToAttributeConverter`

oliverklee:


src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -164,11 +165,12 @@ private function mapSimpleCssProperty(string $property, string $value, \DOMEleme
         }
 
         $mapping = $this->cssToHtmlMap[$property];
-        $nodesMatch = !isset($mapping['nodes']) || \in_array($node->nodeName, $mapping['nodes'], true);
-        $valuesMatch = !isset($mapping['values']) || \in_array($value, $mapping['values'], true);
+        $nodesMatch = !isset($mapping['nodes']) || \in_array($node->nodeName, (array)$mapping['nodes'], true);
+        $valuesMatch = !isset($mapping['values']) || \in_array($value, (array)$mapping['values'], true);
```
oliverklee, line 169: Oh, I think I'll use `is_array` checks instead. Hold on.

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -20,7 +20,7 @@ class CssToAttributeConverter extends AbstractHtmlProcessor
      * only for certain values, the mapping is an object with a whitelist
      * of nodes and values.
      *
-     * @var mixed[][]
+     * @var array<string, array<string|string[]>>
```
JakeQZ, line 23: Could we be more explicit in the type hint for `$cssToHtmlMap`, e.g. `array<string, array{attribute: string, nodes: ?array<int, string>, values: ?array<int, string>}>`?  (I'm guessing that the nullable operator would work and be needed for optional elements.)

Failing that, the (`string`) key type of the inner array could be specified.

src/HtmlProcessor/CssToAttributeConverter.php
JakeQZ, line 169: If we can be more explicit in the type hint for `$cssToHtmlMap`, the cast wouldn't be needed - ?

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -164,11 +165,12 @@ private function mapSimpleCssProperty(string $property, string $value, \DOMEleme
...
         $mapping = $this->cssToHtmlMap[$property];
-        $nodesMatch = !isset($mapping['nodes']) || \in_array($node->nodeName, $mapping['nodes'], true);
-        $valuesMatch = !isset($mapping['values']) || \in_array($value, $mapping['values'], true);
+        $nodesMatch = !isset($mapping['nodes']) || \in_array($node->nodeName, (array)$mapping['nodes'], true);
+        $valuesMatch = !isset($mapping['values']) || \in_array($value, (array)$mapping['values'], true);
         $canBeMapped = $nodesMatch && $valuesMatch;
         if ($canBeMapped) {
-            $node->setAttribute($mapping['attribute'], $value);
+            $attributeMapping = \is_string($mapping['attribute']) ? $mapping['attribute'] : '';
+            $node->setAttribute($attributeMapping, $value);
```
JakeQZ, line 173: If we can be more explicit in the type hint for `$cssToHtmlMap`, the `is_string` test wouldn't be needed - ?

JakeQZ (changes requested):
I'm wondering if the type hint for `$cssToHtmlMap` could be more specific - ?

JakeQZ:
Also just noticed the class name is wrong in the commit / PR title.

oliverklee:
I've now moved the changed to `HtmlPruner` to its own PR: #1011

src/HtmlProcessor/CssToAttributeConverter.php
oliverklee: Done.

src/HtmlProcessor/CssToAttributeConverter.php
oliverklee: Done.

src/HtmlProcessor/CssToAttributeConverter.php
oliverklee: Done.

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -20,7 +20,7 @@ class CssToAttributeConverter extends AbstractHtmlProcessor
      * only for certain values, the mapping is an object with a whitelist
      * of nodes and values.
      *
-     * @var mixed[][]
+     * @var array<string, array{attribute: string, nodes?: array<int, string>, values?: array<int, string>}>
```
JakeQZ, line 23: Oh, so the `?` goes *here* for an array element which *must either exist with the given type or not exist at all* (i.e. cannot be a `null` entry)?

Where do you get this information from?  I only found out about the `array{key: type}` syntax by seeing it used elsewhere and/or in the Psalm error report.  I've said it before and I'll say it again: "Documentation!" 😉

oliverklee: I've found the contents of the Psalm documentation to be quite good, and its navigation to be an utter mess. I usually google what I need to find.

In this case, googling for "psalm arrays" lead me to this page explaining the syntax:
https://psalm.dev/docs/annotating_code/type_syntax/array_types/

JakeQZ (approved):
Neat.

src/HtmlProcessor/CssToAttributeConverter.php
JakeQZ: Actually the documenation is quite good.  It's just hard to navigate and search.  But so is the internet as a whole.  Often when I put "Psalm" in the search box, a lot of the results are about religion, even if I include "PHP".  But I think including `site:psalm.dev` in the search box is the way forward.  Thanks for the pointer :)

JakeQZ: Sorry, didn't properly read the first paragraph of your comment before replying.  But it seems we agree on both aspects :)

I prefer to use DuckDuckGo and fall back to Google if I can't find what I'm looking for with the former.  With DDG (and presumably also Bing, etc.), the top result for "psalm arrays" is "Topical Bible: Arrays - Arrays (1 Occurrence) Esther 6:8 let the royal apparel be brought with which the king arrayshimself, and the horse that the king rides upon, and on the head of which the royal crown is set;"  I won't post the link, but even by posting the snippet I am probably further confusing the AIs.

PR #1005, opened 2021-04-12
[CLEANUP] Add type hints for stdClass object properties

JakeQZ:
Psalm supports `object{}` type hints to provide type hints for the properties
of a `stdClass` object.  Using these in the DocBlocks allows various type hints
and (redundant) casts within the code to be eliminated.

PHPCS however does not currently understand them, so some DocBlock sniffs are
locally disabled for now.

### DIFF
--- a/src/Utilities/CssConcatenator.php
+++ b/src/Utilities/CssConcatenator.php
@@ -43,13 +43,19 @@ class CssConcatenator
      * Array of media rules in order.  Each element is an object with the following properties:
      * - string `media` - The media query string, e.g. "@media screen and (max-width:639px)", or an empty string for
      *   rules not within a media query block;
-     * - \stdClass[] `ruleBlocks` - Array of rule blocks in order, where each element is an object with the following
+     * - object[] `ruleBlocks` - Array of rule blocks in order, where each element is an object with the following
      *   properties:
      *   - mixed[] `selectorsAsKeys` - Array whose keys are selectors for the rule block (values are of no
      *     significance);
      *   - string `declarationsBlock` - The property declarations, e.g. "margin-top: 0.5em; padding: 0".
      *
-     * @var array<int, \stdClass>
+     * @var array<int, object{
+     *   media: string,
+     *   ruleBlocks: array<int, object{
+     *     selectorsAsKeys: array<string, array-key>,
+     *     declarationsBlock: string
+     *   }>
+     * }>
      */
     private $mediaRules = [];
 
@@ -66,25 +72,21 @@ class CssConcatenator
         $selectorsAsKeys = \array_flip($selectors);
 
         $mediaRule = $this->getOrCreateMediaRuleToAppendTo($media);
-        /** @var array<int, \stdClass> $ruleBlocks */
         $ruleBlocks = $mediaRule->ruleBlocks;
         $lastRuleBlock = \end($ruleBlocks);
 
-        $hasSameDeclarationsAsLastRule = $lastRuleBlock instanceof \stdClass
+        $hasSameDeclarationsAsLastRule = \is_object($lastRuleBlock)
             && $declarationsBlock === $lastRuleBlock->declarationsBlock;
         if ($hasSameDeclarationsAsLastRule) {
-            /** @var array<string, array-key> $lastRuleBlock->selectorsAsKeys */
             $lastRuleBlock->selectorsAsKeys += $selectorsAsKeys;
         } else {
-            /** @var array<string, array-key> $lastRuleBlockSelectors */
-            $lastRuleBlockSelectors = $lastRuleBlock instanceof \stdClass ? $lastRuleBlock->selectorsAsKeys : [];
-            $hasSameSelectorsAsLastRule = $lastRuleBlock instanceof \stdClass
+            $lastRuleBlockSelectors = \is_object($lastRuleBlock) ? $lastRuleBlock->selectorsAsKeys : [];
+            $hasSameSelectorsAsLastRule = \is_object($lastRuleBlock)
                 && self::hasEquivalentSelectors($selectorsAsKeys, $lastRuleBlockSelectors);
             if ($hasSameSelectorsAsLastRule) {
-                $lastDeclarationsBlockWithoutSemicolon = \rtrim(\rtrim((string)$lastRuleBlock->declarationsBlock), ';');
+                $lastDeclarationsBlockWithoutSemicolon = \rtrim(\rtrim($lastRuleBlock->declarationsBlock), ';');
                 $lastRuleBlock->declarationsBlock = $lastDeclarationsBlockWithoutSemicolon . ';' . $declarationsBlock;
             } else {
-                /** @var array<array-key, \stdClass> $mediaRule->ruleBlocks */
                 $mediaRule->ruleBlocks[] = (object)\compact('selectorsAsKeys', 'declarationsBlock');
             }
         }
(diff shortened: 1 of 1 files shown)


JakeQZ:
It occurred to me after reviewing changes here yesterday that it ought to be possible to use `object{...}` type hints in the same way as `array{...}` type hints, and indeed it is.  Some useful links at https://github.com/phpstan/phpstan/issues/2923

src/Utilities/CssConcatenator.php
```diff
@@ -134,36 +142,45 @@ private static function hasEquivalentSelectors(array $selectorsAsKeys1, array $s
...
+     *          ruleBlocks: array<int, object{
+     *            selectorsAsKeys: array<string, array-key>,
+     *            declarationsBlock: string
+     *          }>
+     *        } $mediaRule
      *
      * @return string CSS for the media rule.
      */
-    private static function getMediaRuleCss(\stdClass $mediaRule): string
+    private static function getMediaRuleCss(object $mediaRule): string
```
oliverklee, line 160: In this case, we'd like to limit this to `stdClass`, and not allow instance of any class, don't we?

oliverklee (changes requested):
Does this also make any difference in the generated Psalm baseline?

src/Utilities/CssConcatenator.php
JakeQZ: If the parameter is `stdClass`, Psalm complains that it doesn't match `object` in the DocBlock.  I'll see if it's possible to specify that it's both a `stdClass` and what its properties are...

src/Utilities/CssConcatenator.php
```diff
@@ -134,36 +142,45 @@ private static function hasEquivalentSelectors(array $selectorsAsKeys1, array $s
             && \count($selectorsAsKeys1) === \count($selectorsAsKeys1 + $selectorsAsKeys2);
     }
 
+    // phpcs:disable Squiz.Commenting.FunctionComment.MissingParamTag
+    // phpcs:disable Squiz.Commenting.FunctionComment.MissingParamName
+    // phpcs:disable Squiz.Commenting.FunctionComment.IncorrectTypeHint
```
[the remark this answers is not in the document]
oliverklee: Maybe we can split this: one (small) PR to disable the corresponding sniffs, then merge that PR, and then set the type annotations here after.

JakeQZ:
> Does this also make any difference in the generated Psalm baseline?

No, because all the warnings for the file were already eliminated by #988.

src/Utilities/CssConcatenator.php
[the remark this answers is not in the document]
JakeQZ: Do we want to disable all of the `Squiz.Commenting.FunctionComment` sniffs (we already have a lot of `exclude`s for the group), or just the additional ones that cause issues here?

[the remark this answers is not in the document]
oliverklee: Let's only disable those that are in the way (assuming that something will still be left from that sniff).

src/Utilities/CssConcatenator.php
```diff
@@ -66,25 +72,21 @@ public function append(array $selectors, string $declarationsBlock, string $medi
...
             && $declarationsBlock === $lastRuleBlock->declarationsBlock;
         if ($hasSameDeclarationsAsLastRule) {
-            /** @var array<string, array-key> $lastRuleBlock->selectorsAsKeys */
             $lastRuleBlock->selectorsAsKeys += $selectorsAsKeys;
         } else {
-            /** @var array<string, array-key> $lastRuleBlockSelectors */
-            $lastRuleBlockSelectors = $lastRuleBlock instanceof \stdClass ? $lastRuleBlock->selectorsAsKeys : [];
-            $hasSameSelectorsAsLastRule = $lastRuleBlock instanceof \stdClass
+            $lastRuleBlockSelectors = \is_object($lastRuleBlock) ? $lastRuleBlock->selectorsAsKeys : [];
+            $hasSameSelectorsAsLastRule = \is_object($lastRuleBlock)
```
oliverklee, line 84: Please keep the `instanceof` checks here as those are more specific than `is_object`.

src/Utilities/CssConcatenator.php
```diff
@@ -66,25 +72,21 @@ public function append(array $selectors, string $declarationsBlock, string $medi
         $selectorsAsKeys = \array_flip($selectors);
 
         $mediaRule = $this->getOrCreateMediaRuleToAppendTo($media);
-        /** @var array<int, \stdClass> $ruleBlocks */
         $ruleBlocks = $mediaRule->ruleBlocks;
         $lastRuleBlock = \end($ruleBlocks);
 
-        $hasSameDeclarationsAsLastRule = $lastRuleBlock instanceof \stdClass
+        $hasSameDeclarationsAsLastRule = \is_object($lastRuleBlock)
```
oliverklee, line 78: Please let's keep the `instanceof` here, too.

src/Utilities/CssConcatenator.php
```diff
@@ -102,12 +104,18 @@ public function getCss(): string
...
+     *     declarationsBlock: string
+     *   }>
+     * }
      */
-    private function getOrCreateMediaRuleToAppendTo(string $media): \stdClass
+    private function getOrCreateMediaRuleToAppendTo(string $media): object
     {
         $lastMediaRule = \end($this->mediaRules);
-        if ($lastMediaRule instanceof \stdClass && $media === $lastMediaRule->media) {
+        if (\is_object($lastMediaRule) && $media === $lastMediaRule->media) {
```
oliverklee, line 118: Ditto.

oliverklee (changes requested):

src/Utilities/CssConcatenator.php
JakeQZ: Again, Psalm complained about `\stdClass` not being the same as `object`.  But if `\stdClass` can be used in the type annotation, these can be reverted.

src/Utilities/CssConcatenator.php
JakeQZ: Unfortunately, Psalm doesn't recognize `\stdClass{property: value-type}` - it gives an `InvalidDocBlock` error.

src/Utilities/CssConcatenator.php
JakeQZ: Unfortunately, Psalm doesn't recognize `\stdClass{property: value-type}`, but if these use `instanceof \stdClass` it seems to forget all of the type information provided in the DocBlock and go on to report errors like

> ERROR: MixedOperand - src/Utilities/CssConcatenator.php:81:13 - Left operand cannot be mixed (see https://psalm.dev/059)

whereever the properties are accessed in a code path that would be executed in the case that `instanceof \stdClass` is true.

However, I'm seeing many more Psalm errors than you are (I have no idea why, I am using the same Psalm version), so I'll see what happens on the server...

src/Utilities/CssConcatenator.php
JakeQZ: This one can actually be retained using `instanceof \stdClass`, presumably because Psalm does not need to be able to determine the type of `$lastMediaRule->media` because it is only used as an operand to `===`.

JakeQZ:
It does not seem possible to retain the explicit use of `\stdClass` in the PHP code without also retaining most if not all the type hints that were added in #988, which would seem to defeat the point of adding the property type hints in the DocBlocks.

So I'm not sure how to proceed from here.  @oliverklee, WDYT?

oliverklee:
Then's lets to what we need to make Psalm happy, i.e., I'm fine with having `object` instead of `stdClass` where needed.

JakeQZ:
> 
> 
> Then's lets to what we need to make Psalm happy, i.e., I'm fine with having `object` instead of `stdClass` where needed.

OK.  I have removed the local disabling of sniffs, but otherwise not changed anything since the original commit (and reverted the experimental commit to see if Psalm would be happy with `stdClass` on the build server).  For consistency, there didn't seem to be much point reverting the one instance where Psalm was OK with `instanceof \stdClass` given it wasn't in any of the others.

oliverklee (approved):

PR #1008, opened 2021-04-13
[FEATURE] Set align attribute for <th> elements

JakeQZ:
Since this is set for `<td>` elements, it makes sense to also set it for `<th>`
elements, as both represent table cells.  Although these attributes are
obsolete, there may still be some legacy email clients (e.g. older versions of
Outlook) that need it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+- Set `align` attribute of `<th>` elements with `CssToAttributeConverter` (#1008)
 
 ### Changed
 - Also check the unit test code with Psalm (#1003)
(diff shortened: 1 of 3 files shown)


JakeQZ:
[w3resource](https://www.w3resource.com/html/attributes/html-align-attribute.php) suggests this attribute can be set on any of the table content elements (though seems to omit `<caption>` from its list).  However, for `<col>` and `<colgroup>` it is not analogous to a CSS property, because the elements are not ancestors of the table cell elements.  I'm unsure whether it is worthwhile applying the attribute to `<tbody>`, `<thead>`, `<tfoot>` and `<tr>` for it to be 'inherited'.  I'm also unsure about applying it to `<caption>` as that was omitted from the <cite>w3resource</cite> list.  But at a minimum it seems logical that it should be applied to `<th>` as well as `<td>`.

JakeQZ:
@oliverklee, #1004 reminded me this was one of the 'fixes' I'd promised a PR for ages ago (though I think it's an enhancement rather than a fix).  Hopefully there won't be any merge conflicts with that PR.

oliverklee (approved):

PR #1009, opened 2021-04-13
[TASK] Pin PHPUnit to version 8

oliverklee:
This avoid warnings about using PHPUnit features that deprecated in
PHPUnit 9.


JakeQZ (approved):
Yep, the warnings are starting to get annoying ;)

PR #1012, opened 2021-04-13
[CLEANUP] Drop a redundant null initialization of a field in the tests

oliverklee:


JakeQZ:
Psalm doesn't seem to like it :/

oliverklee:
Strange. I'll have a look.

oliverklee:
If Psalm doesn't like this, I won't like this either anymore … closing.

PR #1013, opened 2021-04-13
[CLEANUP] Move the development tools to `.phive/`

oliverklee:
This reduces the number of non-hidden files/directories in the
project root, and it also moves the PHIVE-installed PHARS next to
the PHIVE configuration file.


JakeQZ (commented):
Could there be any risk of a name clash in using PHIVE's own directory?

My 'home' directory has a `.phive` subdirectory, within which are a few files and directories.  Notably, one is called `phars` and is where PHIVE puts the `.phars` it downloads.  So for consistency, on the one hand, the `.phar`s could be placed in a such-named subdirectory.  On the other hand, this might increase the risk of a clash.  But maybe that [the `.phive` subdirectory of the user's home directory] is a completely different kettle of fish.

I'm leaning towards let's go with this, and we can always revisit later if there are any issues or we think of a better arrangement...

oliverklee:
I don't think that there's any risk involved.

As far as I understand, PHIVE will look different places for different files for its configuration, and `.phive/` is one of those directories. So apart from reading that file (and writing it on updates), other files in that directory should not be affected.

JakeQZ (approved):
OK.

Issue #994, opened 2021-04-10
Big architecture rework


oliverklee:
In addition to the small architectural problems (`CssInliner` is too big and does too much, arrays and `stdClass` instead of dedicated classes), I currently see two big problems I'd like to solve:

1. The `HtmlProcessor` family of classes (including `CssInliner`) currently mix two responsibilities:
1.1. changing the HTML (which they should keep doing) 
1.2. parsing and storing the HTML (which I'd like to move to a dedicated class (or family of classes)
2. There currently is no way Emogrifier classes can be dependency-injected (e.g., by the Symfony container), which also means that there is no way to mock them in unit tests. We probably need a class that can emogrify multiple HTML documents per instance to make this usable in real-world scenarios.

The first set of problems is solvable without any (major) breaking changes, I think: We need to move stuff out to other classes. This also makes solving the second big problem easier to solve as the code then is easier understand and modify (and the risk of breaking things is reduced if we have smaller classes with clear contracts).

Maybe we can have a (new) `Emogrifier` (naming still TBD) class that wraps `CssInliner` etc. using the [Facade design pattern](https://en.wikipedia.org/wiki/Facade_pattern). We still need to determine a new, clean interface on how the new class should handle multiple processors while still being able to process multiple HTML documents. Maybe we can have another intermediate class returns by the new `Emogrifier` methods that holds a reference to the artifacts for a single HTML document.

@JakeQZ, @jjriv, @zoliszabo  What do you think?

JakeQZ:
I'll look into the Facade design pattern and have a think...

On the topic of `CssInliner` being too big, this was partly touched upon in #912.  The suggestion there is that we can move the CSS parsing to a separate class, which will ultimately become just a wrapper for the third-party library (and may then be removed or 'unwrapped').  I am thinking of having a class named `Css` (based on what its objects represent, rather than what it does, so not `CssParser`), whose constructor takes a string of CSS, and which has methods to retrieve the parsed content.  Initially, the main method would be `getRuleBlocks` returning arrays in the form used internally by `CssInliner` (i.e. elements with `media`, `selector` and `declarationsBlock`, etc.).  There would probably need to be one or two other methods to retrieve things like `@charset` and `@import` which are passed on by `CssInliner` but have no property declarations to inline.

JakeQZ:
[...]

PR #1014, opened 2021-04-13
[CLEANUP] Move CSS parsing to a separate class

JakeQZ:
Part of #734 and #544.  Relates to #994.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -6,6 +6,7 @@ namespace Pelago\Emogrifier;
 
 use Pelago\Emogrifier\HtmlProcessor\AbstractHtmlProcessor;
 use Pelago\Emogrifier\Utilities\CssConcatenator;
+use Pelago\Emogrifier\Utilities\ParsedCss;
 use Symfony\Component\CssSelector\CssSelectorConverter;
 use Symfony\Component\CssSelector\Exception\ParseException;
 
@@ -28,18 +29,6 @@ class CssInliner extends AbstractHtmlProcessor
      * @var int
      */
     private const CACHE_KEY_COMBINED_STYLES = 2;
-
-    /**
-     * This regular expression pattern will match any uninlinable at-rule with nested statements, along with any
-     * whitespace immediately following.  Currently, any at-rule apart from `@media` is considered uninlinable.  The
-     * first capturing group matches the at sign and identifier (e.g. `@font-face`).  The second capturing group matches
-     * the nested statements along with their enclosing curly brackets (i.e. `{...}`), and via `(?2)` will match deeper
-     * nested blocks recursively.
-     *
-     * @var string
-     */
-    private const UNINLINABLE_AT_RULE_MATCHER
-        = '/(@(?!media\\b)[\\w\\-]++)[^\\{]*+(\\{[^\\{\\}]*+(?:(?2)[^\\{\\}]*+)*+\\})\\s*+/i';
 
     /**
      * Regular expression component matching a static pseudo class in a selector, without the preceding ":",
@@ -148,7 +137,7 @@ class CssInliner extends AbstractHtmlProcessor
 
     /**
      * array of data describing CSS rules which apply to the document but cannot be inlined, in the format returned by
-     * `parseCssRules`
+     * {@see collateCssRules}
      *
      * @var array<array-key, array{
      *          media: string,
@@ -190,17 +179,10 @@ class CssInliner extends AbstractHtmlProcessor
         if ($this->isStyleBlocksParsingEnabled) {
             $combinedCss .= $this->getCssFromAllStyleNodes();
         }
-
-        $cssWithoutComments = $this->removeCssComments($combinedCss);
-        [$cssWithoutCommentsCharsetOrImport, $cssImportRules]
-            = $this->extractImportAndCharsetRules($cssWithoutComments);
-        [$cssWithoutCommentsOrUninlinableAtRules, $cssAtRules]
-            = $this->extractUninlinableCssAtRules($cssWithoutCommentsCharsetOrImport);
-
-        $uninlinableCss = $cssImportRules . $cssAtRules;
+        $parsedCss = new ParsedCss($combinedCss);
 
         $excludedNodes = $this->getNodesToExclude();
-        $cssRules = $this->parseCssRules($cssWithoutCommentsOrUninlinableAtRules);
+        $cssRules = $this->collateCssRules($parsedCss);
         $cssSelectorConverter = $this->getCssSelectorConverter();
         foreach ($cssRules['inlinable'] as $cssRule) {
             try {
@@ -228,7 +210,7 @@ class CssInliner extends AbstractHtmlProcessor
         $this->removeImportantAnnotationFromAllInlineStyles();
 
         $this->determineMatchingUninlinableCssRules($cssRules['uninlinable']);
-        $this->copyUninlinableCssToStyleNode($uninlinableCss);
+        $this->copyUninlinableCssToStyleNode($parsedCss);
 
         return $this;
     }
@@ -522,112 +504,6 @@ class CssInliner extends AbstractHtmlProcessor
     }
 
     /**
-     * Removes comments from the supplied CSS.
-     *
-     * @param string $css
-     *
-     * @return string CSS with the comments removed
-     */
-    private function removeCssComments(string $css): string
-    {
-        return \preg_replace('%/\\*[^*]*+(?:\\*(?!/)[^*]*+)*+\\*/%', '', $css);
-    }
-
-    /**
-     * Extracts `@import` and `@charset` rules from the supplied CSS.  These rules must not be preceded by any other
-     * rules, or they will be ignored.  (From the CSS 2.1 specification: "CSS 2.1 user agents must ignore any '@import'
-     * rule that occurs inside a block or after any non-ignored statement other than an @charset or an @import rule."
-     * Note also that `@charset` is case sensitive whereas `@import` is not.)
-     *
-     * @param string $css CSS with comments removed
-     *
-     * @return array{0: string, 1: string}
-     *         The first element is the CSS with the valid `@import` and `@charset` rules removed.  The second element
-     *         contains a concatenation of the valid `@import` rules, each followed by whatever whitespace followed it
-     *         in the original CSS (so that either unminified or minified formatting is preserved); if there were no
-     *         `@import` rules, it will be an empty string.  The (valid) `@charset` rules are discarded.
-     */
-    private function extractImportAndCharsetRules(string $css): array
-    {
-        $possiblyModifiedCss = $css;
-        $importRules = '';
-
-        while (
-            \preg_match(
-                '/^\\s*+(@((?i)import(?-i)|charset)\\s[^;]++;\\s*+)/',
-                $possiblyModifiedCss,
-                $matches
-            )
-        ) {
-            [$fullMatch, $atRuleAndFollowingWhitespace, $atRuleName] = $matches;
-
-            if (\strtolower($atRuleName) === 'import') {
-                $importRules .= $atRuleAndFollowingWhitespace;
-            }
-
-            $possiblyModifiedCss = \substr($possiblyModifiedCss, \strlen($fullMatch));
-        }
-
-        return [$possiblyModifiedCss, $importRules];
-    }
-
-    /**
-     * Extracts uninlinable at-rules with nested statements (i.e. a block enclosed in curly brackets) from the supplied
-     * CSS.  Currently, any such at-rule apart from `@media` is considered uninlinable.  These rules can be placed
-     * anywhere in the CSS and are not case sensitive.  `@font-face` rules will be checked for validity, though other
-     * at-rules will be assumed to be valid.
-     *
-     * @param string $css CSS with comments, import and charset removed
-     *
-     * @return array{0: string, 1: string}
-     *         The first element is the CSS with the at-rules removed.  The second element contains a concatenation of
-     *         the valid at-rules, each followed by whatever whitespace followed it in the original CSS (so that either
-     *         unminified or minified formatting is preserved); if there were no at-rules, it will be an empty string.
-     */
-    private function extractUninlinableCssAtRules(string $css): array
-    {
-        $possiblyModifiedCss = $css;
-        $atRules = '';
-
-        while (
-            \preg_match(
-                self::UNINLINABLE_AT_RULE_MATCHER,
-                $possiblyModifiedCss,
-                $matches
-            )
-        ) {
-            /** @var array<int, string> $matches */
-            [$fullMatch, $atRuleName] = $matches;
-
-            if ($this->isValidAtRule($atRuleName, $fullMatch)) {
-                $atRules .= $fullMatch;
-            }
-
-            $possiblyModifiedCss = \str_replace($fullMatch, '', $possiblyModifiedCss);
-        }
-
-        return [$possiblyModifiedCss, $atRules];
-    }
-
-    /**
-     * Tests if an at-rule is valid.  Currently only `@font-face` rules are checked for validity; others are assumed to
-     * be valid.
-     *
-     * @param string $atIdentifier name of the at-rule with the preceding at sign
-     * @param string $rule full content of the rule, including the at-identifier
-     *
-     * @return bool
-     */
-    private function isValidAtRule(string $atIdentifier, string $rule): bool
-    {
-        if (\strcasecmp($atIdentifier, '@font-face') === 0) {
-            return \stripos($rule, 'font-family') !== false && \stripos($rule, 'src') !== false;
-        }
-
-        return true;
-    }
-
-    /**
      * Find the nodes that are not to be emogrified.
      *
      * @return array<int, \DOMElement>
@@ -674,9 +550,9 @@ class CssInliner extends AbstractHtmlProcessor
     }
 
     /**
-     * Extracts and parses the individual rules from a CSS string.
+     * Collates the individual rules from a `ParsedCss` object.
      *
-     * @param string $css a string of raw CSS code with comments removed
+     * @param ParsedCss $parsedCss
      *
      * @return array<string, array<array-key, array{
      *           media: string,
@@ -697,9 +573,9 @@ class CssInliner extends AbstractHtmlProcessor
      *           e.g., `color: red; height: 4px;`);
      *         - "line" (the line number, e.g. 42).
      */
-    private function parseCssRules(string $css): array
+    private function collateCssRules(ParsedCss $parsedCss): array
     {
-        $matches = $this->getCssRuleMatches($css);
+        $matches = $parsedCss->getStyleRulesData(\array_keys($this->allowedMediaTypes));
 
         $cssRules = [
             'inlinable' => [],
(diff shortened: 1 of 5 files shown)


src/Utilities/ParsedCss.php
```diff
@@ -0,0 +1,284 @@
...
+     *
+     * @param array<array-key, string> $allowedMediaTypes
+     *
+     * @return array<int, array{media: string, selectors: string, declarations: string}>
+     * Array of string sub-arrays with the following keys:
+     * - "media": the media query string, e.g. "@media screen and (max-width: 480px)", or an empty string if not from an
+     *   `@media` rule;
+     * - "selectors": the CSS selector(s), e.g., "*" or "h1, h2";
+     * - "declarations": the semicolon-separated CSS declarations for that/those selector(s),
+     *   e.g., "color: red; height: 4px;".
```
JakeQZ, line 70: This can be further refactored in due course.

JakeQZ:
This still needs a `TestCase` for the new class, but I'd be interested in your initial comments and feedback (@oliverklee?).  Note that this is just a refactoring: the original code from `CssInliner` has mostly been cut and pasted to the new class, 'warts and all'.

JakeQZ:
This change is analogous to the second commit in #1015, so you can see how it might turn out when we introduce `sabberworm/php-css-parser` to do the 'heavy lifting'.

src/CssInliner.php
```diff
@@ -150,9 +133,15 @@ class CssInliner extends AbstractHtmlProcessor
 
     /**
      * array of data describing CSS rules which apply to the document but cannot be inlined, in the format returned by
-     * `parseCssRules`
+     * {@see collateCssRules}
```
oliverklee, line 136: Why the curly braces around the `@see` annotation?

src/Utilities/ParsedCss.php
```diff
@@ -0,0 +1,284 @@
...
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\Utilities;
+
+/**
+ * Parses and stores a CSS document from a string of CSS, with methods to return the CSS in parts or as data structures.
+ *
+ * @internal
+ */
+class ParsedCss
```
oliverklee, line 12: Thinking about the name: I don't think that the `Parsed` needs to be part of the class name (even if it might make sense for instance variable names to distinguish the parsed CSS and the raw CSS).

oliverklee (changes requested):
I like this refactoring a lot.

src/CssInliner.php
JakeQZ: It's meant to be inline as part of the description, to appear as a link, which is why there's no blank line before it.  It just so happens that the 120 character line length pushes it onto the next line.

src/Utilities/ParsedCss.php
JakeQZ: OK, will change, either to `Css` or `CssDocument`.  Sabberworm uses `CssDocument` which I like, though would mean their FQN would have to be `use`d as something else.

oliverklee: We can `use` it with an alias:

```php
use Vendor\App\Magic as SabberwormMagic;
```

JakeQZ: Yep, that's what I was thinking :)

src/CssInliner.php
```diff
@@ -190,17 +179,10 @@ public function inlineCss(string $css = ''): self
...
         }
-
-        $cssWithoutComments = $this->removeCssComments($combinedCss);
-        [$cssWithoutCommentsCharsetOrImport, $cssImportRules]
-            = $this->extractImportAndCharsetRules($cssWithoutComments);
-        [$cssWithoutCommentsOrUninlinableAtRules, $cssAtRules]
-            = $this->extractUninlinableCssAtRules($cssWithoutCommentsCharsetOrImport);
-
-        $uninlinableCss = $cssImportRules . $cssAtRules;
+        $cssDocument = new CssDocument($combinedCss);
```
oliverklee, line 182: I suggest the name `$parsedCss`.

src/CssInliner.php
```diff
@@ -1305,15 +1072,15 @@ private function removeUnsupportedOfTypePseudoClasses(string $selectorPart): str
      * Applies `$this->matchingUninlinableCssRules` to `$this->domDocument` by placing them as CSS in a `<style>`
      * element.
      *
-     * @param string $uninlinableCss
+     * @param CssDocument $cssDocument
      *        This may contain any `@import` or `@font-face` rules that should precede the CSS placed in the `<style>`
      *        element.  If there are no unlinlinable CSS rules to copy there, a `<style>` element will be created
-     *        containing just `$uninlinableCss`.  `$uninlinableCss` may be an empty string; if it is, and there are no
+     *        containing only the applicable at-rules from `$cssDocument`.  If there are none, and there are also no
      *        unlinlinable CSS rules, an empty `<style>` element will not be created.
```
oliverklee, line 1079: no unlinlinable -> only inlinable

src/Utilities/CssDocument.php
```diff
@@ -0,0 +1,286 @@
...
+     *         `@import` rules, it will be an empty string.  The (valid) `@charset` rules are discarded.
+     */
+    private function extractImportAndCharsetRules(string $css): array
+    {
+        $possiblyModifiedCss = $css;
+        $importRules = '';
+
+        while (
+            \preg_match(
+                '/^\\s*+(@((?i)import(?-i)|charset)\\s[^;]++;\\s*+)/',
```
oliverklee, line 138: Should we move this to a constant? (Maybe then the `preg_match` will also fit in one line.)

src/Utilities/CssDocument.php
```diff
@@ -0,0 +1,286 @@
...
+     *         the valid at-rules, each followed by whatever whitespace followed it in the original CSS (so that either
+     *         unminified or minified formatting is preserved); if there were no at-rules, it will be an empty string.
+     */
+    private function extractNonConditionalAtRules(string $css): array
+    {
+        $possiblyModifiedCss = $css;
+        $atRules = '';
+
+        while (
+            \preg_match(
```
oliverklee, line 177: Will the `preg_match` fit in one line?

src/Utilities/CssDocument.php
```diff
@@ -0,0 +1,286 @@
...
+     *
+     * @return array{0: string, 1: string}
+     *         The first element is the CSS with the at-rules removed.  The second element contains a concatenation of
+     *         the valid at-rules, each followed by whatever whitespace followed it in the original CSS (so that either
+     *         unminified or minified formatting is preserved); if there were no at-rules, it will be an empty string.
+     */
+    private function extractNonConditionalAtRules(string $css): array
+    {
+        $possiblyModifiedCss = $css;
+        $atRules = '';
```
oliverklee, line 174: I propose we put the at rules in a string array and then implode it when it is returned.

src/Utilities/CssDocument.php
```diff
@@ -0,0 +1,286 @@
...
+     * @return array<int, array{css: string, media: string}>
+     */
+    private function splitCssAndMediaQuery(array $allowedMediaTypes): array
+    {
+        $mediaTypesExpression = '';
+        if (!empty($allowedMediaTypes)) {
+            $mediaTypesExpression = '|' . \implode('|', $allowedMediaTypes);
+        }
+
+        $mediaRuleBodyMatcher = '[^{]*+{(?:[^{}]*+{.*})?\\s*+}\\s*+';
```
oliverklee, line 249: Maybe move to a constant?

oliverklee (changes requested):

src/CssInliner.php
JakeQZ: If it's changed to "only inlinable" I think it becomes misleading - the method only deals with uninlinable rules and never copies inlinable rules to the `<style>` element (whether or not there are any).

oliverklee: Apart from the typo in "unlinlinable", isn't "no uninlinable rules" and ""only inlinable rules" semantically the same, but without the double negation?

JakeQZ: "If there are only inlinable CSS rules to copy there" would imply inlinable rules are copied there.  They are not.

oliverklee: Ah, okay, I see.

Then let's please only fix the typo.

src/CssInliner.php
JakeQZ: Renamed the variable (and likewise the parameter of the methods it's passed to).

src/CssInliner.php
JakeQZ: Done.

src/Utilities/CssDocument.php
JakeQZ: Moved to a constant and formatted the `preg_match` as one line.

src/Utilities/CssDocument.php
JakeQZ: Yes.  Reformatted.

src/Utilities/CssDocument.php
JakeQZ: Done.

src/Utilities/CssDocument.php
JakeQZ: Have done now.

JakeQZ:
I've made the changes initially suggested.  I've also finished off writing the `TestCase` so this is ready for full review now - I've cleared the 'draft'/WIP status of the PR.

src/Utilities/CssDocument.php
```diff
@@ -0,0 +1,283 @@
...
+     *         the valid at-rules, each followed by whatever whitespace followed it in the original CSS (so that either
+     *         unminified or minified formatting is preserved); if there were no at-rules, it will be an empty string.
+     */
+    private function extractNonConditionalAtRules(string $css): array
+    {
+        $possiblyModifiedCss = $css;
+        $atRules = [];
+
+        while (\preg_match(self::NON_CONDITIONAL_AT_RULE_PATTERN, $possiblyModifiedCss, $matches)) {
+            [$fullMatch, $atRuleName] = $matches;
```
oliverklee, line 182: Please let's keep the type annotation above this line (copied from `CssInliner`):
```php
** @var array<int, string> $matches */
```

src/Utilities/CssDocument.php
```diff
@@ -0,0 +1,283 @@
...
+     *
+     * @param array<array-key, string> $allowedMediaTypes
+     *
+     * @return array<int, array{css: string, media: string}>
+     */
+    private function splitCssAndMediaQuery(array $allowedMediaTypes): array
+    {
+        $mediaTypesExpression = '';
+        if (!empty($allowedMediaTypes)) {
+            $mediaTypesExpression = '|' . \implode('|', $allowedMediaTypes);
```
oliverklee, line 245: We need to use `preg_quote` here (with the `$delimiter` parameter).

src/Utilities/CssDocument.php
```diff
@@ -0,0 +1,283 @@
...
+
+        $splitCss = [];
+        foreach ($cssSplitForAllowedMediaTypes as $index => $cssPart) {
+            $isMediaRule = $index % 2 !== 0;
+            if ($isMediaRule) {
+                \preg_match('/^([^{]*+){(.*)}[^}]*+$/s', $cssPart, $matches);
+                $splitCss[] = [
+                    'css' => $matches[2],
+                    'media' => $matches[1],
+                ];
```
oliverklee, line 270: Let's compact this array to one line so the method becomes less long (vertically speaking).

src/Utilities/CssDocument.php
```diff
@@ -0,0 +1,283 @@
...
+            } else {
+                $cleanedCss = \trim(\preg_replace($cssCleaningMatchers, '', $cssPart));
+                if ($cleanedCss !== '') {
+                    $splitCss[] = [
```
oliverklee, line 274: Let's compact this array to one line so the method becomes less long (vertically speaking).

tests/Unit/Utilities/CssDocumentTest.php
```diff
@@ -0,0 +1,403 @@
...
+    public function parsesSelector(string $selector): void
+    {
+        $css = $selector . '{ color: green; }';
+        $subject = new CssDocument($css);
+
+        $result = $subject->getStyleRulesData([]);
+
+        self::assertCount(1, $result);
+        $trimmedResultantSelector = \trim($result[0]['selectors']);
+        self::assertSame(\trim($selector), $trimmedResultantSelector);
```
oliverklee, line 41: We the `assertSame` combined with a `trim` multiple times. Maybe move this to a dedicated `assert` function? (Let's keep it here in this test. We can factor it out when we need it in multiple test cases.)

tests/Unit/Utilities/CssDocumentTest.php
```diff
@@ -0,0 +1,403 @@
...
+            'child' => ['p > .classy'],
+            'general sibling' => ['h1 ~ p'],
+            'adjacent sibling' => ['h1 + p'],
+            'list' => ['h1, h2'],
+            'with space after' => ['p '],
+            'with line feed after' => ["p\n"],
+            'with Windows line ending after' => ["p\r\n"],
+            'with TAB after' => ["p\t"],
+            'minified' => ['p>.classy'],
+            'with line feeds within' => ["p\n>\n.classy"],
```
oliverklee, line 79: I propose we split the selectors and the formatting problems into two data providers.

tests/Unit/Utilities/CssDocumentTest.php
```diff
@@ -0,0 +1,403 @@
...
+            'with line feed before' => ["\ncolor: green;"],
+            'with Windows line ending before' => ["\r\ncolor: green;"],
+            'with TAB before' => ["\tcolor: green;"],
+            'with line feed within' => ["color:\ngreen;"],
+            'with line feed after' => ["color: green;\n"],
+            'multiple' => ['color: green; text-decoration: underline;'],
+            'multiple (minified)' => ['color:green;text-decoration:underline'],
+            'multiple (separated by line feed)' => ["color: green;\ntext-decoration: underline;"],
+            'multiple (separated by Windows line ending)' => ["color: green;\r\ntext-decoration: underline;"],
+            'multiple (separated by TAB)' => ["color: green;\ttext-decoration: underline;"],
```
oliverklee, line 132: Ditto.

tests/Unit/Utilities/CssDocumentTest.php
```diff
@@ -0,0 +1,403 @@
...
+    /**
+     * @test
+     *
+     * @param string $mediaQuery
+     * @param string $optionalSeparator
+     * @param string $separator
+     *
+     * @dataProvider provideMediaQueryAndSeparators
+     */
+    public function parsesMediaRule(string $mediaQuery, string $optionalSeparator = '', string $separator = ' '): void
```
oliverklee, line 145: I propose we split this into two tests and two data providers: one for different types, and one for the formatting issues. Then we don't need any parameters to be optional anymore.

tests/Unit/Utilities/CssDocumentTest.php
```diff
@@ -0,0 +1,403 @@
...
+            'Windows line ending' => ["\r\n"],
+            'TAB' => ["\r\n"],
+            'non-conditional at-rule (valid `@font-face`)' => [self::VALID_AT_FONT_FACE_RULE],
+            'comment' => ['/* Test */'],
+            'commented-out style rule' => ['/* p { color: red; } */'],
+            'space before' => ['', ' '],
+            'line feed before' => ['', "\n"],
+            'Windows line ending before' => ['', "\r\n"],
+            'TAB before' => ['', "\r\n"],
+            'non-conditional at-rule (`@charset`) before' => ['', '@charset "UTF-8";'],
```
oliverklee, line 256: Please let's split this up as well.

tests/Unit/Utilities/CssDocumentTest.php
```diff
@@ -0,0 +1,403 @@
...
+            'type alone' => ['tv'],
+            'type with `only`' => ['only tv'],
+            'type and plain value' => ['tv and (max-width: 480px)'],
+            'type and plain value with `not`' => ['tv and not (max-width: 480px)'],
+            'type and plain values with `and`' => ['tv and (min-width: 320px) and (max-width: 480px)'],
+            'type and range (singly bounded)' => ['tv and (height > 600px)'],
+            'type and range (fully bounded)' => ['tv and (400px <= width <= 700px)'],
+            'with line feed before' => ["\ntv"],
+            'with Windows line ending before' => ["\r\ntv"],
+            'with TAB before' => ["\ttv"],
```
oliverklee, line 214: Maybe split this up as well?

oliverklee (changes requested):
Already looks very good! I've added my comments.

tests/Unit/Utilities/CssDocumentTest.php
```diff
@@ -0,0 +1,403 @@
...
+     *
+     * @dataProvider provideValidNonConditionalAtRule
+     */
+    public function rendersValidNonConditionalAtRule(string $atRuleCss, string $cssBefore = ''): void
+    {
+        $subject = new CssDocument($cssBefore . $atRuleCss);
+
+        $result = $subject->renderNonConditionalAtRules();
+
+        self::assertStringContainsString(\trim($atRuleCss), $result);
```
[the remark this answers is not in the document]
JakeQZ: It was just the multiline `@font-face` rule that needed trimming due to the way it was formatted in the code.  I've changed the constant `VALID_AT_FONT_FACE_RULE` it comes from so it is constructed by concatenating strings so as to ensure consistent formatting and no surrounding whitespace.  And removed the `trim` here.

src/Utilities/CssDocument.php
JakeQZ: I think I had cut cut and pasted the code before this was added in.  I've restored it.  Interesting that Psalm neither complains that it is unnecessary nor if it is absent.

src/Utilities/CssDocument.php
JakeQZ: That latent bug has been around for a long time, though unlikely to arise in practice as media types are pretty much restricted to alphanumeric characters (I think).  Fixed.

src/Utilities/CssDocument.php
JakeQZ: Done.

src/Utilities/CssDocument.php
JakeQZ: Done.

tests/Unit/Utilities/CssDocumentTest.php
JakeQZ: Added a static method `assertSameTrimmed`.

There's also another assertion using `trim` but with an `assertThat`, which can't use this - it would need a `Constraint` child class to provide the functionality (along with a `TestCase` for any such new class).  I propose not implementing such right now, as the requirement may change if/when the Sabberworm library is introduced, as this is liable to result in additional whitespace differences which need to be accounted for.

tests/Unit/Utilities/CssDocumentTest.php
JakeQZ: Done.

tests/Unit/Utilities/CssDocumentTest.php
JakeQZ: Done.

tests/Unit/Utilities/CssDocumentTest.php
JakeQZ: Good idea.  Turns out the test with whitespace parameters doesn't need the media query parameter either.

tests/Unit/Utilities/CssDocumentTest.php
JakeQZ: Splitting this one up works well, as the same data provider can be used for both tests.

tests/Unit/Utilities/CssDocumentTest.php
JakeQZ: Done.

JakeQZ:
Have commited changes as suggested.

oliverklee (approved):
Beautiful!

PR #1015, opened 2021-04-13
[TASK] Use sabberworm/php-css-parser to parse the CSS

JakeQZ:


oliverklee:
`sabberworm/php-css-parser` has not been very well-maintained lately. We might need to help out there (or choose a different library).

JakeQZ:
> 
> 
> `sabberworm/php-css-parser` has not been very well-maintained lately. We might need to help out there (or choose a different library).

Last release was 10 months ago.  Last commit was 5 months ago.  But I'm not aware of any new CSS syntaxes being introduced, so there may be no pressing reason to update anything particularly regularly.  Of course new PHP 7 or Psalm-compatible type hints would be nice.

Inevitably the new tests are failing since rebase due to CSS whitespace differences (and `@import "foo.css"` being rendered as `@import url("foo.css")`.

oliverklee:
I'm mostly worried about situations where we need a particular bug fix or a compatibility fix for newer PHP versions. (And I'm hopeful that we can contribute to the project, and that the maintainer will accept our changes.)

oliverklee:
In the meantime, it looks like the maintainers of the Sabberworm CSS parser review and merge new PRs (slowly but steadily). We might need to take care of updating old PRs and fixing old bugs ourselves, though (including bugs reported by Nikita Popov from the PHP core team).

bytestream:
I think adding Sabberworm CSS parser is a good idea. With respect to other libraries, the only other one I'm aware of is https://github.com/Cerdic/CSSTidy but I think you'll find similar issues with this one.

JakeQZ:
> With respect to other libraries, the only other one I'm aware of is https://github.com/Cerdic/CSSTidy

Thanks, I was not aware of that library.  I think it would certainly be worth looking into before committing more heavily to the other one - e.g. starting to use its classes further downstream (which this PR doesn't go as far as).

JakeQZ:
> > https://github.com/Cerdic/CSSTidy
> 
> I think it would certainly be worth looking into before committing more heavily to the other one.

I should have been getting on with something else, but thought it worth checking out how CSSTidy might pan out if we were to decide to go with that instead.  I've created a draft PR, #1076, for that alternative, so you can see what the code changes look like and what tests would have to be modified or disabled.

The advantages of one vs the other appear to be as follows:

CSSTidy
- Seems possibly able to combine shorthand and longhand properties;
- Has potentially revealed a bug in our code.

Sabberworm
- Supports PHP 8;
- Provides classes we could use downstream;
- Rejects invalid property names;
- Rejects invalid `@import` rules.

Coincidentally, this PR itself is now ready for review, should we decide to proceed with the Sabberworm parser...

src/CssInliner.php
```diff
@@ -1073,10 +1073,10 @@ private function removeUnsupportedOfTypePseudoClasses(string $selectorPart): str
...
      *
      * @param CssDocument $parsedCss
-     *        This may contain any `@import` or `@font-face` rules that should precede the CSS placed in the `<style>`
-     *        element.  If there are no uninlinable CSS rules to copy there, a `<style>` element will be created
-     *        containing only the applicable at-rules from `$parsedCss`.  If there are none, and there are also no
-     *        unlinlinable CSS rules, an empty `<style>` element will not be created.
+     *        This may contain any `@import`, `@font-face`, or other at-rules that should precede the CSS placed in the
+     *        `<style>` element.  If there are no uninlinable CSS rules to copy there, a `<style>` element will be
+     *        created containing only the applicable at-rules from `$parsedCss`.  If there are none, and there are also
+     *        no unlinlinable CSS rules, an empty `<style>` element will not be created.
```
JakeQZ, line 1079: I think this comment can be updated as a separate PR.  We already have tests involving `@keyframes`, `@supports`, `@-webkit-keyframes` and `@-moz-document`, so I think it was already outdated.

tests/Unit/CssInlinerTest.php
```diff
@@ -3440,10 +3440,7 @@ public function provideInvalidImportRules(): array
                 '@media (max-width: 640px) { p { color: red; } }' . "\n"
                 . '@import "foo.css";',
             ],
-            '@import after incorrectly-cased @charset rule' => [
-                '@CHARSET "UTF-8";' . "\n"
-                . '@import "foo.css";',
-            ],
+            // broken: @import after incorrectly-cased @charset rule
```
JakeQZ, line 3443: According to CSS spec, `@charset` should be rejected unless it is entirely lowercase (and followed by exaclty one space, not generic whitespace, then opening double quote).

src/CssInliner.php
JakeQZ: See #1077.

JakeQZ:
> Coincidentally, this PR itself is now ready for review, should we decide to proceed with the Sabberworm parser...

Marked as draft again, pending #1077.

JakeQZ:
> Marked as draft again, pending #1077.

Now rebased with #1077.  This is ready for review, in conjunction with the possible alternative in #1076.

tests/Unit/CssInlinerTest.php
JakeQZ: Hmm.  From the specification:

> User agents must ignore any `@import` rule that occurs inside a block or after any non-ignored statement other than an `@charset` or an `@import` rule.

That probably means `@import` after an invalid uppercase `@CHARSET` rule should not be ignored, because the invalid `@CHARSET` rule was ignored and is thus not a "non-ignored statement" - and therefore this test is actually invalid - ?

src/Utilities/CssDocument.php
```diff
@@ -82,23 +61,47 @@ public function __construct(string $css)
...
         $ruleMatches = [];
-        foreach ($splitCss as $cssPart) {
-            // process each part for selectors and definitions
-            \preg_match_all('/(?:^|[\\s^{}]*)([^{]+){([^}]*)}/mi', $cssPart['css'], $matches, PREG_SET_ORDER);
-
-            foreach ($matches as $cssRule) {
+        /** @var CssRenderable $rule */
+        foreach ($this->sabberwormCssDocument->getContents() as $rule) {
+            if ($rule instanceof CssAtRuleBlockList && $rule->atRuleName() === 'media') {
+                /** @var string $mediaQueryList */
```
oliverklee, line 68: I propose we move this code block into a private method.

src/Utilities/CssDocument.php
```diff
@@ -82,23 +61,47 @@ public function __construct(string $css)
...
+                $mediaQueryList = $rule->atRuleArgs();
+                [$mediaType] = \explode('(', $mediaQueryList, 2);
+                if (\trim($mediaType) !== '') {
+                    $mediaTypesExpression = \implode('|', $allowedMediaTypes);
+                    if (!\preg_match('/^\\s*+(?:only\\s++)?+(?:' . $mediaTypesExpression . ')/i', $mediaType)) {
```
oliverklee, line 73: We need to have a `preg_quote` somewhere in here.

src/Utilities/CssDocument.php
```diff
@@ -82,23 +61,47 @@ public function __construct(string $css)
...
+                        continue;
+                    }
+                }
+                $media = '@media ' . $mediaQueryList;
```
oliverklee, line 77: I'm not a big fan of having the types in variable names. Maybe we can rename the variable to `$mediaQueries`?

src/Utilities/CssDocument.php
```diff
@@ -82,23 +61,47 @@ public function __construct(string $css)
...
-                    'declarations' => $cssRule[2],
+                    'media' => '',
+                    'rule' => $rule,
                 ];
             }
         }
 
-        return $ruleMatches;
+        return \array_map(
+            function (array $ruleMatch): array {
```
oliverklee, line 96: Can we make the method static?

And can we provide it with more specific type annotations?

src/Utilities/CssDocument.php
```diff
@@ -110,175 +113,62 @@ public function getStyleRulesData(array $allowedMediaTypes): array
...
-            }
+        switch ($rule->atRuleName()) {
+            case 'media':
+                return false;
+            case 'font-face':
+                return $rule instanceof CssRuleSet
+                    && $rule->getRules('font-family') !== []
+                    && $rule->getRules('src') !== [];
+            default:
+                return true;
```
oliverklee, line 171: Please let's have early returns only for guard clauses.

oliverklee (changes requested):
I've added my thoughts on reading through the first version.

src/Utilities/CssDocument.php
```diff
@@ -82,23 +61,47 @@ public function __construct(string $css)
...
-
         $ruleMatches = [];
-        foreach ($splitCss as $cssPart) {
-            // process each part for selectors and definitions
-            \preg_match_all('/(?:^|[\\s^{}]*)([^{]+){([^}]*)}/mi', $cssPart['css'], $matches, PREG_SET_ORDER);
-
-            foreach ($matches as $cssRule) {
+        /** @var CssRenderable $rule */
+        foreach ($this->sabberwormCssDocument->getContents() as $rule) {
+            if ($rule instanceof CssAtRuleBlockList && $rule->atRuleName() === 'media') {
```
oliverklee, line 67: Would a decicated method in sabberworm/php-css-parser make sense for this (i.e., should we open a ticket)?

src/Utilities/CssDocument.php
```diff
@@ -82,23 +61,47 @@ public function __construct(string $css)
...
+                [$mediaType] = \explode('(', $mediaQueryList, 2);
+                if (\trim($mediaType) !== '') {
+                    $mediaTypesExpression = \implode('|', $allowedMediaTypes);
+                    if (!\preg_match('/^\\s*+(?:only\\s++)?+(?:' . $mediaTypesExpression . ')/i', $mediaType)) {
+                        continue;
+                    }
+                }
+                $media = '@media ' . $mediaQueryList;
+                /** @var CssRenderable $nestedRule */
+                foreach ($rule->getContents() as $nestedRule) {
```
oliverklee, line 79: Would a decicated method in sabberworm/php-css-parser make sense for this (i.e., should we open a ticket)?

src/Utilities/CssDocument.php
```diff
@@ -110,175 +113,62 @@ public function getStyleRulesData(array $allowedMediaTypes): array
...
+     * - other at-rules are assumed to be valid and treated as a black box - `true` is returned.
      *
-     * @param string $atIdentifier name of the at-rule with the preceding at sign
-     * @param string $rule full content of the rule, including the at-identifier
+     * @param CssRenderable $rule
      *
      * @return bool
      */
-    private function isValidAtRule(string $atIdentifier, string $rule): bool
+    private function isValidAtRuleToRender(CssRenderable $rule): bool
```
oliverklee, line 147: Would a decicated method in sabberworm/php-css-parser make sense for this (i.e., should we open a ticket)?

src/Utilities/CssDocument.php
```diff
@@ -110,175 +113,62 @@ public function getStyleRulesData(array $allowedMediaTypes): array
...
-                $cleanedCss = \trim(\preg_replace($cssCleaningMatchers, '', $cssPart));
-                if ($cleanedCss !== '') {
-                    $splitCss[] = ['css' => $cleanedCss, 'media' => ''];
-                }
-            }
+        switch ($rule->atRuleName()) {
+            case 'media':
+                return false;
+            case 'font-face':
+                return $rule instanceof CssRuleSet
```
oliverklee, line 167: Would a decicated method in sabberworm/php-css-parser make sense for this (i.e., should we open a ticket)?

oliverklee (changes requested):

src/Utilities/CssDocument.php
JakeQZ: You mean for `$rule instanceof CssAtRuleBlockList`?  I think it's more than arguable that it doesn't justify a specific testing method, and that the code with the `instanceof` test is actually as succinct as it can be.

Also, for static analysis (e.g. Psalm) the `instanceof` test would probably still be needed even after a separate method returned `true`.

src/Utilities/CssDocument.php
JakeQZ: > 
> 
> Would a decicated method in sabberworm/php-css-parser make sense for this (i.e., should we open a ticket)?

Looks like the wrong copypasta comment here.

I think it would be worth trying to get Psalm integrated into `sabberworm/php-css-parser`, and thus opening a ticket.

Although a year ago I was tearing my hair out with it (probably because of the caching issue which needs fixing in Psalm), it has very much grown on me since then.  It also paves the way forward for a defined standard for more complex DocBlock (and perhaps PHP itself) type annotations for which there have been rumbling discussions in various places for many years.

src/Utilities/CssDocument.php
JakeQZ: No, this filtering is quite specific to our use case.

Here, we discard `@media` rules because they have been processed separately (so that their ordering with respect to other uninlinable CSS is correct, and they can be discarded if they don't match anything).  We could in theory process other conditional at-rules (e.g. #1078) in the same way.

We deliberately discard `@charset` rules, which should otherwise be passed on, because we only support UTF-8.

We have sort-of decided (see #870) to discard `@font-face` rules without at least a `src` and `font-family`, which is the minimum to be meaningful.

We are following the user-agent spec and discarding invalidly-placed `@import` rules (I think it's a "user agents MUST ignore `@import` rules which are preceded by a non-ignored statement other than an `@charset` rule or another `@import` rule" [from memory!]).  This I think is the only aspect where a different, alternative or enhanced behaviour of the Sabberworm parser might be useful more widely.

src/Utilities/CssDocument.php
JakeQZ: Ditto for other `instanceof`.  I don't think so.  I think it's perfectly succinct as it is.

src/Utilities/CssDocument.php
JakeQZ: I think there can be a case for breaking this rule, if all of the following apply:
- The `swtich` statement is the only statement in the method;
- All `case` branches have a `return` statment and not a `break` statement, except where a 'follow-through' is allowed or for the last `case` or `default` if `null` or `void` is the obviosuly desired return value in that case.

However, maintaining that standard through future code changes may be harder than it seems.  So I will probably change it when I revisit the code changes on this - perhaps tomorrow.

JakeQZ: > if all of the following apply

Such a method would be atomic.  If there was any future need to add pre-amble or post-amble code, then it should be moved to a new atomic method, with the original method becoming a wrapper with the additional pre- or post- code.  Though we know from experience it often doesn't turn out that way.

src/Utilities/CssDocument.php
JakeQZ: > I think it would be worth trying to get Psalm integrated into `sabberworm/php-css-parser`, and thus opening a ticket.

At a minimum, it would be good to have proper type annotations on the public methods, both the basic PHP ones and the more detailed DocBlock ones that Psalm (and perhaps other tools) recognize.

It should also be easily possible to add Psalm to that project, and generate a current 'baseline', without making any actual code changes.

tests/Unit/CssInlinerTest.php
oliverklee: Probably. Then let's disable/change the corresponding test (preferably in a separate PR).

src/Utilities/CssDocument.php
oliverklee: I've actually started to add static type checking with PHPStan to `sabberworm/php-css-parser` and now am fixing the issues one by one: https://github.com/sabberworm/PHP-CSS-Parser/pull/227

src/Utilities/CssDocument.php
JakeQZ: Good spot again.  This was added after review in #1014, but the commit here was based on an older code base so did not include it.

src/Utilities/CssDocument.php
JakeQZ: Will do.  Just looked to see if it would be possible to create a separate method for this as a separate PR (as the functionality is pre-existing), but it seems not - the pre-existing code is tangled up with a `preg_split`, and I don't see much point untangling it if it's going to be replaced anyway.

JakeQZ: Done.

src/Utilities/CssDocument.php
JakeQZ: Now done.  I already replied to this - where has my reply gone?

src/Utilities/CssDocument.php
JakeQZ: According to the [spec, or at least MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@media#media-query-list), it is quite specifically called a "media query list".  Changing it to something else I think could sow more confusion.

src/Utilities/CssDocument.php
JakeQZ: Done.

src/Utilities/CssDocument.php
JakeQZ: > 
> 
> Please let's have early returns only for guard clauses.

I've changed the `switch` statement, though not the preceding code.

I don't think I agree with a "one size fits all" approach regarding where a return statement is or isn't acceptable, provided the complexity of an individual method can be kept low, as I've indicated above.

JakeQZ:
I've hopefully now resolved all the code review issues, and rebased to include #1081.  Unfortunately the latter has uncovered a bug in the Sabberworm CSS parser which is beyond the scope of this PR, so it will still not be as clean a commit as desired, because some newly-added tests must now be temporarily disabled.

src/Utilities/CssDocument.php
```diff
@@ -82,23 +61,45 @@ public function __construct(string $css)
...
-        foreach ($splitCss as $cssPart) {
-            // process each part for selectors and definitions
-            \preg_match_all('/(?:^|[\\s^{}]*)([^{]+){([^}]*)}/mi', $cssPart['css'], $matches, PREG_SET_ORDER);
-
-            foreach ($matches as $cssRule) {
+        /** @var CssRenderable $rule */
+        foreach ($this->sabberwormCssDocument->getContents() as $rule) {
+            if ($rule instanceof CssAtRuleBlockList) {
+                $media = $this->getFilteredAtIdentifierAndRule($rule, $allowedMediaTypes);
+                if ($media !== null) {
```
oliverklee, line 69: I'd prefer `is_string` here instead so we can communicate what it should be, not what it shouldn't be.

src/Utilities/CssDocument.php
```diff
@@ -82,23 +61,45 @@ public function __construct(string $css)
...
+                    /** @var CssRenderable $nestedRule */
+                    foreach ($rule->getContents() as $nestedRule) {
+                        if ($nestedRule instanceof CssDeclarationBlock) {
+                            $ruleMatches[] = [
```
oliverklee, line 73: Let's fold this assignment into one line so the method takes less vertical screen space.

src/Utilities/CssDocument.php
```diff
@@ -82,23 +61,45 @@ public function __construct(string $css)
...
+                        }
+                    }
+                }
+            } elseif ($rule instanceof CssDeclarationBlock) {
                 $ruleMatches[] = [
-                    'media' => $cssPart['media'],
-                    'selectors' => $cssRule[1],
-                    'declarations' => $cssRule[2],
+                    'media' => '',
+                    'rule' => $rule,
```
oliverklee, line 83: Ditto.

src/Utilities/CssDocument.php
```diff
@@ -82,23 +61,45 @@ public function __construct(string $css)
...
         }
 
-        return $ruleMatches;
+        return \array_map(
+            /**
+             * @param array{media: string, rule: CssDeclarationBlock} $ruleMatch
+             *
+             * @return array{media: string, selectors: string, declarations: string}
+             */
+            static function (array $ruleMatch): array {
```
oliverklee, line 94: Let's use a non-anonymous function instead: Then we can use its name to communicate its purpose.

src/Utilities/CssDocument.php
```diff
@@ -110,177 +111,104 @@ public function getStyleRulesData(array $allowedMediaTypes): array
...
+                $result = false;
+                break;
+            case 'font-face':
+                $result = $rule instanceof CssRuleSet
+                    && $rule->getRules('font-family') !== []
+                    && $rule->getRules('src') !== [];
+                break;
+            default:
+                $result = true;
+                break;
```
oliverklee, line 209: The `default` case generally should not have a `break`.

oliverklee (changes requested):
Looks very good! Now we only need the final polish.

src/Utilities/CssDocument.php
JakeQZ: Fair point, well made.  Changed.

src/Utilities/CssDocument.php
JakeQZ: OK.  Done.

src/Utilities/CssDocument.php
JakeQZ: Ditto.

src/Utilities/CssDocument.php
JakeQZ: Done.  It's got to be a contender for the crappiest method I've ever written, and I hope will be gone when we replace these arrays with proper classes, but I'm all for atomizing code into the smallest constituent parts.

src/Utilities/CssDocument.php
JakeQZ: Only if it's the last case.  Some weirdos like to put the `default` first.  But we're not weird, so I've removed the redundant `break`.

JakeQZ:
> 
> 
> Looks very good! Now we only need the final polish.

Have made suggested changes.

Do I get the prize of not having to make a rebase involving a file rename? ;))

JakeQZ:
> 
> 
> ... rebased to include #1081. Unfortunately the latter has uncovered a bug in the Sabberworm CSS parser which is beyond the scope of this PR, so it will still not be as clean a commit as desired, because some newly-added tests must now be temporarily disabled.

https://github.com/sabberworm/PHP-CSS-Parser/issues/261

oliverklee (approved):
Beautiful!

PR #1018, opened 2021-04-14
[CLEANUP] Remove a redundant CSS data cache

JakeQZ:
The cache with the key `CACHE_KEY_CSS` was storing all of the parsed CSS data
as a single entry, but would only ever be set once and never accessed.  It does
not make sense to call `inlineCss` again on the same document with the same CSS,
but in any case this method clears all caches as its first action.

This will reduce the memory footprint after `inlineCss` returns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 - Drop support for Symfony 5.1 (#972)
 - Drop support for PHP 7.1 (#967)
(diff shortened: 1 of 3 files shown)


JakeQZ:
It's possible that this cache was once useful, but as the code has changed over the years it has become redundant.

It could be made useful by making it `static` and not having `inlineCss` clear it.  However, that would result in a permanent memory footprint after Emogrifier has been used, unless explicitly cleared by the client, which would probably put unnecessary onus on the client.

oliverklee (approved):

PR #1019, opened 2021-04-14
[FEATURE] Composer scripts for Psalm

JakeQZ:
These save having to remember the command line arguments and/or type a more
intricate command:
- `composer psalm:baseline`: Update the `psalm.baseline.xml` to remove any issues
  that have been fixed;
- `composer psalm:cc`: Clear Psalm's cache for the project; sometimes this may
  be required to resolve spurious errors that might arise after switching
  between branches in the same directory.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -95,7 +95,7 @@
         "composer:normalize": "@php \"./.phive/composer-normalize\"",
         "php:fix": "@php \"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix config/ src/ tests/",
         "php:version": "@php -v | grep -Po 'PHP\\s++\\K(?:\\d++\\.)*+\\d++(?:-\\w++)?+'",
-        "psalm:baseline": "@php \"./.phive/psalm\" --update-baseline",
+        "psalm:baseline": "@php \"./.phive/psalm\" --set-baseline=psalm.baseline.xml",
         "psalm:cc": "@php \"./.phive/psalm\" --clear-cache"
     },
     "support": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -94,7 +94,9 @@
         "ci:tests:unit": "@php \"./vendor/bin/phpunit\"",
         "composer:normalize": "@php \"./.phive/composer-normalize\"",
         "php:fix": "@php \"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix config/ src/ tests/",
-        "php:version": "@php -v | grep -Po 'PHP\\s++\\K(?:\\d++\\.)*+\\d++(?:-\\w++)?+'"
+        "php:version": "@php -v | grep -Po 'PHP\\s++\\K(?:\\d++\\.)*+\\d++(?:-\\w++)?+'",
+        "psalm:baseline": "@php \"./.phive/psalm\" --update-baseline",
```
oliverklee, line 98: We'll need to use `--set-baseline=psalm.baseline.xml` here instead to regenerate the baseline AFAIK.

oliverklee (changes requested):

composer.json
JakeQZ: I think we're looking at different functionality - `--set-baseline` adds new issues to the baseline, whereas `--update-baseline` removes issues that have been fixed (but does not add any new issues).  I don't know if they can be combined - or maybe `--set-baseline` also removes fixed issues (if it regenerates it from scratch, it will do by definition), so that's the one we want anyway?

oliverklee: Then let's first find out whether `---set-baseline` also removes fixed issues, and then come to an informed decision.

JakeQZ: Yep, `--set-baseline` covers both bases, so have changed to that.

oliverklee (approved):

PR #1020, opened 2021-04-14
[CLEANUP] Drop Psalm warnings for PHPUnit 9

oliverklee:
We have pinned PHPUnit to version 8 and hence do not get warnings
for using functionality that has been deprecated in PHPUnit 9 yet.


JakeQZ (approved):
Haha, when running the script in #1019, even though I had the latest `composer.json`, I had omitted to run `composer update`.  Sorry.

PR #1021, opened 2021-04-15
[CLEANUP] Improve the type annotations in `CssInliner` some more

oliverklee:


src/CssInliner.php
```diff
@@ -78,7 +78,11 @@ class CssInliner extends AbstractHtmlProcessor
     private $allowedMediaTypes = ['all' => true, 'screen' => true, 'print' => true];
 
     /**
-     * @var array<int, array<string, mixed>>
+     * @var array{
+     *         0: array<string, int>,
+     *         1: array<string, array<string, string>>,
+     *         2: array<string, string>
```
JakeQZ, line 84: I assume it would be asking too much to use the constant names in the type annotation.

JakeQZ (approved):

PR #1022, opened 2021-04-15
[CLEANUP] Add newline in DocBlock before long descriptions

JakeQZ:
This will allow for changing the parameter or return value type (or parameter
name) with a clean commit, without having to reformat the description for 120
character line-lengths.

Also
- Align multiline `@param` and `@return` bodies with the first line (i.e. so
  that each subsequent line of description (or in some cases, type) is at the
  position after the space following the at-tag;
- Remove 'sentence' capitalization and full stop from some descriptions which
  are not sentences;
- Move part of a parameter description for `parseCssShorthandValue` to the
  method description where it belongs;
- Use backticks around parts of descriptions which are clearly code.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -518,10 +518,11 @@ class CssInliner extends AbstractHtmlProcessor
      *
      * @param string $css CSS with comments removed
      *
-     * @return string[] The first element is the CSS with the valid `@import` and `@charset` rules removed.  The second
-     *         element contains a concatenation of the valid `@import` rules, each followed by whatever whitespace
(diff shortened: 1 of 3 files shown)


JakeQZ:
A while ago I looked at how other projects formatted their `@param` and `@return` in the DocBlock.  I found consistency with `@param` (which is carried forward here) but a whole variety of indentations after `@return`.  Now Psalm introduces some potentially very long type specifications, the ultimate conclusion for the formatting seems kind-of inevitable.

oliverklee (approved):

PR #1024, opened 2021-04-16
[CLEANUP] Fix more Psalm warnings in `CssInliner`

oliverklee:


src/CssInliner.php
```diff
@@ -924,11 +925,16 @@ private function splitCssAndMediaQuery(string $css): array
...
      *
      * @param \DOMElement $node
-     * @param string[][] $cssRule
+     * @param array{
+     *            media: string,
+     *            selector: string,
+     *            hasUnmatchablePseudo: bool,
+     *            declarationsBlock: string,
+     *            line: int
+     *        } $cssRule
```
JakeQZ, line 934: I see the original type hint indicated an array of arrays, when in fact it should have been a simple associative array.

JakeQZ (approved):

PR #1035, opened 2021-04-19
[TASK] Support more file types in the `.editorconfig`

oliverklee:
Now we also have formatting options for YAML and JSON files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -22,3 +22,13 @@ indent_size = 4
 [*.xml]
 indent_style = space
 indent_size = 4
+
+# YAML files
+[*.{yml,yaml}]
+indent_style = space
+indent_size = 2
+
+# JSON files
+[*.json]
+indent_style = space
+indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -31,4 +31,4 @@ indent_size = 2
 # JSON files
 [*.json]
 indent_style = space
-indent_size = 2
+indent_size = 4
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -22,3 +22,13 @@ indent_size = 4
 [*.xml]
 indent_style = space
 indent_size = 4
+
+# YAML files
+[*.{yml,yaml}]
```
JakeQZ, line 27: I can confirm this works with Notepad++ with EditorConfig plugin.

.editorconfig
```diff
@@ -22,3 +22,13 @@ indent_size = 4
...
+indent_style = space
+indent_size = 2
+
+# JSON files
+[*.json]
+indent_style = space
+indent_size = 2
```
JakeQZ, line 34: Our `composer.json` is indented with 4 spaces.

JakeQZ (changes requested):
Shouldn't it be 4 spaces indent for `.json`?

JakeQZ:
Also noticed the 4 spaces indent already here for `.md`.  Aren't we indenting `.md` with 2 spaces (e.g. bulleted lists in README)?  Or have we not really decided?

.editorconfig
oliverklee: Changed an repushed.

oliverklee:
I think we have not really decided. :-)

- https://www.markdownguide.org/basic-syntax/ does not say anything about indentation styles.
- https://markdown-guide.readthedocs.io/en/latest/basics.html says "4 spaces"
- https://google.github.io/styleguide/docguide/style.html also says 4 spaces

We might need to reformat our markdown files after this.

JakeQZ (approved):

PR #1037, opened 2021-04-19
[BUGFIX] Add missing semicolon in CSS test data

JakeQZ:
The `@font-face` rule used for testing CSS parsing was missing a semicolon
between its properties.

Emogrifier currently passes on most of these types of at-rules unmodified,
without checking their validity too much.  Other CSS parsers might not be so
generous.


oliverklee (approved):
LGTM.

PR #1039, opened 2021-04-21
[CLEANUP] Autoformat YAML, Markdown and XML files

oliverklee:


.github/workflows/ci.yml
```diff
@@ -91,7 +91,7 @@ jobs:
 
     runs-on: ubuntu-20.04
 
-    needs: [php-lint, composer-validate]
+    needs: [ php-lint, composer-validate ]
```
JakeQZ, line 94: Why spaces immediately inside the braces?  We don't have this formatting in PHP and JSON, and wouldn't have it in JavaScript (if we had any).  It seems inconsistent to me, and I don't see a good reason for it.

.phive/phars.xml
```diff
@@ -1,9 +1,10 @@
...
-  <phar name="phpcs" version="^3.5.8" installed="3.5.8" location="./.phive/phpcs" copy="true"/>
-  <phar name="phpmd" version="^2.9.1" installed="2.9.1" location="./.phive/phpmd" copy="true"/>
-  <phar name="psalm" version="^4.7.0" installed="4.7.0" location="./.phive/psalm" copy="true"/>
+    <phar name="composer-normalize" version="^2.13.3" installed="2.13.3" location="./.phive/composer-normalize"
+          copy="true"/>
+    <phar name="phar-io/phive" version="^0.14.4" installed="0.14.4" location="./.phive/phive" copy="true"/>
+    <phar name="php-cs-fixer" version="^2.16.7" installed="2.16.7" location="./.phive/php-cs-fixer" copy="true"/>
+    <phar name="phpcs" version="^3.5.8" installed="3.5.8" location="./.phive/phpcs" copy="true"/>
+    <phar name="phpmd" version="^2.9.1" installed="2.9.1" location="./.phive/phpmd" copy="true"/>
+    <phar name="psalm" version="^4.7.0" installed="4.7.0" location="./.phive/psalm" copy="true"/>
```
JakeQZ, line 9: While we have 4 spaces indent for XML files in general, what happens when this file is regenerated by PHIVE?  Will it revert back to 2 spaces indent and need reformatting?

.phive/phars.xml
```diff
@@ -1,9 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phive xmlns="https://phar.io/phive">
-  <phar name="composer-normalize" version="^2.13.3" installed="2.13.3" location="./.phive/composer-normalize" copy="true"/>
-  <phar name="phar-io/phive" version="^0.14.4" installed="0.14.4" location="./.phive/phive" copy="true"/>
-  <phar name="php-cs-fixer" version="^2.16.7" installed="2.16.7" location="./.phive/php-cs-fixer" copy="true"/>
-  <phar name="phpcs" version="^3.5.8" installed="3.5.8" location="./.phive/phpcs" copy="true"/>
-  <phar name="phpmd" version="^2.9.1" installed="2.9.1" location="./.phive/phpmd" copy="true"/>
-  <phar name="psalm" version="^4.7.0" installed="4.7.0" location="./.phive/psalm" copy="true"/>
+    <phar name="composer-normalize" version="^2.13.3" installed="2.13.3" location="./.phive/composer-normalize"
+          copy="true"/>
```
JakeQZ, line 4: Do we need to restrict XML files to 80 character line lengths?  If we have 120 character line lengths in general, I don't see why XML should be an exception.

Also, if an element with many/long attributes needs to be split on to multiple lines, I'd much rather this format:

```xml
    <phar
        name="composer-normalize"
        version="^2.13.3"
        installed="2.13.3"
        location="./.phive/composer-normalize"
        copy="true"
    />
```

I think it is much more logical and readable for those situations.  Can autoformatters achieve this?

JakeQZ (changes requested):
More questions than answers I'm afraid.  Actually I find issues with pretty much all of the reformatting :/ ...

.github/workflows/ci.yml
oliverklee: That's the default setting of the PhpStorm YAML autoformatter. (I'd like contributors to be able to use the PhpStorm default settings as much as possible when they work on Emogrifier so it's as small as a hassle for them as possible.)

JakeQZ: OK.

JakeQZ (approved):

PR #1041, opened 2021-04-22
[TASK] getCssRegularExpression without delimiters

JakeQZ:
This will allow the method, correspondingly renamed from
`getCssNeedleRegularExpressionPattern` to `getCssRegularExpressionMatcher`, to
be used more flexibly, not necessarily for finding needles in haystacks.  It
also allows slight simplifications and tightening-up of its tests.

Update the DocBlock for the method to reflect this change and previous changes.

Also now compute the regular expression version of the CSS in the `Constraint`
constructors - this will be more efficient if the `Constraint` is re-used.

### DIFF
--- a/tests/Support/Constraint/CssConstraint.php
+++ b/tests/Support/Constraint/CssConstraint.php
@@ -49,35 +49,35 @@ abstract class CssConstraint extends Constraint
     /x';
 
     /**
-     * Processing of @media rules may involve removal of some unnecessary whitespace from the CSS placed in the <style>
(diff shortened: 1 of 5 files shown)


JakeQZ:
I'm going with the following nomenclature in the context of regex:
- a 'matcher' is a part of a regular expression, and may correspond with the `$matches` callback argument (e.g. to `preg_replace_callback`);
- a 'pattern' is fully-fledged with delimiters and modifiers, and usable as the `$pattern` argument (e.g. to `preg_match`).

oliverklee (approved):

PR #1042, opened 2021-04-22
[CLEANUP] Move CssConstraint DocBlock info re c'tor there

JakeQZ:
Remove the information about the constructor for `CssConstraint` from the
DocBlock for the class, and add it to that for the constructor where it belongs.

Also further explain why the constructor is retained (see #1002).

Also change constructor visibility to `protected` (for good measure) - the class
is not intended to be instantiated in itself (and is `abstract` anyway).


JakeQZ:
I only ended up making this change because it seemed odd that the constructor was the last method in the file, so was about to move it ahead of the others.  Then I realized the others were all `static`.  (I generally try to order classes: constants, properties, static methods, non-static methods.  Within each group, greater visibility first.  Then constructor or `init` method first, and drill down to lower level methods after higher level ones.)

JakeQZ:
Commit title looks a bit confusing.  Trying to keep it within 50 chars is difficult.

oliverklee (approved):

PR #1044, opened 2021-04-24
[TASK] Add an IsEquivalentCss Constraint

JakeQZ:
And associated assertion methods.

### DIFF
--- a/tests/Unit/Support/Constraint/IsEquivalentCssTest.php
+++ b/tests/Unit/Support/Constraint/IsEquivalentCssTest.php
@@ -1,0 +1,84 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\Tests\Unit\Support\Constraint;
+
+use Pelago\Emogrifier\Tests\Support\Constraint\IsEquivalentCss;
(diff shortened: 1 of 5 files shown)


tests/Unit/Support/Constraint/IsEquivalentCssTest.php
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * @return array<string, array{0: string, 1: string}>
+     */
+    public function provideNonEquivalentCss(): array
+    {
+        $datasets = $this->provideCssNeedleNotFoundInHaystack() + $this->provideCssNeedleFoundInLargerHaystack();
+
+        $transposedDatasets = \array_map(
+            static function (array $dataset): array {
```
oliverklee, line 63: Please let's have PHPDoc type annotations for this static function.

oliverklee (changes requested):

tests/Unit/Support/Constraint/IsEquivalentCssTest.php
JakeQZ: Done.  I've not added type annotations for the other closure because it's parameter and return value are both strings, so a DocBlock would not provide any additional information.

oliverklee (approved):

PR #1047, opened 2021-04-24
[TASK] Allow syntactic variation of @import

JakeQZ:
When testing for equivalent or needle-in-haystack CSS, allow the URL parameter
of the `@import` rule to be enclosed in quotes (single or double) or not, and/or
wrapped by the CSS `url` function or not.

Other CSS parsers may render this rule equivalently but using a slightly
different syntax.

### DIFF
--- a/tests/Support/Constraint/CssConstraint.php
+++ b/tests/Support/Constraint/CssConstraint.php
@@ -45,15 +45,50 @@ abstract class CssConstraint extends Constraint
     /x';
 
     /**
+     * This is for matching a URL in a string that has already been converted from CSS into a regular expression pattern
+     * to match it, thus needs to match special characters in their escaped form, and whitespace as `\s++`.  It matches
+     * one or more characters which are not quotes, whitespace, a semicolon, or a closing parenthesis, optionally
+     * enclosed in single or double quotes, and not beginning with `url(`.  The URL without the enclosing quotes is
+     * captured in its second group (the first group is needed to match the optional opening quote so that the closing
+     * quote can be matched).
+     *
+     * @var string
+     */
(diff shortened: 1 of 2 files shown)


JakeQZ:
Catering for CSS variations beyond whitespace is starting to make testing a bit more complicated, but we are nearly there, at least as far as allowing for the Sabberworm CSS parser is concerned.

I have wondered if a wholly different approach might be better (e.g. adding a CSS parser to the testing environment), but concluded such alternatives would require even more work and refactoring.  However, that's not to say it's off the table moving forwards.  At the moment, I'd like reach a point where we can switch to using the Sabberworm CSS parser with a clean commit (changing just one class, `CssDocument`), with minimal effort.

JakeQZ:
The force-push added a couple of extra tests I thought of, and changed dataset descriptions to use "\`@media\`" instead of "media".  It's still ready for review.

JakeQZ:
An alternative approach that might be easier to follow would be to provide a list of possible valid expected renderings for the tests.

For example, with `import "foo.css";`, the valid renderings might be `['import "foo.css";', 'import url("foo.css");']` (and could be extended to cater for more possibilities as required).

JakeQZ:
I've added an alternative PR as #1050.  Either one of these PRs is intended to be merged, but not both, depending on which is preferred on grounds of simplicity, maintainability, etc.

(The apprach here adjusts the regular expression matcher to cater for the variations.  The alternative in #1050 explicitly lists possible renderings of the `@import` rule instead.)

oliverklee (approved):
Then let's use this one as it allows the tests themselves to be more simple.

PR #1048, opened 2021-04-25
[CLEANUP] Assert invalid at-rule not found

JakeQZ:
Instead of asserting that the resultant CSS matches the valid CSS, assert that
it does not contain the invalid at-rule, by simply checking that `@` followed by
the rule name (of the invalid rule) is not found.  This avoids having to cater
for variations in the rendering of the valid CSS.

### DIFF
--- a/tests/Unit/Utilities/CssDocumentTest.php
+++ b/tests/Unit/Utilities/CssDocumentTest.php
@@ -434,13 +434,9 @@ final class CssDocumentTest extends TestCase
 
         $result = $subject->renderNonConditionalAtRules();
 
-        self::assertThat(
-            \trim($result),
-            self::logicalOr(
-                self::identicalTo(''),
-                self::identicalTo(\trim($cssBefore))
-            )
-        );
+        \preg_match('/@[\\w\\-]++/', $atRuleCss, $atAndRuleNameMatches);
+        $atAndRuleName = $atAndRuleNameMatches[0];
+        self::assertStringNotContainsString($atAndRuleName, $result);
     }
 
     /**


tests/Unit/Utilities/CssDocumentTest.php
```diff
@@ -434,13 +434,9 @@ public function notRendersInvalidNonConditionalAtRule(string $atRuleCss, string
...
         $result = $subject->renderNonConditionalAtRules();
 
-        self::assertThat(
-            \trim($result),
-            self::logicalOr(
-                self::identicalTo(''),
-                self::identicalTo(\trim($cssBefore))
-            )
-        );
+        \preg_match('/@[\\w\\-]++/', $atRuleCss, $atAndRuleNameMatches);
```
oliverklee, line 437: What does the double `+` mean?

JakeQZ: The first `+` is the usual 'one or more instances' repetition operator.  The second `+` is the [possessive quantifier](https://www.regular-expressions.info/possessive.html) telling the regex engine not to backtrack when attempting to match whatever is following - it imporives efficiency and reduces the memory footprint because it does not then have to store a tree of backtracking information.  It can be used wherever the following matcher couldn't possibly match anything the matcher it is applied to could (or if there is no following matcher, as in this case).  For example, `[\w\-]++\-` could never match anything, because the first matcher will also match any hyphen, and with no backtracking allowed the engine can't then step back a character to see if the final hyphen can be matched.

oliverklee (approved):

PR #1050, opened 2021-04-28
[TASK] Allow syntactic variation of @import

JakeQZ:
When testing that valid at-rules are rendered, allow the URL parameter of the
`@import` rule to be possibly wrapped by the CSS `url` function.

Other CSS parsers may render this rule equivalently but using a slightly
different syntax.


JakeQZ:
Please note this is an alternative solution achieving the same as #1047.  Either one of these PRs is intended to be merged but not both.  This appoach explicitly lists alterative possible renderings of the `@import` rule, instead of adjusting the regular expression matcher to cater for the variations.

oliverklee:
Closing, as I have just merged the other variant (#1047).

Issue #1051, opened 2021-04-28
HTML5 converted to HTML4


bytestream:
Given valid HTML5 (according to https://validator.nu/) `CSSInliner` converts `<a><table><tr><td>hello world</td></tr></table></a>` to `<a></a><table><tr><td>hello world</td></tr></table>`

```php
<?php

$html = '<!DOCTYPE html><html lang="en"><head><title>hello world</title></head><body><a><table><tr><td>hello world</td></tr></table></a></body></html>';

echo CssInliner::fromHtml($html)->inlineCss()->render();
/*
<!DOCTYPE html>
<html lang="en">
<head>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
<title>hello world</title>
</head>
<body>
<a></a><table><tr><td>hello world</td></tr></table>
</body>
</html>
*/
```

Relates to #828

JakeQZ:
Hi Kieran,

Good find.

I created a branch with some tests for the transparent content model (including `<ins>` and `<del>` as well as `<a>` for the transparent element, and a few other examples of flow content for it to contain).  See https://github.com/MyIntervals/emogrifier/commit/86e49736aad4e65a8dcb8f3fa32e42940c63cad8.

It seems only the specific case of `<table>` being a child of `<a>` fails.  If the `<table>` is wrapped in a `<div>` (which I would suggest as a workaround), it parses OK.

Note that it also seems to fail with PHP 8 (though only ran with lowest dependencies before aborting due to other failures), which was supposed to have some improvements to DOM handling.  Behaviour is the same on my Windows system as the (presumably Linux) build server.

I think this bug should be reported to those concerned.  If it's with PHP or an extension (i.e. DOM), there's [bugs.php.net](https://bugs.php.net/); if it's with `libxml`, I don't know what the reportage lines are but am sure some exist.  (Sounds like it shouldn't be with `libxml` by name but I don't know how much HTML handling is intertwined with it.)

In the meantime, this does add some gravitas to getting #828 resolved...

You will have hopefully seen #994.  As part of that, we would like to decouple the HTML parsing/rendering from the processing.  Doing so would make it easier to resolve #828 more cleanly.  You would be more than welcome to contribute to the discussion there, particularly on the aspect of decoupling the HTML parsing/rendering.

Cheers,
Jake

bytestream:
Hey Jake!

Thanks for doing some work on this.

> It seems only the specific case of <table> being a child of <a> fails

[...]

PR #1053, opened 2021-04-30
[TASK] Always run the CI jobs for all supported PHP versions

oliverklee:
This makes sure that we know exactly which PHP versions are affected
by change that has broken something.


JakeQZ (approved):

JakeQZ:
It can now be seen clearly that #1051 is an issue for all PHP versions with both lowest and latest dependencies.

PR #1057, opened 2021-05-04
[TASK] Allow case insensitivity in CSS RGB values

JakeQZ:
When testing for matching CSS, allow RGB values in CSS property declarations to
be upper or lower case.

### DIFF
--- a/tests/Support/Traits/CssDataProviders.php
+++ b/tests/Support/Traits/CssDataProviders.php
@@ -91,7 +91,18 @@ trait CssDataProviders
             $equivalentCssPropertyDeclarations
         );
 
-        return $datasetsWithPropertyDelcaration;
+        $equivalentPropertyDeclarationsWithRgbValue = [
+            'property declaration with lowercase RGB value' => ['color: #0f0;'],
+            'property declaration with uppercase RGB value' => ['color: #0F0;'],
+        ];
+
+        /** @var array<string, array{0: string, 1: string}> $datasetsWithPropertyDelcarationWithRgbValue */
+        $datasetsWithPropertyDelcarationWithRgbValue = DataProviders::cross(
+            $equivalentPropertyDeclarationsWithRgbValue,
+            $equivalentPropertyDeclarationsWithRgbValue
+        );
+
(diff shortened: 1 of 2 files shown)


tests/Support/Traits/CssDataProviders.php
```diff
@@ -91,7 +91,18 @@ public function provideEquivalentCssComponents(): array
             $equivalentCssPropertyDeclarations
         );
 
-        return $datasetsWithPropertyDelcaration;
+        $equivalentPropertyDeclarationsWithRgbValue = [
+            'property declaration with lowercase RGB value' => ['color: #0f0;'],
+            'property declaration with uppercase RGB value' => ['color: #0F0;'],
```
oliverklee, line 96: Please let's test with 6-digit color codes as well.

oliverklee (changes requested):

tests/Support/Traits/CssDataProviders.php
oliverklee: Oh, and maybe we also need to research whether "RED" also is the same as "red" or "Red".

JakeQZ: > 
> 
> Please let's test with 6-digit color codes as well.

Done.  And for completeness also check that all of the 6 hex digits A-F are treated case insensitively.

> 
> 
> Oh, and maybe we also need to research whether "RED" also is the same as "red" or "Red".

I only want to relax the CSS comparison tests as much as necessary to allow existing tests to pass where we see the CSS being rendered syntacticly differently but functionally equivalently.

I think `RED` and `red` would be equivalent as a CSS property value (since CSS syntax is mostly case insensitive, except for parts that are not under the control of CSS such as HTML `class` and `id` attributes, font names and URIs).

However, we don't currently have any tests for which the Sabberworm CSS parser would render colour names differently-cased from the current expected results.  Thus there is currently no need to allow for this for tests to pass.  We can continue to make the stronger assertions that they are the same and the same case.

(I'm aware of a case of a unit `PX` being converted to lowercase `px`, but we can address that separately from this PR.)

oliverklee (approved):

Issue #1054, opened 2021-05-01
Adjust the `.editorconfig`


oliverklee:
Related to #1038.

I've done some research concerning Markdown line lengths and XML indention.

Markdown line length limit:
- GitLab: [100 characters](https://docs.gitlab.com/ee/development/documentation/styleguide/#text)
- GitLab: [complete paragraph in one single line](https://about.gitlab.com/handbook/markdown-guide/#paragraphs-breaks-and-horizontal-lines)
- GitLab: [80-100 characters](https://about.gitlab.com/blog/2016/10/11/wrapping-text/)
- TYPO3 Core markdown files: [80 characters](https://raw.githubusercontent.com/TYPO3/TYPO3.CMS/master/README.md) (except when longer due to long URLs)

XML indentation:
- PHPUnit configuration file generation: 4 spaces (independent of `.editorconfig` settings)
- PHIVE configuration file update on `phive update`: 2 spaces (independent of `.editorconfig` settings)

So we should use 2 spaces indentation for XML, and use 80 characters line length for Markdown.

JakeQZ:
Somewhat alluded to in the above links is the idea of wrapping prose after every sentence, as this allows for cleaner commits when making minor alterations to it.
This could equally be applied to markdown, HTML, and DocBlock and other comments.
In fact this is what I have been doing with HTML.
However, this comment is not actually rendered as intended because [GFM](https://duckduckgo.com/?q=github+flavored+markdown) in some instances (e.g. release notes on GitHub, and of course issue comments) will render line breaks literally.

The line length argument seems to be based on there being _some_ upper limit to avoid extremely long lines and allow for easier review (i.e. with review comments per line), rather than what, specifically, it is exactly.  The same applies to PHP (and other) code, where we enforce a limit of 120 characters (via PHPCS).  Without a really compelling argument for a different line length for different file types, I would favour project-wide consistency, so that editors can be set up (with the help of `.editorconfig`) with suitable window sizes etc. for the project.

The GFM issue may suggest not wrapping at all in markdown that is rendered by GitHub, but this only seems to be an issue with release notes, not with the README or viewing other `.md` files within the source tree.  Maybe there's a flag somewhere to control the behaviour.

JakeQZ:
[...]

PR #1059, opened 2021-05-06
[CLEANUP] Remove URL reference from DocBlock

JakeQZ:
The description was initially copied-and-pasted from a similar DocBlock, but its
new home has nothing to do with URLs.

### DIFF
--- a/tests/Support/Constraint/CssConstraint.php
+++ b/tests/Support/Constraint/CssConstraint.php
@@ -49,9 +49,9 @@ abstract class CssConstraint extends Constraint
     /x';
 
     /**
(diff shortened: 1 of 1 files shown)


JakeQZ:
This commit may be cleaner due to already having newlines after sentences in the DocBlock - see https://github.com/MyIntervals/emogrifier/issues/1054#issuecomment-830716841

That was not deliberate.  WinMerge copes well with 'diffing' changes where odd words are moved onto the next line.  GitHub does not.  But I would say that the choice and availability of tools should not compromise other factors.

oliverklee (approved):

PR #1062, opened 2021-05-20
[TASK] Allow alternative style attribute formatting

JakeQZ:
Modify the test `inlineCssFormatsCssDeclarations` to allow removal of line
breaks between 'shadows' in the `text-shadow` property value, as there is no
requirement to preserve them.  This allows using a third party CSS parser which
may remove them.

### DIFF
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -8,6 +8,7 @@ use Pelago\Emogrifier\CssInliner;
 use Pelago\Emogrifier\HtmlProcessor\AbstractHtmlProcessor;
 use Pelago\Emogrifier\Tests\Support\Traits\AssertCss;
 use Pelago\Emogrifier\Utilities\CssDocument;
+use PHPUnit\Framework\Constraint\Constraint;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\CssSelector\CssSelectorConverter;
 use Symfony\Component\CssSelector\Exception\SyntaxErrorException;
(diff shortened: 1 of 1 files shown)


tests/Unit/CssInlinerTest.php
```diff
@@ -1369,10 +1370,12 @@ public function formattedCssDeclarationDataProvider(): array
             'one declaration with linefeed in property value' => [
                 "text-shadow:\n1px 1px 3px #000,\n1px 1px 1px #000;",
                 "text-shadow: 1px 1px 3px #000,\n1px 1px 1px #000;",
+                'text-shadow: 1px 1px 3px #000,1px 1px 1px #000;',
             ],
             'one declaration with Windows line ending in property value' => [
                 "text-shadow:\r\n1px 1px 3px #000,\r\n1px 1px 1px #000;",
                 "text-shadow: 1px 1px 3px #000,\r\n1px 1px 1px #000;",
+                'text-shadow: 1px 1px 3px #000,1px 1px 1px #000;',
```
oliverklee, line 1378: Those two don't seem to fit in the "… with (linefeed|Windows line ending) in property value" lists as they have neither. Maybe move them somewhere else? Or are the line endings missing?

oliverklee (changes requested):

tests/Unit/CssInlinerTest.php
JakeQZ: The first element is the property declaration for the input CSS.  The remaining element(s) are (possibilities for) the expected style attribute value.  We don't need to require that the line endings within the property value after the comma are preserved, so can allow the style attribute to be rendered with or without them.  (Currently they are preserved, but with Sabberworm's CSS parser they would be removed.)

oliverklee (approved):

PR #1067, opened 2021-05-23
[CLEANUP] Extract method to self-cross datasets

JakeQZ:
Though merely expanding one parameter into two duplicates, this allows an
intermediate variable to be eliminated (or, rather, moved to the new method)
where it was/is required to provide type information for static analysis (e.g.
Psalm).

### DIFF
--- a/tests/Support/Traits/CssDataProviders.php
+++ b/tests/Support/Traits/CssDataProviders.php
@@ -27,7 +27,7 @@ trait CssDataProviders
      */
     public function provideEquivalentCompleteCss(): array
     {
-        $equivalentCssWithAtMediaRuleSelectorListAndPropertyDeclaration = [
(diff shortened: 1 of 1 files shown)


tests/Support/Traits/CssDataProviders.php
```diff
@@ -27,7 +27,7 @@ public function provideEquivalentCss(): array
      */
     public function provideEquivalentCompleteCss(): array
     {
-        $equivalentCssWithAtMediaRuleSelectorListAndPropertyDeclaration = [
+        $datasetsWithAtMediaRuleSelectorListAndPropertyDeclaration = static::crossDatasetWithItself([
```
oliverklee, line 30: Could we also use `self::` instead of `static::` in all changed places?

JakeQZ: We can.  Have changed it.

oliverklee (approved):

PR #1070, opened 2021-05-24
[TASK] Allow single or double quotes in CSS

JakeQZ:
This is primarily currently targeted at CSS property values, allowing single
quotes to be exchanged for double quotes or vice versa (but not removed or
added), provided there are no quotes within the string.

### DIFF
--- a/tests/Support/Constraint/CssConstraint.php
+++ b/tests/Support/Constraint/CssConstraint.php
@@ -52,8 +52,13 @@ abstract class CssConstraint extends Constraint
             ([^\'"\\(\\)\\s]++)             #   quotes, parentheses or whitespace
             \\g{-2}                         #
             \\s*+\\)                        #
+        |(?<=[\\s:])                        # - singly or doubly quoted string, surrounded by whitespace or delimiters
+            ([\'"])                         #   of a property value (i.e. a colon is allowed before, and a semicolon or
+            ([^\'"]*+)                      #   closing curly brace is allowed after), with the quote captured in group
+            \\g{-2}                         #   13 and the string in group 14
+            (?=[\\s;\\}])                   #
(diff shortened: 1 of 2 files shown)


JakeQZ:
I don't know what to do about the PHPCS cyclomatic complexity warning, if anything other than disabling it for this specific method.

Changing the `if`...`else`s to a `switch` (after determining the index of the first nonempty `$matches` entry) doesn't help.

The only thing I can think of would be to create an array of functions, one for each `$matches` index...

oliverklee:
About the cyclomantic complexity: We can have an array `array-key => $regularExpressionEquivalent`, and if the correspondig array element is non-empty, we use the corresponding `regularExpressionEquivalent`.

JakeQZ:
> 
> 
> About the cyclomantic complexity: We can have an array `array-key => $regularExpressionEquivalent`, and if the correspondig array element is non-empty, we use the corresponding `regularExpressionEquivalent`.

I think there's a way of doing something like that without constructing the mostly unused replacements each time, e.g. if the replacements use `$1`, `$2` etc. (as they would if using `preg_replace` rather than `preg_replace_callback`) for the matched parts that need to be `preg_quote`d, which are then substituted accordingly when that replacement is selected...

JakeQZ:
> I think there's a way of doing something like that without constructing the mostly unused replacements each time, e.g. if the replacements use `$1`, `$2` etc. (as they would if using `preg_replace` rather than `preg_replace_callback`) for the matched parts that need to be `preg_quote`d, which are then substituted accordingly when that replacement is selected...

See #1071.

JakeQZ:
I've rebased with #1017 now merged, and reimplemented this to extend the new array instead of the old `elseif`s.

oliverklee (approved):

PR #1071, opened 2021-05-24
[CLEANUP] Use array of CSS regex replacements

JakeQZ:
This is instead of a (now) long sequence of `elseif`s, and reduces cyclomantic
complexity.  The replacements are expanded so that references to capturing
groups (like `$1`) are replaced with the `preg_quote`d match actually captured.

### DIFF
--- a/tests/Support/Constraint/CssConstraint.php
+++ b/tests/Support/Constraint/CssConstraint.php
@@ -86,6 +86,36 @@ abstract class CssConstraint extends Constraint
     /x';
 
     /**
+     * @var string
+     */
+    private const URL_REPLACEMENT_MATCHER = '(?:([\'"]?+)$8$10\\g{-1})';
+
+    /**
+     * @var string
+     */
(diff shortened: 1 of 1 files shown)


tests/Support/Constraint/CssConstraint.php
```diff
@@ -134,26 +164,22 @@ protected static function getCssRegularExpressionMatcher(string $css): string
...
+                    static function (array $referenceMatches) use ($matches): string {
+                        return \preg_quote($matches[(int)$referenceMatches[1]] ?? '', '/');
+                    },
+                    $replacement
+                );
+                break;
+            }
+        }
+
+        if ($regularExpressionEquivalent === null) {
```
oliverklee, line 182: I'd prefer `!is_string` here to make this check more specific (for the benefit of the reader and static code analysis).

JakeQZ: OK.  Changed.

oliverklee (approved):

PR #1074, opened 2021-05-25
[TASK] Use CSS assertions in some more tests

JakeQZ:
This allows some additional flexibility where the CSS may be rendered slightly
differently (but functionally equivalently) if a different CSS parser is used.

### DIFF
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -1518,10 +1518,9 @@ final class CssInlinerTest extends TestCase
 
         $subject->inlineCss('p {' . $cssDeclaration . '}');
 
-        self::assertStringContainsString(
-            '<p style="' . $cssDeclaration . '">target</p>',
-            $subject->renderBodyContent()
-        );
+        $result = $subject->renderBodyContent();
+        self::assertStringContainsString('<p style=', $result);
+        self::assertContainsCss($cssDeclaration, $result);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tests/Unit/CssInlinerTest.php
```diff
@@ -2002,7 +2000,7 @@ public function inlineCssWithValidMinifiedMediaQueryContainsInnerCss(string $css
 
         $subject->inlineCss($css);
 
-        self::assertStringContainsString('<style type="text/css">' . $css . '</style>', $subject->render());
+        self::assertContainsCss('<style type="text/css">' . $css . '</style>', $subject->render());
```
oliverklee, line 2003: `assertContainsCss` expects only the CSS declaration, but not the `style` node, if I understand this correctly. Or am I mistaken?

JakeQZ: It can cope with the `<style>` element as well - e.g. https://github.com/MyIntervals/emogrifier/blob/main/tests/Support/Constraint/CssConstraint.php#L32-L33.

Mostly it would just allow some whitespace differences within it, unless any parts of it (such as attribute values) looked very much like parts of CSS where variation is catered for.

JakeQZ: PS. There's a data provider [`provideEquivalentCssInStyleTags`](https://github.com/MyIntervals/emogrifier/blob/main/tests/Support/Traits/CssDataProviders.php#L142) used in tests such as [`StringContainsCssTest::matchesHaystackContainingNeedle`](https://github.com/MyIntervals/emogrifier/blob/main/tests/Unit/Support/Constraint/StringContainsCssTest.php#L21-L31).

oliverklee: Okay, thanks! It looks like we should improve the documentation in that regard. I've created #1075 to track this.

oliverklee (approved):

PR #1081, opened 2021-06-08
[BUGFIX] Allow @import after invalid @charset

JakeQZ:
From the spec: "User agents must ignore any `@import` rule that occurs inside a
block or after any non-ignored statement other than an `@charset` or an
`@import` rule."

This implies that an invalid `@charset` rule would be an "ignored statment" and
thus not in itself mean that any following `@import` rules must be ignored.

Since valid `@charset` rules are currently simply discarded (only UTF-8 is
supported), the change is to also (similarly) dicard invalid `@charset` rules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Drop support for PHP 7.1 (#967)
 
 ### Fixed
+- Allow `@import` after ignored invalid `@charset` (@1081)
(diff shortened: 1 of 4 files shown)


JakeQZ:
> 
> 
> I think some tests should be added to `CssDocumentTest` - though they might somewhat overlap with tests in `CssInlinerTest`, `CssDocument` is the class whose behaviour is changed here.

Now added and repushed.

oliverklee (approved):

Issue #1078, opened 2021-05-30
Support @supports


JakeQZ:
This is the de-facto standard now for determining 'browser' or email client CSS support for a specific feature.

We should probably process these CSS rules as we do `@media` rules so that their placement in the `<style>` element containing the uninlinable CSS is correct (and those that can't possibly apply can be eliminated).

If anyone can think of anything more clever we can do with `@supports` rules, please send your suggestions on a postcard.

PR #1082, opened 2021-06-09
[CLEANUP] Refactor into a new `StyleRule` class

oliverklee:


oliverklee:
I'll also particularly appreciate feedback concerning the naming of things and the namespace.

phpcs.xml
```diff
@@ -139,7 +139,6 @@
     <rule ref="PEAR.WhiteSpace.ScopeClosingBrace"/>
     <rule ref="Squiz.WhiteSpace.CastSpacing"/>
     <rule ref="Squiz.WhiteSpace.LogicalOperatorSpacing"/>
-    <rule ref="Squiz.WhiteSpace.OperatorSpacing"/>
```
JakeQZ, line 142: Yep, this is a PITA as it doesn't like the ternary operator split onto multiple lines.  It's supposed to ensure that there is a space between operator and operand (which we want to enforce) but seems buggy with expressions over multiple lines.  We should probably push for a fix in PHPCS so that we can still use it for single-line expressions.

src/CssParsing/StyleRule.php
```diff
@@ -0,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\CssParsing;
```
JakeQZ, line 5: I prefer to name classes based on what their objects represent, rather than how they might be used.

So I would go for just `Css`.  We could also move the `CssDocument` class into this namespace, renaming it to just `Document`.  (It is currently only in `Utilities` because the wasn't an obvious namespace for it at the time.)

src/CssParsing/StyleRule.php
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * This class represents a CSS style rule, including a media query, selectors and a declarations block.
+ *
+ * @internal
+ */
+class StyleRule
```
JakeQZ, line 12: [Definitely the right name I think.](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleRule)

src/CssParsing/StyleRule.php
```diff
@@ -0,0 +1,86 @@
...
+{
+    /**
+     * @var string
+     */
+    private $mediaQuery;
```
JakeQZ, line 17: Technically it's not just the media query, it also contains the `"@media "` preceding it.

But I think we can be more generic here (see #1078).  It could be any enclosing at-rule.  Based on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/At-rule)'s nomenclature, it's 'at, identifier and rule'.  In context, it's the 'enclosing', 'parent' or 'nesting' one of those.

src/CssParsing/StyleRule.php
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * @var array<int, string>
+     */
+    private $selectors;
+
+    /**
+     * @var string
+     */
+    private $declarationsBlock;
```
JakeQZ, line 27: These can be replaced with a single [`DeclarationBlock`](https://github.com/sabberworm/PHP-CSS-Parser/blob/master/src/RuleSet/DeclarationBlock.php) property once #1015 is merged.

src/CssParsing/StyleRule.php
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * @param string $mediaQuery the media query, e.g. `@media screen and (max-width: 480px)`, or an empty string
+     * @param string $selectors the comma-separated selectors, e.g., `*` or `h1, h2`, must not be empty
+     * @param string $declarationsBlock the comma-separated selectors, e.g., `*` or `h1, h2`, may be empty
```
JakeQZ, line 32: Ditto the comments for the corresponding properties.

src/CssParsing/StyleRule.php
```diff
@@ -0,0 +1,86 @@
...
+
+        $this->mediaQuery = \trim($mediaQuery);
+        $this->selectors = \array_map('trim', \explode(',', $selectors));
+        $this->declarationsBlock = \trim($declarationsBlock);
+    }
+
+    /**
+     * @returns string the media query, e.g. `@media screen and (max-width: 480px)`, or an empty string
+     */
+    public function getMediaQuery(): string
```
JakeQZ, line 50: See above comment.

src/CssParsing/StyleRule.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        return $this->mediaQuery;
+    }
+
+    /**
+     * Checks whether the media query is non-empty and has any non-whitespace characters.
+     */
+    public function hasNonEmptyMediaQuery(): bool
```
JakeQZ, line 58: See above comment.

src/CssParsing/StyleRule.php
```diff
@@ -0,0 +1,86 @@
...
+     */
+    public function getSelectors(): array
+    {
+        return $this->selectors;
+    }
+
+    /**
+     * @return string the semicolon-separated CSS declarations, e.g., `color: red; height: 4px;`
+     */
+    public function getDeclarationsBlock(): string
```
JakeQZ, line 74: I think it should be "declaration block" (singlular).  [A variety of other descriptions are available](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration), though use of the term 'map' would be misleading if the data is raw and not actually arranged as key-value pairs.

JakeQZ (commented):
I was thinking that after #1015 is completed, we should, moving forwards, be able to use the provided [`DeclarationBlock`](https://github.com/sabberworm/PHP-CSS-Parser/blob/master/src/RuleSet/DeclarationBlock.php) class further downstream to replace the associative arrays.  However, it doesn't AFAICS have a back-link to the parent `@media` (or other containing) rule, if any, so we would still need something to link the two.

I would actually suggest waiting until #1015 is completed before moving forward with this, because it should then be possible to replace the selector and declaration block properties with a single object of the class from Sabberworm.  But perhaps I am biased with a vested interest in not wanting to have to merge this into that ;))

src/CssParsing/StyleRule.php
JakeQZ: This then brings about the question: couldn't `CssInliner` be called just `Inliner` and exist in the `Css` namespace?

It doesn't feel like it belongs there, or that it should be renamed such.  It's an active class.  These other classes are passive, data only.  So perhaps the namespace should be `Data\Css` - though I don't like `Data` as a namespace name, it seems (for want of a better word) naff.

JakeQZ: > I prefer to name classes based on what their objects represent, rather than how they might be used.

Just to clarify, active classes can be named based on whay they (themselves) do, whereas passive (data-only) classes should be named based on what they represent (or store as data).

phpcs.xml
JakeQZ: Can this be a separate PR, and an issue report against PHPCS?

src/CssParsing/StyleRule.php
JakeQZ: I went with `containingAtRule` in #1015, though am not completely happy with that.  It implies it's an object when in fact it's just a string.  But I'm not doing Microsoft-style incorporation of types into variable names, whether pre- or post-fix.  There should be a way to describe that part of the rule (without it's nested content).

phpcs.xml
oliverklee: I have reported the issue now: https://github.com/squizlabs/PHP_CodeSniffer/issues/3377

Will create a separate PR in a minute.

src/CssParsing/StyleRule.php
oliverklee: Moved.

Concerning `CssInliner`: The class does not represent something that "is" CSS (like the other classes in the `Css` namespace). However, it does something with the HTML, and hence would belong into the `HtmlProcessor` namespace, I think.

src/CssParsing/StyleRule.php
oliverklee: Done.

src/CssParsing/StyleRule.php
oliverklee: Done.

src/CssParsing/StyleRule.php
JakeQZ: > However, it does something with the HTML, and hence would belong into the `HtmlProcessor` namespace, I think.

I think so too; the other classes there are like its siblings.  However, moving it would have a backwards compatibility issue.  Given #994 pending, I would avoid making such a change right now.  (I would also be keen to put in backwards-compatibility workarounds for any otherwise breaking change whenever possible - in the case of a renamed class, that would be a class with the old name that simply extends the newly renamed class - but wouldn't want such workarounds to pile up unnecessarily by renaming things on a whim without a clear plan.)

src/CssParsing/StyleRule.php
oliverklee: Done.

src/Css/StyleRule.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * @returns string e.g. `@media screen and (max-width: 480px)`, or an empty string
+     */
+    public function getContainingAtRule(): string
+    {
+        return $this->containingAtRule;
+    }
+
+    /**
+     * Checks whether the containing at-rule is non-empty and has any non-whitespace characters.
```
JakeQZ, line 77: Indeed it won't have just whitespace because it is `trim`med in the constructor.  And I think it is implied from the DocBlock for `getContainingAtRule` that this won't return a string containing only whitespace.  I also note this is well tested 👍

tests/Unit/Css/StyleRuleTest.php
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * @test
+     *
+     * @param array<int, array{property: string, value: string}> $declarations
+     * @param string $expected
+     *
+     * @dataProvider provideDeclarations
+     */
+    public function getDeclarationAsTextReturnConcatenatedDeclarationsFromRules(
```
JakeQZ, line 89: Probably should be `Returns` rather than `Return` in the method name.

tests/Unit/Css/StyleRuleTest.php
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * @test
+     *
+     * @param array<int, string> $selectors
+     *
+     * @dataProvider provideSelectors
+     */
+    public function getSelectorsReturnsSelectorsProvidedInConstructor(array $selectors): void
```
JakeQZ, line 35: `To` might be preferable to `In` in the method name.  There is a similarly named method that needs changing (see below); the naming for each should be consistent.

tests/Unit/Css/StyleRuleTest.php
```diff
@@ -0,0 +1,182 @@
...
+        }
+        $styleRule = new StyleRule($declarationBlock, '');
+
+        self::assertSame($expected, $styleRule->getDeclarationAsText());
+    }
+
+    /**
+     * @test
+     */
+    public function getContainingAtRuleReturnsContainingAtRuleProvidedByConstructor(): void
```
JakeQZ, line 107: Should probably be `To` rather than `By` in the method name.

JakeQZ (changes requested):
Very good 👍

Just a couple of minor issues with the test method names.

src/Css/StyleRule.php
oliverklee: I'm note quite sure whether you're praising the method name or suggesting a change … :confused:

JakeQZ: This was more a comment on the internal method logic with `!== ''`, which could result in returning `true` for an at-rule consisting of only whitespace, if it were not `trim`med at some point (which it is).  But if it were to, the test `hasContainingAtRuleForEmptyContainingAtRuleReturnsFalse` would fail anyway, as would `getContainingAtRuleTrimsContainingAtRule`.

JakeQZ (approved):
Neat; nicely done.

src/Css/StyleRule.php
oliverklee: Ah, I understand.

I try to code and name the tests in a way that the assume as little as possible about the way the subject handles its data internally (i.e., to write the tests as black-boxy as possible).

PR #1083, opened 2021-06-09
[CLEANUP] Drop an unused naming rule from the PHPMD ruleset

oliverklee:
We actually think that long variable names can be a good thing,
and we do not want PHPMD to stop us from using them.

In order to avoid the impression of "we need to fix our code before
we can enable the commented-out rule", the rule now has been
removed from our PHPMD ruleset altogether.

### DIFF
--- a/config/phpmd.xml
+++ b/config/phpmd.xml
@@ -36,7 +36,6 @@
     <rule ref="rulesets/design.xml/DevelopmentCodeFragment"/>
 
     <!--<rule ref="rulesets/naming.xml/ShortVariable"/>-->
-    <!--<rule ref="rulesets/naming.xml/LongVariable"/>-->
     <rule ref="rulesets/naming.xml/ShortMethodName"/>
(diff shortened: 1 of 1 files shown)


JakeQZ (commented):
These were commented-out anyway, so not in force.

Nowadays long variable names have a pretty-much undetectable effect on performance, so even the traditional loop variable `$i` can be renamed to indicate what exactly it is an index into.

However, can't we enable the `ShortVariable` rule?

JakeQZ:
> Naming things is a style-related issue and hence should be covered by
> PHP_CodeSniffer, not PHPMD.

OK.  But does PHPCS cover this?

oliverklee:
> OK. But does PHPCS cover this?

I had assumed that it does, but I have checked and have not been able to find a corresponding rule.

oliverklee:
> However, can't we enable the `ShortVariable` rule?

We indeed can, and I have: #1088

oliverklee:
I have update the PR accordingly.

JakeQZ (approved):

PR #1086, opened 2021-06-10
[CLEANUP] Move the internal `CssDocument` class

oliverklee:


src/CssInliner.php
```diff
@@ -179,7 +179,7 @@ public function inlineCss(string $css = ''): self
         if ($this->isStyleBlocksParsingEnabled) {
             $combinedCss .= $this->getCssFromAllStyleNodes();
         }
-        $parsedCss = new CssDocument($combinedCss);
+        $parsedCss = new Document($combinedCss);
```
JakeQZ, line 182: I think it should perhaps be `use`d `as` `CssDocument`.  Reading through the code makes one wonder: "what sort of document?"

tests/Unit/Css/DocumentTest.php
```diff
@@ -33,7 +33,7 @@ final class CssDocumentTest extends TestCase
     public function parsesSelector(string $selector): void
     {
         $css = $selector . '{ color: green; }';
-        $subject = new CssDocument($css);
+        $subject = new Document($css);
```
JakeQZ, line 36: Though in the `TestCase` I think it's fine to `use` it `as` `Document`, as it's quite clear this is the class under test.

tests/Unit/CssInlinerTest.php
```diff
@@ -3759,7 +3759,7 @@ public function copyUninlinableCssToStyleNodeHasNoSideEffects(): void
         $copyUninlinableCssToStyleNode->setAccessible(true);
 
         $domDocument = $subject->getDomDocument();
-        $cssDocument = new CssDocument('');
+        $cssDocument = new Document('');
```
JakeQZ, line 3762: Probably here `use` `as` `CssDocument` because it's not the class under test.

JakeQZ (commented):
Glad you agree on the `Css` namespace.

However, I am thinking when `use`d in other classes, code using `Document` lacks self-explanation.  Perhaps it should be `use`d `as` `CssDocument` by classes in other namespaces - ?

Otherwise changes look fine.

oliverklee:
Then let's go for `CssDocument` in all places.

JakeQZ:
> 
> 
> Then let's go for `CssDocument` in all places.

I would actually suggest that within the `Css` namespace the class is simply named `Document` because the second `Css` is redundant.  It is only where it is `use`d that there may need to be more clarity.  For example, see https://github.com/MyIntervals/emogrifier/blob/task/sabberworm-css-parser/src/Utilities/CssDocument.php#L7-L15.

oliverklee:
I'd prefer to keep it as `CssDocument` as we in Emogrifier need to deal with two kinds of documents all the time: the DOM document, and then CSS documents. Having the class named `CssDocument` would take this into account. (I.e., first and foremost, the naming needs to be helpful in the context of this project.)

JakeQZ:
> 
> 
> I'd prefer to keep it as `CssDocument` as we in Emogrifier need to deal with two kinds of documents all the time: the DOM document, and then CSS documents. Having the class named `CssDocument` would take this into account. (I.e., first and foremost, the naming needs to be helpful in the context of this project.)

The point I was making is that it already has `Css` in its name as part of the namespace, which would form it's FQN.  `…\Css\CssDocument` looks like an odd repetition of `Css`.  It's just where it's used (the context) that clarification would be helpful, which I think could be achieved with `as` in the `use` statement.

YMMV, but I have found that, e.g., placing ancestor and dependent classes in their own namespace, with the base class literally called `Base` (because its FQN accurately describes it), seems to work quite well.

Either approach is reasonable, and we can easliy rename internal classes later if there is a good reason.  Thus I am happy to approve this (as it stands) if I can't twist your arm - let me know either way.

oliverklee:
> Thus I am happy to approve this (as it stands) if I can't twist your arm - let me know either way.

Yes, please let's keep it as `CssDocument` for now.

(I'm going to rebase in a minute.)

oliverklee:
Rebased and repushed.

JakeQZ (approved):

PR #1087, opened 2021-06-12
Temporarily disable the `Squiz.WhiteSpace.OperatorSpacing` sniff

oliverklee:
This sniff has a bug that gets in the way of an upcoming refactoring
of our code. Disable the sniff for the time being until the bug
has been fixed.


oliverklee:
This isn't a bug, but a configuration problem on our side. Closing.

PR #1088, opened 2021-06-12
[CLEANUP] Enforce a minimal length for variable names

oliverklee:


src/CssInliner.php
```diff
@@ -419,9 +419,9 @@ private function normalizeStyleAttributes(\DOMElement $node): void
     {
         $normalizedOriginalStyle = \preg_replace_callback(
             '/-?+[_a-zA-Z][\\w\\-]*+(?=:)/S',
-            /** @param array<int, string> $m */
-            static function (array $m): string {
-                return \strtolower($m[0]);
+            /** @param array<int, string> $mixedCaseString */
+            static function (array $mixedCaseString): string {
+                return \strtolower($mixedCaseString[0]);
```
JakeQZ, line 424: Not sure about the new variable name.  I think originally `$m` stood for `$matches`.  It's not actually a string but a 1-length array of the regular expression matches (and I thought we were trying to avoid types within variable names). 

I would suggest something like `$propertyNameMatches` (because a property name is what is being matched here).

JakeQZ (changes requested):
One of the renamed variables could have a more accurate name, otherwise looks fine.

src/CssInliner.php
oliverklee: Done.

JakeQZ (approved):

PR #1089, opened 2021-06-12
[TASK] Also allow newlines before the `?` operator

oliverklee:


phpcs.xml
```diff
@@ -138,7 +138,11 @@
     <rule ref="PEAR.WhiteSpace.ObjectOperatorIndent"/>
     <rule ref="PEAR.WhiteSpace.ScopeClosingBrace"/>
     <rule ref="Squiz.WhiteSpace.CastSpacing"/>
-    <rule ref="Squiz.WhiteSpace.LogicalOperatorSpacing"/>
```
JakeQZ, line 141: Looks like the wrong rule is replaced.

phpcs.xml
```diff
@@ -138,7 +138,11 @@
...
+    <rule ref="Squiz.WhiteSpace.OperatorSpacing">
+        <properties>
+            <property name="ignoreNewlines" value="true"/>
+        </properties>
+    </rule>
     <rule ref="Squiz.WhiteSpace.OperatorSpacing"/>
```
JakeQZ, line 146: It should be this one being replaced.

JakeQZ (changes requested):
The wrong rule has been replaced - `LogicalOperatorSpacing` should be left unchanged (though I'm not sure what it does, it doesn't seem to be documented).

phpcs.xml
oliverklee: Good catch! Done.

JakeQZ (approved):

PR #1090, opened 2021-06-13
[TASK] Upgrade to PHP CS Fixer 3

oliverklee:


config/php-cs-fixer.php
```diff
@@ -20,30 +19,31 @@
...
             'is_null' => true,
             'linebreak_after_opening_tag' => true,
             'lowercase_cast' => true,
             'magic_constant_casing' => true,
-            'method_separation' => true,
             'modernize_types_casting' => true,
+            'multiline_whitespace_before_semicolons' => true,
             'native_function_casing' => true,
-            'native_function_invocation' => true,
+            'native_function_invocation' => ['include' => ['@internal']],
```
JakeQZ, line 46: To keep the previous behaviour, we should probably have `@all` rather than `@internal`.  Though we don't and are unlikely ever to have any global functions, it's possible we may use some from a third-party library.

config/php-cs-fixer.php
```diff
@@ -6,12 +6,11 @@
     die('This script supports command line usage only. Please check your command.');
 }
 
-return \PhpCsFixer\Config::create()
+return (new \PhpCsFixer\Config())
     ->setRiskyAllowed(true)
     ->setRules(
         [
             '@DoctrineAnnotation' => true,
-            '@PHP56Migration' => true,
```
JakeQZ, line 14: This was already covered by `array_syntax`.

JakeQZ (changes requested):
I've checked the config changes; all deprecated or changed rules are correctly replaced or updated - apart from (it appears) `native_function_invocation`.

config/php-cs-fixer.php
oliverklee: Done.

JakeQZ (approved):

PR #1091, opened 2021-06-15
[CLEANUP] Group the used PHP CS Fixer rules

oliverklee:
The grouping follows the list of sections in the documentation at
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/doc/rules/index.rst.

This should make the configuration file more manageable.

(This change neither removes any rules nor does it add any.)

### DIFF
--- a/config/php-cs-fixer.php
+++ b/config/php-cs-fixer.php
@@ -58,6 +58,7 @@ return (new \PhpCsFixer\Config())
             'single_line_comment_style' => true,
 
             // constant notation
+            // (no rules used from this section)
 
             // control structure
             'elseif' => true,
@@ -87,6 +88,7 @@ return (new \PhpCsFixer\Config())
             'is_null' => true,
 
             // list notation
+            // (no rules used from this section)
 
             // namespace notation
             'no_leading_namespace_whitespace' => true,
--- a/config/php-cs-fixer.php
+++ b/config/php-cs-fixer.php
@@ -52,6 +52,7 @@ return (new \PhpCsFixer\Config())
             'no_php4_constructor' => true,
 
             // class usage
+            // (no rules used from this section)
 
             // comment
             'no_empty_comment' => true,
@@ -94,6 +95,7 @@ return (new \PhpCsFixer\Config())
             'no_leading_namespace_whitespace' => true,
 
             // naming
+            // (no rules used from this section)
 
             // operator
             'concat_space' => ['spacing' => 'one'],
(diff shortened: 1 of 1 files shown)


oliverklee:
Oh, the fixer does not like this. This is kind of ironic. On it.

JakeQZ (commented):
> This should make the configuration file more manageable.

I think so.

> (This change neither removes any rules nor does it add any.)

I diffed with WinMerge (which has moved block detection), and can confirm this.  (And I think the expansion of `yoda_syle` into a multiline array makes it clearer, since it has three elements, whereas others with only one or two elements are perfectly readable single-line.  In terms of readability this actually suggests a coding standard: array definitions should be multiline if they comprise three or more elements, but otherwise single-line if possible.)

I was a bit confused initally about the comments with nothing immediately below.  I thought they might correspond to major section headings.  Though it is clear from the [doc](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/doc/rules/index.rst) that they are headings for empty lists.

It might be better to clarify that (e.g. by adding "(no rules used from this section)"), or not have the comments for unused sections at all.  WDYT?

oliverklee:
> It might be better to clarify that (e.g. by adding "(no rules used from this section)"), or not have the comments for unused sections at all. WDYT?

I'll go for the comments (as I plan to add more fixers to our rule set soon).

config/php-cs-fixer.php
```diff
@@ -20,59 +21,102 @@
...
-            'no_extra_blank_lines' => true,
-            'no_leading_import_slash' => true,
-            'no_leading_namespace_whitespace' => true,
             'no_php4_constructor' => true,
-            'no_short_bool_cast' => true,
-            'no_singleline_whitespace_before_semicolons' => true,
-            'no_spaces_after_function_name' => true,
-            'no_spaces_inside_parenthesis' => true,
+
+            // class usage
```
JakeQZ, line 54: Missing comment explaining comment.

config/php-cs-fixer.php
```diff
@@ -20,59 +21,102 @@
...
+            'dir_constant' => true,
+            'is_null' => true,
+
+            // list notation
+            // (no rules used from this section)
+
+            // namespace notation
+            'no_leading_namespace_whitespace' => true,
+
+            // naming
```
JakeQZ, line 96: Ditto.

JakeQZ (changes requested):
> I'll go for the comments (as I plan to add more fixers to our rule set soon).

OK.  But you missed a couple.

config/php-cs-fixer.php
oliverklee: Thanks, missed those two. Fixed and repushed.

JakeQZ (approved):

PR #1093, opened 2021-06-15
[TASK] Rework the `.editorconfig`

oliverklee:
- use sane defaults instead of redundant settings for each file type
- add a default line length limit for all types to avoid
  overeager line breaking by IDEs
- use shorter lines for Markdown to match what many projects use
- use a 2-space indent for XML files to match what PHIVE uses
  when it updates its configuration file
- drop redundant comments

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,34 +1,21 @@
 # top-most EditorConfig file
 root = true
 
-# Unix-style newlines with a newline ending every file
+# Unix-style newlines with a newline ending every file, and with same defaults
 [*]
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
-
-# PHP files
-[*.php]
 indent_style = space
 indent_size = 4
+max_line_length = 120
 
-# MD files
 [*.md]
-indent_style = space
-indent_size = 4
+max_line_length = 80
 
-# XML files
-[*.xml]
-indent_style = space
-indent_size = 4
-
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -1,34 +1,21 @@
 # top-most EditorConfig file
 root = true
 
-# Unix-style newlines with a newline ending every file
+# Unix-style newlines with a newline ending every file, and with same defaults
```
JakeQZ, line 4: Typo: should be "sane".  Unless you mean to say that every file type will have the same defaults, which kind-of goes without saying.

JakeQZ (changes requested):
Makes sense to move the sane/same defaults into `[*]`.  But I'm guessing you intended to call them "sane" in the comment - ?

.editorconfig
oliverklee: Fixed.

JakeQZ (approved):
> it's all the sane to me

😁

(Doesn't seem possible to add 'reaction' to a commit.)

JakeQZ:
> it's all the sane to me

Don't know if you know of Henning Wehn (anglicized German comedian), but that's the sort of thing he would say :)

oliverklee:
> Don't know if you know of Henning Wehn (anglicized German comedian), but that's the sort of thing he would say :)

I hadn't heard of him and have watched one of his performances right now. Quite funny, and very, very German. :-)

Issue #144, opened 2014-12-15
Refactor the cache-related methods


oliverklee:
We should have a field $cacheKeys.

komsan2537:
😀😀

PR #1097, opened 2021-07-11
[FEATURE] Add a simple string cache

oliverklee:
This class then will later be used to replace the current caching
that is based on nested arrays.

Part of #144.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -39,6 +39,8 @@
   <rule ref="Squiz.Commenting.DocCommentAlignment"/>
   <rule ref="Squiz.Commenting.EmptyCatchComment"/>
   <rule ref="Squiz.Commenting.FunctionComment">
+    <exclude-pattern>*/tests/*</exclude-pattern>
+
     <!-- Allow no comment for self-describing parameter and exception class names. -->
     <exclude name="Squiz.Commenting.FunctionComment.MissingParamComment"/>
     <exclude name="Squiz.Commenting.FunctionComment.EmptyThrows"/>
--- a/src/Caching/SimpleStringCache.php
+++ b/src/Caching/SimpleStringCache.php
@@ -17,6 +17,8 @@ namespace Pelago\Emogrifier\Caching;
  *     $cachedValue = $cache->get($value);
  * }
  * ```
+ *
+ * @internal
  */
 class SimpleStringCache
 {
@@ -34,7 +36,7 @@ class SimpleStringCache
      */
     public function has(string $key): bool
     {
-        $this->checkForNonEmptyKey($key);
+        $this->assertNotEmptyKey($key);
 
         return isset($this->values[$key]);
     }
@@ -68,7 +70,7 @@ class SimpleStringCache
      */
     public function set(string $key, string $value): void
     {
-        $this->checkForNonEmptyKey($key);
+        $this->assertNotEmptyKey($key);
 
         $this->values[$key] = $value;
     }
@@ -76,7 +78,7 @@ class SimpleStringCache
     /**
      * @throws \InvalidArgumentException
      */
-    private function checkForNonEmptyKey(string $key): void
+    private function assertNotEmptyKey(string $key): void
     {
         if ($key === '') {
             throw new \InvalidArgumentException('Please provide a non-empty key.', 1625995840);
--- a/src/Caching/SimpleStringCache.php
+++ b/src/Caching/SimpleStringCache.php
@@ -1,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\Caching;
+
+/**
+ * This cache caches string values with string keys. It is not PSR-6-compliant.
+ *
+ * Usage:
+ *
+ * ```php
+ * $cache = new SimpleStringCache();
+ * $cache->set($key, $value);
+ * …
+ * if ($cache->has($key) {
+ *     $cachedValue = $cache->get($value);
+ * }
+ * ```
+ */
+class SimpleStringCache
+{
+    /**
+     * @var array<string, string>
+     */
+    private $values = [];
+
+    /**
+     * Checks whether there is an entry stored for the given key.
+     *
+     * @param string $key the key to check; must not be empty
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function has(string $key): bool
+    {
+        $this->checkForNonEmptyKey($key);
+
+        return isset($this->values[$key]);
+    }
+
+    /**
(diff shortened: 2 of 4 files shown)


JakeQZ (commented):
Could the class implement [`ArrayAccess`](https://www.php.net/manual/en/class.arrayaccess.php) (instead of the `set`/`get`/`has` methods)?  This would also mean fewer changes to the existing code which could continue to get, set and test the cached values using the `[]` operator and `isset`.

oliverklee:
As far as I understand it, this would mostly remove the benefit of this class over using a plain array:
- typing the keys to strings
- typing the values to strings
- having more "speaking" method names.

So I'd prefer to now implement `ArrayAccess` for these reasons.

JakeQZ:
> 
> 
>     * typing the keys to strings
> 
>     * typing the values to strings

Yeah, I wondered about that, and can't think of or find a solution.  Unlike C++, PHP does not have templates so it's not possible to `implment ArrayAccess<string, string>`.

src/Caching/SimpleStringCache.php
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * ```php
+ * $cache = new SimpleStringCache();
+ * $cache->set($key, $value);
+ * …
+ * if ($cache->has($key) {
+ *     $cachedValue = $cache->get($value);
+ * }
+ * ```
+ */
```
JakeQZ, line 20: Should the class be `@internal`?

src/Caching/SimpleStringCache.php
```diff
@@ -0,0 +1,85 @@
...
+     */
+    private $values = [];
+
+    /**
+     * Checks whether there is an entry stored for the given key.
+     *
+     * @param string $key the key to check; must not be empty
+     *
+     * @throws \InvalidArgumentException
+     */
```
JakeQZ, line 34: Should there be `@return bool` or is this superfluous as it does not provide any additional information?

src/Caching/SimpleStringCache.php
```diff
@@ -0,0 +1,85 @@
...
+    public function set(string $key, string $value): void
+    {
+        $this->checkForNonEmptyKey($key);
+
+        $this->values[$key] = $value;
+    }
+
+    /**
+     * @throws \InvalidArgumentException
+     */
```
JakeQZ, line 78: Again, I assume `@param string $key` would be superfluous so is not needed.

src/Caching/SimpleStringCache.php
```diff
@@ -0,0 +1,85 @@
...
+    private function checkForNonEmptyKey(string $key): void
```
JakeQZ, line 79: Wondering about the method name.  Would something like `assertNotEmptyKey` be more descriptive?  If `assert` may sound like conflating with test methods, perhaps `verify` or `confirm` (or something like) - these verbs all imply that an empty key will be rejected, whereas `checkFor` doesn't really say what will be done when the key is/isn't found to be empty.

tests/Unit/Caching/SimpleStringCacheTest.php
```diff
@@ -0,0 +1,153 @@
...
+        $this->expectExceptionCode(1625995840);
+        $this->expectExceptionMessage('Please provide a non-empty key.');
+
+        $this->subject->has('');
+    }
+
+    /**
+     * @test
+     */
+    public function hasInitiallyIsFalseForAnyKey(): void
```
JakeQZ, line 40: Perhaps `Returns` rather than `Is` in the method name - ?

tests/Unit/Caching/SimpleStringCacheTest.php
```diff
@@ -0,0 +1,153 @@
...
+
+        $this->subject->set($key, $value);
+
+        self::assertSame($value, $this->subject->get($key));
+    }
+
+    /**
+     * @test
+     */
+    public function getForKeyThatHasBeenSetTwiceReturnsValueThatHasBeenSetLast(): void
```
JakeQZ, line 130: Weirdly, it feels more natural to say `getForKeyThatHasBeenSetTwiceReturnsValueThatWasSetLast` even though the verb tenses don't seem to agree.  Though it could probably be named `getForKeyThatHasBeenSetTwiceReturnsValueSetLast` to more concisely convey the same meaning and dodge the grammar issue.

tests/Unit/Caching/SimpleStringCacheTest.php
```diff
@@ -0,0 +1,153 @@
...
+     */
+    public function setAndGetCanAlsoReturnEmptyString(): void
+    {
+        $key = 'hello';
+        $value = '';
+
+        $this->subject->set($key, $value);
+
+        self::assertSame($value, $this->subject->get($key));
+    }
```
JakeQZ, line 152: I think this should be combined with `getForKeyThatHasBeenSetReturnsSetValue`, with a data provider providing a non-empty string and an empty string.

JakeQZ (changes requested):
Looks generally good, and I can't think of any tests that are missing.

A few minor issues and questions.

oliverklee:
@JakeQZ I immensely appreciate your detailed reviews! :heart: On it.

phpcs.xml
```diff
@@ -39,6 +39,8 @@
   <rule ref="Squiz.Commenting.DocCommentAlignment"/>
   <rule ref="Squiz.Commenting.EmptyCatchComment"/>
   <rule ref="Squiz.Commenting.FunctionComment">
+    <exclude-pattern>*/tests/*</exclude-pattern>
+
```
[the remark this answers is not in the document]
oliverklee: As a separation between two kinds of `exclude`s: by pattern, and by name. I slightly prefer having the blank line here, but would be fine with removing it as well.

WDYT?

src/Caching/SimpleStringCache.php
oliverklee: I think it's superfluous.

JakeQZ: OK, I kind-of thought so too.  (Also, we have recently removed insistence on superfluous DocBlock annotations from the sniffs.  We could perhaps even insist on not having superfluous DocBlock annotations if there is a sniff for it.)

phpcs.xml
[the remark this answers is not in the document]
JakeQZ: Happy either way.  Just wanted to check it was deliberate and what the reason was.

JakeQZ (approved):
Lovely.  (Sorry for the delay getting back to you, been away quite a but and also busy on other projects.)

PR #1108, opened 2021-09-14
[CLEANUP] Fix Psalm warnings

oliverklee:


src/CssInliner.php
```diff
@@ -419,15 +419,14 @@ private function normalizeStyleAttributes(\DOMElement $node): void
     {
         $normalizedOriginalStyle = \preg_replace_callback(
             '/-?+[_a-zA-Z][\\w\\-]*+(?=:)/S',
-            /** @param array<int, string> $propertyNameMatches */
+            /** @param array<array-key, string> $propertyNameMatches */
```
JakeQZ, line 422: Strange that the newer Psalm doesn't like `int` as the array key - the spec for `preg_replace` says it is a numeric array.

JakeQZ (approved):

PR #1111, opened 2021-09-16
[TASK] Drop support for PHP 7.2

oliverklee:


JakeQZ (approved):

JakeQZ:
The CI build seems to be waiting for the PHP 7.2 tasks to be completed (which will never happen).  Is this a caching issue, or is there some other setting within GitHub that needs updating?

oliverklee:
That was a repository setting. I've fixed it now.

PR #1112, opened 2021-09-16
[TASK] Upgrade to PHPUnit 9

oliverklee:


JakeQZ (approved):
I seem to remember that we could not previously allow `^8.x || ^9.x` becuase of the warnings about deprecated assert methods, but maybe now we can also allow `9.x || 10.x` or even later versions of PHPUnit if there are no such similar issues - this would give us whatever benefits there are of the later versions when testing with compatible PHP versions, though I think beyond the scope of this initial PR.

PR #1113, opened 2021-09-16
[BUGFIX] Only support up to 69 atomic expressions in a selector

oliverklee:
Starting from 70 atomic expressions, PHP will run into the nesting level
limit that was recently added to libxml:

https://gitlab.gnome.org/GNOME/libxml2/-/commit/6f1470a5d6e3e369fe93f52d5760ba7c947f0cd1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 ### Changed
+- Only support up to 69 atomic expressions in a selector (#1113)
 - Require `sabberworm/php-css-parser:^8.4.0` (#1134)
 - Upgrade to PHPUnit 9 (#1112)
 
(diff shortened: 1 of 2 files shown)


oliverklee:
With numbers from 70 and up, I get a crash on my local machine when running the corresponding test.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 ### Changed
+- Only support an operator nesting level up to 69 (#1113)
```
JakeQZ, line 11: Internally in libxml it's a nesting level limit, but when translated back to what appears in the selector, the limit 69 is the number of atomic expressions in a selector (i.e. without combinators), so I think the description should reflect that.

JakeQZ (commented):
I think the description of this change could be clarified.

CHANGELOG.md
oliverklee: Updated accordingly, rebased and repushed.

JakeQZ (approved):

PR #1124, opened 2021-12-10
[TASK] Drop the special CI builds for Symfony 6

oliverklee:
Now that Symfony 6 has been released, it will be used by default
when our build tests with the highest possible dependencies.
So we do not need to have a special build for symfony 6.0.x-dev
anymore.


JakeQZ (changes requested):
This effectively undoes #1109.  However, I think the changelog entry from that needs to be updated.

JakeQZ (approved):
I have just noticed that the changelog entry "Test with Symfony 6-dev" applies to and was added before our 6.0.0 release, so we don't want to change or remove it in any way.  Nor do I think there is any need for a changelog entry about the change here to remove it - it is expected we continually test with latest versions (it just so happened that we released our 6.0.0 slightly before Symfony but wanted to be reasonably sure of compatibility).

JakeQZ:
I wonder if this will actually resolve our Symfony-related test failures - though if so I'd still like to know what's in the 6.x-dev branch that appears to break something.

JakeQZ:
> I wonder if this will actually resolve our Symfony-related test failures

No, they are still failing, and since #1134 appear also to be failing for `lowest` dependencies - which seems to suggest an issue actually introduced with CSS Parser 8.4 rather than Symfony.

PR #1125, opened 2021-12-10
[TASK] Migrate the PHPUnit configuration to the 9.5 format

oliverklee:


JakeQZ (approved):
This should result in the same behaviour AFAICT.

PR #1137, opened 2021-12-16
[TASK] Always run the unit tests in the same order

### LINKED ISSUE #1136: Disable PHPUnit test ordering
PHPUnit has some kind of 'AI' feature that runs tests in a non-logical order it thinks might be more useful, base on what happened on the last test run.

It isn't more useful.  It's unpredictable, clearly doesn't work properly, and is a PITA.

We need to add the command line switch (whatever it is) that disables this non-intelligence - unless you think you can work with it - ?

oliverklee:
Also disable the PHPUnit result caching as this is only relevant
for running failing tests first.

Fixes #1136

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -95,8 +95,8 @@
         "ci:tests": [
             "@ci:tests:unit"
         ],
-        "ci:tests:sof": "@php \"./vendor/bin/phpunit\" --stop-on-failure",
-        "ci:tests:unit": "@php \"./vendor/bin/phpunit\"",
+        "ci:tests:sof": "@php \"./vendor/bin/phpunit\" --stop-on-failure --do-not-cache-result",
+        "ci:tests:unit": "@php \"./vendor/bin/phpunit\" --do-not-cache-result",
         "composer:normalize": "@php \"./.phive/composer-normalize\"",
         "php:fix": "@php \"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix config/ src/ tests/",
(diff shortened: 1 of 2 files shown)


JakeQZ (commented):
I added a comment to #1136.  Not 100% sure we want to make this change, but, if, in light of reading that, you are happy to proceed with this change I am probably happy too.  Also wondering if both changes are needed, and why the original `executionOrder` of `depends,defects` seems to actually work like `defects,random` (assuming it works like goal-difference ranking - rank by this, then by that, then by the other).

composer.json
```diff
@@ -95,8 +95,8 @@
         "ci:tests": [
             "@ci:tests:unit"
         ],
-        "ci:tests:sof": "@php \"./vendor/bin/phpunit\" --stop-on-failure",
-        "ci:tests:unit": "@php \"./vendor/bin/phpunit\"",
+        "ci:tests:sof": "@php \"./vendor/bin/phpunit\" --stop-on-failure --do-not-cache-result",
+        "ci:tests:unit": "@php \"./vendor/bin/phpunit\" --do-not-cache-result",
```
JakeQZ, line 99: Yep, caching often causes problems if not absolutely perfectly thought out and implemented.  In our case, it is unlikely to save any significant CPU time, though we should always be mindful of the CO<sub>2</sub> footprint.

JakeQZ (approved):
You've not disagreed with my re-questioning of whether we should make this change, thus I approve.

Thanks for sorting this.

Issue #1129, opened 2021-12-13
Failing tests with CSS Parser 8.4


JakeQZ:
E.g. https://github.com/MyIntervals/emogrifier/runs/4501517317

```
There were 2 failures:

1) Pelago\Emogrifier\Tests\Unit\Css\CssDocumentTest::parsesSelector with data set "attribute (value match)" ('[href=https://example.org]')
Failed asserting that actual size 0 matches expected size 1.

/home/runner/work/emogrifier/emogrifier/tests/Unit/Css/CssDocumentTest.php:40

2) Pelago\Emogrifer\Tests\Unit\CssInlinerTest::inlineCssInDebugModeForInvalidCssSelectorThrowsException
Failed asserting that exception of type "Symfony\Component\CssSelector\Exception\SyntaxErrorException" is thrown.
```

Looks like something has changed in Symfony that we need to update for - the tests don't fail with "lowest" dependencies.

(Not sure why I've only just got the linked build report for a commit several days ago nor why it wasn't picked up by CI tests.)

JakeQZ:
> (Not sure why I've only just got the linked build report for a commit several days ago nor why it wasn't picked up by CI tests.)

Oh, we have some kind of weekly build, and it seems to be labelled based on the last commit rather than anything more useful.

I think it's just coincidence that _two_ 3rd-party libs have just been updated in a way which breaks our CI build.

oliverklee:
The first failure might be this bug: https://github.com/symfony/symfony/issues/44516

JakeQZ:
I confirmed these result from using CSS Parser 8.4 and pass with 8.3.1.

The second is because the invalid rule is now stripped out by CSS Parser so it never reaches Symfony for it to throw an exception.  Will need to look into what behaviour/debug settings are avaible so we can catch the exception.

JakeQZ:
The other looks very closely related but is a bug introduced into CSS Parser.  With the new version, the valid CSS rule `[href=https://example.org]{ color: green; }` is now being discarded.

I'm guessing a change has been made to discard genuininely invalid CSS selectors more in line with the spec, however in this case there is a false positive.

JakeQZ:
> is a bug

...unless I'm mistaken and `[href=https://example.org]` is actually an invalid selector, but AFAIK there is no requirement to quote the value in this case.

JakeQZ:
[...]

PR #1121, opened 2021-12-10
[TASK] Update the development tools

oliverklee:
Also reorder the entries in the `composer.json` to match the rules
of the latest composer-normalize version.

PR #1138, opened 2021-12-16
[BUGFIX] Disable tests that are failing due to a bug in the CSS parser

oliverklee:
As this is a bug in a 3rd-party library, we should fix it there, and
our build should not be red in the meantime.

Part of #1129

### DIFF
--- a/tests/Unit/Css/CssDocumentTest.php
+++ b/tests/Unit/Css/CssDocumentTest.php
@@ -66,7 +66,9 @@ final class CssDocumentTest extends TestCase
             'class' => ['.classy'],
             'ID' => ['#toc'],
             'attribute (presence)' => ['[title]'],
-            'attribute (value match)' => ['[href=https://example.org]'],
+            // disabled as this currently breaks due to a bug in the CSS parser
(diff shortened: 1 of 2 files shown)


tests/Unit/Css/CssDocumentTest.php
```diff
@@ -66,7 +66,9 @@ public function provideSelector(): array
             'class' => ['.classy'],
             'ID' => ['#toc'],
             'attribute (presence)' => ['[title]'],
-            'attribute (value match)' => ['[href=https://example.org]'],
+            // disabled as this currently breaks due to a bug in the CSS parser
+            // @see https://github.com/sabberworm/PHP-CSS-Parser/issues/347
+            // 'attribute (value match)' => ['[href=https://example.org]'],
```
JakeQZ, line 71: Surprised PHPCS isn't moaning that the comment is mostly code.

JakeQZ (approved):
Yep, [I think so too](https://github.com/MyIntervals/emogrifier/pull/1121#issuecomment-996343484).

Issue #1140, opened 2021-12-17
Re-enable tests temporarily disabled in #1138


JakeQZ:
See #1138

Issue #1139, opened 2021-12-17
Chain our debug setting through to CSS Parser's leniency setting


JakeQZ:
See https://github.com/sabberworm/PHP-CSS-Parser/issues/346.

We just need to pass on `CssInliner::$debug` through `Pelago\Emogrifier\Css\CssDocument` to `Sabberworm\CSS\Parser`, mostly with one line of code to be added:

```php
$parserSettings = \Sabberworm\CSS\Settings::create()->withLenientParsing(!$debug);
```

This will resolve one of the failing tests in #1129.

But life is never simple.  Simply applying the requried code change leads to a whole bunch of additional test failures, which need to be looked into.  On initial glance, quite a few are because we expect an exception for malformed CSS but would now be getting a different class of exception.  But there are other failures of a different nature too.

Issue #1135, opened 2021-12-14
Add more selector edge case tests


JakeQZ:
The first was discovered by another test class more by chance, and (I think) should be tested specfically by `CssInlinerTest`.  The second was [reported against Symfony](https://github.com/symfony/symfony/issues/44516).

- `[href=https://example.org]`
- `#test\:colon`

JakeQZ:
Though OTOH, we should trust the reliability of the 3rd party libraries (to parse the CSS and convert the selectors to XPaths) and thus adding our own tests is duplicating effort.

oliverklee:
This is connected that we're kind of mixing up unit tests and functional tests in our tests: Unit tests should test only one class (and how it interacts with the environment, while the 3rd-party code should not be executed), while functional tests should test multiple parts of the system together (with our code possibly calling 3rd-parts code).

As a more pragmatic approach, I propose we
1. first cover these edge cases in our tests
2. then contribute the corresponding tests to the 3rd-party libraries (in this case, the PHP CSS Parser project)
3. then remove most of the corresponding tests from our tests

WDYT?

JakeQZ:
I guess the situation we currently have has arisen because originally there was just one class.

As we have refactored functionality into other classes or 3rd-party libraries we have often retained the original tests whilst adding new test cases for the new classes.

The above approach makes sense.  There are probably quite a few test methods in CssInlinerTest that belong elsewhere.

JakeQZ:
Reviewing the tests in `CssInlinerTest` might suggest some poignant refactoring.

For one thing, I think we can move selector precedence calculation to a separate class and test that independently.

JakeQZ:
> move selector precedence calculation to a separate class

Possibly using PHP's [Comparable](https://wiki.php.net/rfc/comparable) `interface`.  Oh...

JakeQZ:
We should reorder the tests in `CssInlinerTest` based on functionality being tested - e.g. related to CSS parsing, converting selectors to XPath, ranking selector precedence, merging CSS properties, or various other bits and bobs I haven't quite identified in that list.

This will help us identify areas for refactoring.  It could be done as a single commit that changes nothing (though I wish GitHub had moved-block detection support like WinMerge, which would obviously make this easier - but as it's a commercial Microsoft product and we are not paying customers, I doubt we will ever get any movement there with a feature request).

I think this may be a reasonable approach towards making #994 more achievable, or even figuring out how it would be achieved.

[...]

PR #1142, opened 2021-12-21
[TASK] Pass debug setting on to CSS parser

JakeQZ:
This partially addresses both #1139 and #1140.

However, if the CSS contains nested at-rules, the debug setting won't currently
be passed on, due to https://github.com/sabberworm/PHP-CSS-Parser/issues/127.

Some tests have been adapted to cater for an exception in debug mode with only
some of the data.

### DIFF
--- a/src/Css/CssDocument.php
+++ b/src/Css/CssDocument.php
@@ -13,6 +13,7 @@ use Sabberworm\CSS\Property\Import as CssImport;
 use Sabberworm\CSS\Renderable as CssRenderable;
 use Sabberworm\CSS\RuleSet\DeclarationBlock as CssDeclarationBlock;
 use Sabberworm\CSS\RuleSet\RuleSet as CssRuleSet;
+use Sabberworm\CSS\Settings as ParserSettings;
 
 /**
  * Parses and stores a CSS document from a string of CSS, and provides methods to obtain the CSS in parts or as data
@@ -36,7 +37,7 @@ class CssDocument
     private $isImportRuleAllowed = true;
 
     /**
-     * @param string $css
+     * @param string $css {@see https://github.com/squizlabs/PHP_CodeSniffer/issues/3521}
      * @param bool $debug
      *        If this is `true`, an exception will be thrown if invalid CSS is encountered.
      *        Otherwise the parser will try to do the best it can.
@@ -44,15 +45,23 @@ class CssDocument
     public function __construct(string $css, bool $debug)
     {
         // CSS Parser currently throws exception with nested at-rules (like `@media`) in strict parsing mode
-        // @see https://github.com/sabberworm/PHP-CSS-Parser/issues/127
-        $parserSettings = \Sabberworm\CSS\Settings::create()->withLenientParsing(
-            !$debug ||
-            \preg_match('/@(?:media|supports|(?:-webkit-|-moz-|-ms-|-o-)?+(keyframes|document))\\b/', $css) === 1
-        );
+        $parserSettings = ParserSettings::create()->withLenientParsing(!$debug || static::hasNestedAtRule($css));
 
         // CSS Parser currently throws exception with non-empty whitespace-only CSS in strict parsing mode, so `trim()`
         // @see https://github.com/sabberworm/PHP-CSS-Parser/issues/349
         $this->sabberwormCssDocument = (new CssParser(\trim($css), $parserSettings))->parse();
+    }
+
+    /**
+     * Tests if a string of CSS appears to contain an at-rule with nested rules
+     * (`@media`, `@supports`, `@keyframes`, `@document`,
+     * the latter two additionally with vendor prefixes that may commonly be used).
+     *
+     * @see https://github.com/sabberworm/PHP-CSS-Parser/issues/127
+     */
+    private static function hasNestedAtRule(string $css): bool
+    {
+        return \preg_match('/@(?:media|supports|(?:-webkit-|-moz-|-ms-|-o-)?+(keyframes|document))\\b/', $css) === 1;
     }
 
     /**
--- a/src/Css/CssDocument.php
+++ b/src/Css/CssDocument.php
@@ -37,7 +37,7 @@ class CssDocument
     private $isImportRuleAllowed = true;
 
     /**
-     * @param string $css {@see https://github.com/squizlabs/PHP_CodeSniffer/issues/3521}
+     * @param string $css
      * @param bool $debug
      *        If this is `true`, an exception will be thrown if invalid CSS is encountered.
      *        Otherwise the parser will try to do the best it can.
@@ -45,7 +45,7 @@ class CssDocument
     public function __construct(string $css, bool $debug)
     {
         // CSS Parser currently throws exception with nested at-rules (like `@media`) in strict parsing mode
-        $parserSettings = ParserSettings::create()->withLenientParsing(!$debug || static::hasNestedAtRule($css));
+        $parserSettings = ParserSettings::create()->withLenientParsing(!$debug || $this->hasNestedAtRule($css));
 
         // CSS Parser currently throws exception with non-empty whitespace-only CSS in strict parsing mode, so `trim()`
         // @see https://github.com/sabberworm/PHP-CSS-Parser/issues/349
@@ -59,7 +59,7 @@ class CssDocument
      *
      * @see https://github.com/sabberworm/PHP-CSS-Parser/issues/127
      */
-    private static function hasNestedAtRule(string $css): bool
+    private function hasNestedAtRule(string $css): bool
     {
         return \preg_match('/@(?:media|supports|(?:-webkit-|-moz-|-ms-|-o-)?+(keyframes|document))\\b/', $css) === 1;
     }
--- a/src/Css/CssDocument.php
+++ b/src/Css/CssDocument.php
@@ -37,10 +37,22 @@ class CssDocument
 
     /**
      * @param string $css
+     * @param bool $debug
+     *        If this is `true`, an exception will be thrown if invalid CSS is encountered.
+     *        Otherwise the parser will try to do the best it can.
      */
-    public function __construct(string $css)
+    public function __construct(string $css, bool $debug)
     {
-        $this->sabberwormCssDocument = (new CssParser($css))->parse();
+        // CSS Parser currently throws exception with nested at-rules (like `@media`) in strict parsing mode
+        // @see https://github.com/sabberworm/PHP-CSS-Parser/issues/127
+        $parserSettings = \Sabberworm\CSS\Settings::create()->withLenientParsing(
+            !$debug ||
+            \preg_match('/@(?:media|supports|(?:-webkit-|-moz-|-ms-|-o-)?+(keyframes|document))\\b/', $css) === 1
+        );
+
+        // CSS Parser currently throws exception with non-empty whitespace-only CSS in strict parsing mode, so `trim()`
+        // @see https://github.com/sabberworm/PHP-CSS-Parser/issues/349
+        $this->sabberwormCssDocument = (new CssParser(\trim($css), $parserSettings))->parse();
     }
 
     /**
--- a/tests/Unit/Css/CssDocumentTest.php
+++ b/tests/Unit/Css/CssDocumentTest.php
@@ -24,6 +24,11 @@ final class CssDocumentTest extends TestCase
         . '  font-family: "Foo Sans";' . "\n"
         . '  src: url("/foo-sans.woff2") format("woff2");' . "\n}";
 
+    private static function createDebugSubject(string $css): CssDocument
+    {
+        return new CssDocument($css, true);
+    }
+
     /**
      * @test
      *
@@ -436,8 +441,6 @@ final class CssDocumentTest extends TestCase
     /**
      * @test
      *
-     * @param string $css
-     *
      * @dataProvider provideValidAtCharsetRules
      * @dataProvider provideInvalidAtCharsetRules
      */
@@ -453,8 +456,6 @@ final class CssDocumentTest extends TestCase
     /**
      * @test
      *
-     * @param string $css
-     *
      * @dataProvider provideValidAtCharsetRules
      */
     public function discardsValidAtCharsetRuleInDebugMode(string $css): void
@@ -468,8 +469,6 @@ final class CssDocumentTest extends TestCase
 
     /**
      * @test
-     *
-     * @param string $css
      *
      * @dataProvider provideInvalidAtCharsetRules
      */
@@ -512,9 +511,6 @@ final class CssDocumentTest extends TestCase
 
     /**
      * @test
-     *
-     * @param string $atRuleCss
-     * @param string $cssBefore
      *
      * @dataProvider provideInvalidNonConditionalAtRule
      */
@@ -588,11 +584,6 @@ final class CssDocumentTest extends TestCase
         $result = $subject->renderNonConditionalAtRules();
 
         self::assertSame('', $result);
-    }
-
-    private static function createDebugSubject(string $css): CssDocument
-    {
-        return new CssDocument($css, true);
     }
 
     /**
--- a/tests/Unit/Css/CssDocumentTest.php
+++ b/tests/Unit/Css/CssDocumentTest.php
@@ -441,8 +441,9 @@ final class CssDocumentTest extends TestCase
     /**
      * @test
      *
-     * @dataProvider provideValidAtCharsetRules
-     * @dataProvider provideInvalidAtCharsetRules
+     * @dataProvider provideValidAtCharsetRule
+     * @dataProvider provideInvalidAtCharsetRuleWhichCausesException
+     * @dataProvider provideInvalidAtCharsetRuleWhichDoesNotCauseException
      */
     public function discardsValidOrInvalidAtCharsetRuleNotInDebugMode(string $css): void
     {
@@ -456,7 +457,7 @@ final class CssDocumentTest extends TestCase
     /**
      * @test
      *
-     * @dataProvider provideValidAtCharsetRules
+     * @dataProvider provideValidAtCharsetRule
      */
     public function discardsValidAtCharsetRuleInDebugMode(string $css): void
     {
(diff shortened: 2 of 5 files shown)


src/Css/CssDocument.php
```diff
@@ -37,10 +37,22 @@ class CssDocument
...
      */
-    public function __construct(string $css)
+    public function __construct(string $css, bool $debug)
     {
-        $this->sabberwormCssDocument = (new CssParser($css))->parse();
+        // CSS Parser currently throws exception with nested at-rules (like `@media`) in strict parsing mode
+        // @see https://github.com/sabberworm/PHP-CSS-Parser/issues/127
+        $parserSettings = \Sabberworm\CSS\Settings::create()->withLenientParsing(
+            !$debug ||
+            \preg_match('/@(?:media|supports|(?:-webkit-|-moz-|-ms-|-o-)?+(keyframes|document))\\b/', $css) === 1
```
oliverklee, line 50: Please let's move this `preg_match` to a local variable with a name that communicates its meaning.

tests/Unit/Css/CssDocumentTest.php
```diff
@@ -437,15 +441,52 @@ public function provideCssWithoutNonConditionalAtRules(): array
...
 
         $result = $subject->renderNonConditionalAtRules();
 
         self::assertSame('', $result);
     }
 
+    /**
+     * @test
+     *
+     * @param string $css
```
oliverklee, line 456: Generally, please let's not have PHPDoc annotations that add nothing to what the native type declarations already communicate. (We probably still need to removed the redundant annotations IIRC.)

tests/Unit/Css/CssDocumentTest.php
```diff
@@ -516,13 +583,18 @@ public function provideInvalidNonConditionalAtRule(): array
...
     {
-        $subject = new CssDocument('@media screen { p { color: red; } }');
+        $subject = self::createDebugSubject('@media screen { p { color: red; } }');
 
         $result = $subject->renderNonConditionalAtRules();
 
         self::assertSame('', $result);
     }
 
+    private static function createDebugSubject(string $css): CssDocument
```
oliverklee, line 593: Please let's have this method at the top directly below `setUp`/`tearDown` as this  method is closely related to `setUp`.

tests/Unit/Css/CssDocumentTest.php
```diff
@@ -437,15 +441,52 @@ public function provideCssWithoutNonConditionalAtRules(): array
...
+     */
+    public function throwsExceptionForOrDiscardsInvalidAtCharsetRuleInDebugMode(string $css): void
+    {
+        try {
+            $subject = self::createDebugSubject($css);
+
+            $result = $subject->renderNonConditionalAtRules();
+
+            self::assertSame('', $result);
+        } catch (UnexpectedEOFException $exception) {
```
oliverklee, line 484: Please let's use `$this->expectException`, `expectExceptionCode` and `expectExceptionMessage` instead.

Ditto for other tests that check for exceptions.

tests/Unit/Css/CssDocumentTest.php
```diff
@@ -437,15 +441,52 @@ public function provideCssWithoutNonConditionalAtRules(): array
...
+     * @dataProvider provideInvalidAtCharsetRules
+     */
+    public function throwsExceptionForOrDiscardsInvalidAtCharsetRuleInDebugMode(string $css): void
+    {
+        try {
+            $subject = self::createDebugSubject($css);
+
+            $result = $subject->renderNonConditionalAtRules();
+
+            self::assertSame('', $result);
```
oliverklee, line 483: Will this line ever be reached?

oliverklee (changes requested):

tests/Unit/Css/CssDocumentTest.php
JakeQZ: Agree.  We've already removed the CI check that tried to enforce unnecessary DocBlock.  This was a result of copypasta.

I have hopefully now removed unnecessary DocBlock annotations from all methods touched by this change (whether existing or new).

There are definitely others that need cleaning up.

tests/Unit/Css/CssDocumentTest.php
JakeQZ: I replied at length but it's been lost because GitHub's UI/UX is rubbish and I accidentally clicked a link.

JakeQZ: https://github.com/MyIntervals/emogrifier/pull/1142/files#diff-2eda2d27b064a67021c7004fd070acb00ac6bf08695521484af6c087c6eec0baR1425

JakeQZ: In the above link it was possible to split test data up on logical grounds - one set of data syntactically invalid, the other merely with property names/values that don't conform to the spec.

JakeQZ: In this case there doesn't appear to be any logical split between which test data causes an exception and for which the invalid rules are simply discarded.

JakeQZ: We already test in non-debug mode that the invalid rules are discarded.

JakeQZ: In debug mode, to throw an exception in all possible cases of invalid CSS would, I think, be considered a feature request that is never likely to be honoured (very high-hanging fruit, and who is ever going to take the time to pick it?).

JakeQZ: Thus we have a situation where the test can pass if either of the following conditions are met:
- An exception is thrown;
- The invalid CSS rule is discarded.

JakeQZ: Usually in PHPUnit we can use `logicalOr` to test for this.

JakeQZ: But if one of the conditions is an exception being thrown, AFAIAA, this is not possible.

JakeQZ: Hence this construct.

tests/Unit/Css/CssDocumentTest.php
JakeQZ: See above.

tests/Unit/Css/CssDocumentTest.php
JakeQZ: Agree.  Wasn't sure where to put it, but have given this a bit of thought - see my other comment.  Moved.

Now even wondering if constants and properties need necessarily be placed at the top of the class, or could be placed close to where they are relevant...

src/Css/CssDocument.php
JakeQZ: Also added a `use` for the settings class.

JakeQZ (commented):
Have made some changes as suggested, and otherwise made the case for things being as originally proposed.

tests/Unit/Css/CssDocumentTest.php
JakeQZ: I mean below.

JakeQZ: > Please let's use `$this->expectException`, `expectExceptionCode` and `expectExceptionMessage` instead.

Though if there is a `doNotExpectException` method the test could be written without the `try`/`catch` - i.e. we expect an exception but if we reach some later point we'll make an assertion instead, and cancel the expectation of the exception.  Is that possible?

JakeQZ: > Though if there is a `doNotExpectException` method the test could be written without the `try`/`catch` ... Is that possible?

`expectException(null)` would do the trick, except that the parameter is not documented as nullable and a Psalm error is generated.

tests/Unit/Css/CssDocumentTest.php
oliverklee: AFAIK, the recommended order (which also is enforced by some style checker rules IIRC) for "things" in a class is:

1. using traits
2. class constants
3. properties
4. methods

tests/Unit/Css/CssDocumentTest.php
oliverklee: We (as the Emogrifier maintainers) have the secret knowledge of what actually happens ;-) - and I think it's fair that we decide on one possible behavior to be the desired one and test for that. Otherwise, this would make the tests really confusing to read (because the desired behavior then is not clear to the reader).

tests/Unit/Css/CssDocumentTest.php
oliverklee: To test that an exception is _not_ thrown, we can do one of the following things:

a) state that not throwing the exception is the main statement of the test, and use a `@doesNotPerformAssertions` annotation
b) state that the return value (or something similar) is the main statement of the test, and assert for that

tests/Unit/Css/CssDocumentTest.php
JakeQZ: I remember seeing this before.  Perhaps you posted it here.  It is logical.  I was pushing the boundaries by suggesting methods could be interleaved with properties (but why not?).  But on the other side of the coin are the sticklers for insisting that all `static` methods be grouped together, then they be grouped by `public`/`protected`/`private`.  I have seen the unmaintainability of that, e.g. when the signature of a method needs changing.  But what about when a property needs to be changed into a method (or, rarely, vice versa)?

These are slightly rhetorical thoughts and mostly meant as food for thought for your Christmas dinner.  Have a good one 😊 ❤️@🎄.

tests/Unit/Css/CssDocumentTest.php
JakeQZ: What actually happens here is down to the behaviour of `sabberworm/php-css-parser`.  We only know what happens because we have looked into it.  We can influence it and I note you have had a number of PRs accepted.  But at the end of the day it is a 3rd party library.

Ultimately we may be able to dispense with this intermediary class and use the Sabberworm classes directly, but we are a little way off that yet.

Some invalid CSS throws an excpetion (in debug mode).  Some doesn't.  If it throws an excpetion in debug mode, that's a bonus.  If it doesn't, it's no big deal.  But we don't control that.  This relates back to #1135 - to what extent are we testing 3rd party libraries duplicitously?

tests/Unit/Css/CssDocumentTest.php
JakeQZ: But we can't, it seems, have a *'throws exception'* ***OR*** *'satisfies assertion'* test.  Or, at least, not easily.

oliverklee: Then please let's test for one of the two. (We can also add the other case as a test and mark it as `self::markAsSkipped()` (if I remember the method name correctly).

tests/Unit/Css/CssDocumentTest.php
```diff
@@ -488,6 +525,32 @@ public function notRendersInvalidNonConditionalAtRule(string $atRuleCss, string
         self::assertStringNotContainsString($atAndRuleName, $result);
     }
 
+    /**
+     * @test
+     *
+     * @param string $atRuleCss
+     * @param string $cssBefore
```
oliverklee, line 532: We can drop these two as these are redundant with the native type annotations.

tests/Unit/Css/CssDocumentTest.php
```diff
@@ -488,6 +525,32 @@ public function notRendersInvalidNonConditionalAtRule(string $atRuleCss, string
...
+            $subject = self::createDebugSubject($cssBefore . $atRuleCss);
+
+            $result = $subject->renderNonConditionalAtRules();
+
+            \preg_match('/@[\\w\\-]++/', $atRuleCss, $atAndRuleNameMatches);
+            $atAndRuleName = $atAndRuleNameMatches[0];
+            self::assertStringNotContainsString($atAndRuleName, $result);
+        } catch (UnexpectedTokenException $exception) {
+            // Passed test
+            self::expectNotToPerformAssertions();
```
oliverklee, line 550: Ditto about testing only one case (and possibly having the other case in a separate, deactivated, test).

tests/Unit/Css/CssDocumentTest.php
```diff
@@ -22,6 +24,11 @@ final class CssDocumentTest extends TestCase
         . '  font-family: "Foo Sans";' . "\n"
         . '  src: url("/foo-sans.woff2") format("woff2");' . "\n}";
 
+    private static function createDebugSubject(string $css): CssDocument
```
oliverklee, line 27: Please let's not make this static (as there is no need to make it static).

src/Css/CssDocument.php
```diff
@@ -36,11 +37,31 @@ class CssDocument
...
+    }
+
+    /**
+     * Tests if a string of CSS appears to contain an at-rule with nested rules
+     * (`@media`, `@supports`, `@keyframes`, `@document`,
+     * the latter two additionally with vendor prefixes that may commonly be used).
+     *
+     * @see https://github.com/sabberworm/PHP-CSS-Parser/issues/127
+     */
+    private static function hasNestedAtRule(string $css): bool
```
oliverklee, line 62: Please let's not make this method static (as there is no need to make it static).

oliverklee (changes requested):
Sorry for the delay - I think reviewing this got lost for me over the holidays.

src/Css/CssDocument.php
```diff
@@ -36,11 +37,31 @@ class CssDocument
     private $isImportRuleAllowed = true;
 
     /**
-     * @param string $css
+     * @param string $css {@see https://github.com/squizlabs/PHP_CodeSniffer/issues/3521}
```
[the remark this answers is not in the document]
JakeQZ: The `@see` link is merely to a PHPCS bug (https://github.com/squizlabs/PHP_CodeSniffer/issues/3521) which means it's not possible to annotate only some parameters (so if one parameter needs a description, all must be listed in the DocBlock even if it's redundant to do so).  Thus I've simply removed the link.

src/Css/CssDocument.php
JakeQZ: OK.  There are arguments either way but have changed.

tests/Unit/Css/CssDocumentTest.php
JakeQZ: Changed.

tests/Unit/Css/CssDocumentTest.php
JakeQZ: Removed.

tests/Unit/Css/CssDocumentTest.php
JakeQZ: OK.  I've change this (and the other similar test) to expect an exception, but if one isn't thrown mark the test as skipped - since in strict (debug) mode the CSS parser only throws an exception with some of the datasets, and we are already testing that the invalid rules from these datasets are dropped in lenient (non-debug) mode.

tests/Unit/Css/CssDocumentTest.php
JakeQZ: Ditto.

JakeQZ (commented):
I've hopefully resolved all the comments from the latest review - see individual replies.

src/Css/CssDocument.php
```diff
@@ -37,10 +38,30 @@ class CssDocument
...
-    public function __construct(string $css)
+    public function __construct(string $css, bool $debug)
     {
-        $this->sabberwormCssDocument = (new CssParser($css))->parse();
+        // CSS Parser currently throws exception with nested at-rules (like `@media`) in strict parsing mode
+        $parserSettings = ParserSettings::create()->withLenientParsing(!$debug || $this->hasNestedAtRule($css));
+
+        // CSS Parser currently throws exception with non-empty whitespace-only CSS in strict parsing mode, so `trim()`
+        // @see https://github.com/sabberworm/PHP-CSS-Parser/issues/349
+        $this->sabberwormCssDocument = (new CssParser(\trim($css), $parserSettings))->parse();
```
oliverklee, line 52: Why the added `trim`?

tests/Unit/Css/CssDocumentTest.php
```diff
@@ -488,6 +523,25 @@ public function notRendersInvalidNonConditionalAtRule(string $atRuleCss, string
...
+     */
+    public function throwsExceptionForInvalidNonConditionalAtRuleInDebugMode(
+        string $atRuleCss,
+        string $cssBefore = ''
+    ): void {
+        $this->expectException(UnexpectedTokenException::class);
+
+        $this->createDebugSubject($cssBefore . $atRuleCss);
+
+        self::markTestSkipped(
```
oliverklee, line 539: Please let's have the `markTestSkipped` calls at the start of the test so no other code gets run. (This also helps the reader.)

oliverklee (changes requested):

src/Css/CssDocument.php
JakeQZ: For the reason in the comment above - see https://github.com/sabberworm/PHP-CSS-Parser/issues/349.

tests/Unit/Css/CssDocumentTest.php
JakeQZ: That would skip the test for all datasets, including the ones that currently pass.

The test data consists of CSS that contains some invalid declarations.  We test to confirm that in non-debug mode the invalid CSS is dropped.  In debug ("strict") mode, the CSS parser throws exceptions for some but not all invalid CSS constructs.  So we cannot assert that an exception is thrown for all the CSS strings in the data.

But if we don't assert it for any, we may as well not bother with the test at all, or have a dummy test that does not use the data provider but calls `markTestSkipped()` so we just see one skipped test in the results (not one for each dataset).

An alternative would be to split the datasets (data providers) up so that one provides datasets for which the CSS parser currently throws an exception, and one provides those for which it doesn't.  Though there doesn't appear to be any particular distinctive feature of those for which an exception is thrown - it's just cases where the strictness isn't quite as strict as the spec dictates.

WDYT?

oliverklee: > An alternative would be to split the datasets (data providers) up so that one provides datasets for which the CSS parser currently throws an exception, and one provides those for which it doesn't.

That's the approach I'd prefer.

JakeQZ: OK.  I've split up the data accordingly, only passing the CSS for which the CSS parser throws an exception to the tests for that, with a comment in the DocBlock indicating that the other CSS is not currently tested.

oliverklee (approved):
Neat!

PR #1144, opened 2022-01-07
[BUGFIX] Stop using a literal string as the rule value in a test

oliverklee:
Our previous usage violated the class contract.

### DIFF
--- a/tests/Unit/Css/StyleRuleTest.php
+++ b/tests/Unit/Css/StyleRuleTest.php
@@ -8,6 +8,7 @@ use Pelago\Emogrifier\Css\StyleRule;
 use PHPUnit\Framework\TestCase;
 use Sabberworm\CSS\Rule\Rule;
 use Sabberworm\CSS\RuleSet\DeclarationBlock;
+use Sabberworm\CSS\Value\RuleValueList;
 
 /**
  * @covers \Pelago\Emogrifier\Css\StyleRule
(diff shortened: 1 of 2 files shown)


tests/Unit/Css/StyleRuleTest.php
```diff
@@ -8,6 +8,8 @@
 use PHPUnit\Framework\TestCase;
 use Sabberworm\CSS\Rule\Rule;
 use Sabberworm\CSS\RuleSet\DeclarationBlock;
+use Sabberworm\CSS\Value\CSSString;
```
JakeQZ: This class isn't referenced by name.

tests/Unit/Css/StyleRuleTest.php
```diff
@@ -93,7 +94,9 @@ public function getDeclarationAsTextReturnsConcatenatedDeclarationsFromRules(
         $declarationBlock = new DeclarationBlock();
         foreach ($declarations as $declaration) {
             $rule = new Rule($declaration['property']);
-            $rule->setValue($declaration['value']);
```
JakeQZ, line 96: I guess this worked because the test only ends up doing an implicit `__tostring()` on the value in `Rule::render()`, so we got away with providing a string instead of a `Value`.

JakeQZ (commented):
I think you've just dealt with my comment which is now outdated.

JakeQZ (approved):

oliverklee:
> I think you've just dealt with my comment which is now outdated.

Coincidentally, I've just researched courses on the topic of time travel on Coursera. :-)

PR #1147, opened 2022-05-21
[TASK] Update composer-normalize

oliverklee:


JakeQZ (approved):
Assuming the binary file is correct.

PR #1148, opened 2022-05-21
[TASK] Update Psalm

oliverklee:


psalm.xml
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0"?>
 <psalm
-  totallyTyped="true"
```
JakeQZ, line 3: To be equivalent, we should set [`errorLevel`](https://psalm.dev/docs/running_psalm/configuration/#errorlevel) to `1`.

JakeQZ (changes requested):
We need `errorLevel="1"` in the config file to replace `totallyTyped="true"`.

psalm.xml
oliverklee: Done.

JakeQZ (approved):

PR #1152, opened 2022-05-24
[TASK] Include Dependabot configuration in the release checklist

oliverklee:
Also use real numbers for the affected enumeration in the README
as PhpStorm now does this automatically, and working against PhpStorm
would be too much of a hassle.


JakeQZ (approved):
> Also use real numbers for the affected enumeration in the README
as PhpStorm now does this automatically, and working against PhpStorm
would be too much of a hassle.

And it makes no difference as far as Markdown is concerned AFAIK :)

PR #1154, opened 2022-06-10
Run CI with PHP 8.2

TimWolla:
Tests appear to pass just fine locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,6 +163,10 @@ jobs:
           tools: composer:v2
           coverage: none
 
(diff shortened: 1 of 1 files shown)


oliverklee:
Hi @TimWolla, thanks a lot for the PR!

Do you know for which packages we currently need the faked PHP 8.1, i.e., which things currently would break with 8.2?

oliverklee:
Thanks for the explanation, @TimWolla!

oliverklee (approved):
Looks good to me. Thanks for contributing!

PR #1156, opened 2022-06-10
[TASK] Treat PHP versions as strings in ci.yml

TimWolla:
Previously the dot-zero versions would show as integers (e.g. 8 instead of 8.0)
within the titles of the jobs, which is not quite right.


oliverklee (approved):

TimWolla:
You'll likely need to adjust the protected branch config to merge this, because the job is now renamed.

PR #1157, opened 2022-06-25
[FEATURE] Configure Dependabot to update the Composer dev dependencies

oliverklee:


JakeQZ (commented):
Should this also not include `milestone: 12` for PRs it creates?

oliverklee:
It should indeed. :-) Added and repushed.

JakeQZ (approved):
👍

JakeQZ:
Wonder if it's possible to configure the default commit title (and perhaps even commit message).

We have prefixes "[TASK]", "[BUGFIX]", etc.  Perhaps a prefix for these such commits might be "[BOT]".

oliverklee:
Dependabot indeed [allows configuring a prefix](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message). However, the if a prefix is configured, Dependabot will automatically add a colon and a space after the prefix.

For example, if we configured `[BOT]` as the prefix, the resulting commit messages would look like this: `[BOT]: Update foo`

Would this still be an improvement? What do you think, @JakeQZ ?

oliverklee:
Or we could change our prefix scheme for commit messages from this:

`[BUGFIX] Fix space-time continuum`

… to this:

`BUGFIX: Fix space-time continuum`

… which then would work with the Dependabot auto-prefixing.

JakeQZ:
I think we should stick with our [square bracket] prefix which is quite standard.

We can edit the commit title when approving the bot's changes to remove the colon, and it's no big deal if we forget to do so.

The bot's commit message is also wrong becuase it says to "permit" the latest version when it should say "enforce".

JakeQZ:
Also, what the bot does is likely to change, as it is new, so we should mostly ignore its current style.

PR #1162, opened 2022-07-17
[TASK] Drop support for Symfony 5.3

oliverklee:
Symfony 5.3 has reached its end of life.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Support for PHP 7.3 will be removed in Emogrifier 8.0.
 
 ### Removed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Support for PHP 7.3 will be removed in Emogrifier 8.0.
 
 ### Removed
-- Drop support for Symfony 3.x (#1120)
+- Drop support for Symfony 3.x and 5.3 (#1120, #1162)
```
JakeQZ, line 20: I think it's right to mention this in the changelog even when dropping support for a minor version, in case someone is relying on that specific version.

JakeQZ (approved):

PR #1165, opened 2022-08-04
Mark `phpunit.xml` as `export-ignore` in `.gitattributes`

TimWolla:


oliverklee (approved):
Looks good. Thank you for this contribution, @TimWolla! :heart:

PR #1166, opened 2022-08-08
[TASK] Use PHP 8.2 in CI without faking the version for Composer

oliverklee:
All necessary packages now are available for PHP 8.2, and we
do not need to fake PHP 8.2 as 8.1 in the Composer configuration
in order to get the packages installed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,10 +163,6 @@ jobs:
           tools: composer:v2
           coverage: none
 
-      - name: Fake PHP version for 8.2.
-        run: composer config platform.php 8.1.99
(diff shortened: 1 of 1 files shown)


oliverklee:
It looks like we still need to wait for https://github.com/phpspec/prophecy/issues/556 in Prophecy.

TimWolla:
This should work now, that PHPUnit 9.5.23 is required: #1168

oliverklee:
Indeed. Thanks for keeping an eye on this, @TimWolla! :pray:

JakeQZ (approved):
LGTM.

Issue #1174, opened 2022-09-26
Remove Dependabot


JakeQZ:
Dependabot appears to be a waste of our time.

All it seems to do is up the minimum requried version of PHPUnit whenever there is a new point release of that.  In the PR it creates it says "to permit the latest version".  This is completely wrong.  The latest point release version is always permitted because we use `^` not `=` for the version number in `composer.json`.

However, we run our CI and weekly tests with both `DEPENDENCIES=lowest` and `DEPENDENCIES=latest` (or suchlike).  So they will always be run with the latest version of PHPUnit, as well as an older version that a developer may have kicking around if they have not run `composer update` recently.

I see no need to constantly up the minimum version requirement for PHPUnit if the bugfixes in the point release are not relevant to us.

Whenever there's a new version of PHPUnit, we are asked to approve Dependabot's changes to enforce that that version is now used.  This is a drain on our time, that could be better spent elsewhere.

Also, having a wider allowed range of version numbers for dependencies surely gives us better testing coverage.  Thus, we should not be enforcing specific minimum versions unless we really need them (e.g. for bugfixes we actually need).

I suggest we remove it.

oliverklee:
@JakeQZ You sound really frustrated.

I find it important that we keep our dependencies mostly up to date. Would you be willing to regularly check them (e.g., monthly) and provide a PR for updating them?

JakeQZ:
Monthly wiuld be far too often.

Software updates tend only ever to break things, whilst only fixing the most obscure security vulnerabilities that haven't even been found in the wild.

If the Dbot's commit message was customizable, that would be a start.  The author clearly does not speak English as a first language.

oliverklee:
We probably also can provide a PR to improve the commit messages: https://github.com/dependabot/dependabot-core

JakeQZ:
> We probably also can provide a PR to improve the commit messages: https://github.com/dependabot/dependabot-core

Ideally it would be per-project customizable.  There's no point changing it otherwise.

oliverklee:
> Monthly wiuld be far too often.

What would be a frequency for manually updating the dependencies in which you would be willing to do this?

oliverklee:
And would you be willing to provide the PR to remove Dependabot?

JakeQZ:
> @JakeQZ Oh dear. You sound really frustrated.

[...]

PR #1175, opened 2022-09-27
[TASK] Stop version-controlling the development tool PHARs

### LINKED ISSUE #984: Stop version-controlling the tool PHARs?
https://github.com/MyIntervals/emogrifier/pull/968#issuecomment-815341117:

> Feedback from https://github.com/phar-io/phive/issues/296 suggests we have an "interesting use case".  However feedback from https://github.com/phar-io/phive/issues/295 indicates PHIVE currently does not support installing a version necessarily compatible with the calling environment.

https://github.com/MyIntervals/emogrifier/pull/968#issuecomment-815605094:
> Let's see what the rest of the PHP developer crowd thinks. :-)
>
> I have created a Twitter poll for this: https://twitter.com/oliklee/status/1380082284200595457

oliverklee:
Fixes #984


.phive/phars.xml
```diff
@@ -1,9 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phive xmlns="https://phar.io/phive">
-  <phar name="composer-normalize" version="^2.28.3" installed="2.28.3" location="./.phive/composer-normalize" copy="true"/>
-  <phar name="phar-io/phive" version="^0.14.5" installed="0.14.5" location="./.phive/phive" copy="true"/>
-  <phar name="php-cs-fixer" version="^3.11.0" installed="3.11.0" location="./.phive/php-cs-fixer" copy="true"/>
-  <phar name="phpcs" version="^3.7.1" installed="3.7.1" location="./.phive/phpcs" copy="true"/>
-  <phar name="phpmd" version="^2.13.0" installed="2.13.0" location="./.phive/phpmd" copy="true"/>
-  <phar name="psalm" version="^4.27.0" installed="4.27.0" location="./.phive/psalm" copy="true"/>
```
JakeQZ, line 8: This still seems to be saying that the `phar`s are installed.  But if they've been taken out of version control, they may not be (and probably won't be on the CI build server).  What is the suggested `phars.xml` content before PHIVE is run to install the packages?

Also, I don't know what `copy=[true|false]` means.  I'm sure I could look it up, but if you know and have a quick explanation you could provide, that would save me time hunting around.

Otherwise looks good.  I saw your Twitter poll responses with over 60% in agreement of this approach (from a choice of three).  Well done with that.  (Social media does have a use, after all.)

I assume when the CI build runs, it will install the latest and greatest that satisfies the `version` requirement.  And we won't have a `DEPENDENCIES=lowest` option here (it is not necessary, though useful from a testing POV - see https://github.com/MyIntervals/emogrifier/issues/1174#issuecomment-1262995694).

JakeQZ: > This ...

Not sure why this is appearing twice.

JakeQZ: > > This ...
> 
> Not sure why this is appearing twice.

Maybe just my end.  I've refreshed and it now only appears once.  But is another example of bug-riddled software from Microsoft.

oliverklee: The PHIVE configuration file works a bit like a combination of `composer.json` and `composer.lock`: The `version` requirement determines which versions are allowed when `phive update` is called, while the `installed` version determines the exact version that gets installed when `phive install` is called (which is what we're doing in the CI jobs).

When we're using `copy=true`, the PHAR is installed in the project folder. If we're using `copy=false´, the PHAR is installed on a central directory on the machine and them symlinked to the project directory.

JakeQZ: > The PHIVE configuration file works a bit like a combination of `composer.json` and `composer.lock`: The `version` requirement determines which versions are allowed when `phive update` is called, while the `installed` version determines the exact version that gets installed when `phive install` is called (which is what we're doing in the CI jobs).

Seems a bit weird to me.

But I have always found the way PHIVE works to be quite odd.

I trust your explanation.
 
> When we're using `copy=true`, the PHAR is installed in the project folder. If we're using `copy=false´, the PHAR is installed on a central directory on the machine and them symlinked to the project directory.

Thanks.  I did already discover that previously, but had completely forgotton.

JakeQZ (approved):
Thanks for the explanations.  I think this is good to go.

PR #1179, opened 2022-10-04
[TASK] Remove the Dependabot configuration for Composer updates

### LINKED ISSUE #1174: Remove Dependabot

oliverklee:
We'll update our dependencies manually if and when the need arises
(as the Composer updates were the majority of the PRs by Dependabot).

Dependabot will still update the GitHub actions.

Dependabot will still create PRs for security-relevant updates, though.

Also drop the target milestone from the configuration in order to
reduce the work needed after we have released a new major version.

Fixes #1174

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,13 +6,3 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
-    milestone: 12
-
-  - package-ecosystem: "composer"
-    directory: "/"
-    schedule:
-      interval: "daily"
-    allow:
-      - dependency-type: "development"
-    versioning-strategy: "increase"
-    milestone: 12


JakeQZ (approved):
Looks perfect to me.

> Dependabot will still create PRs for security-relevant updates, though.

Yes, I agree,  Whilst we should not be dictating what versions of dependencies people use, we should at least encourage against continued use of sotware with secvuls.  Also, on the CI build server, we cannot allow software with known security flaws (though GitHub will probably allow it), as that imposes a risk to other users.

PR #1182, opened 2022-10-06
[TASK] Convert deprecations to exceptions in the tests

### LINKED ISSUE #1181: set convertDeprecationsToExceptions="true" in phpunit.xml?
https://github.com/MyIntervals/emogrifier/pull/1180#issuecomment-1268076336

I think this is approprate for the CI build.

But for developers, it could be a PITA.

I think we need to distinguish between developer and release.

And not have things hampering development.

I need to occasionally run a build locally and test against some other software, without having the build failing due to e.g. some `error_log` calls (which will be rejected as settings currently are) to track down a problem.

We need both worlds.  Tight and strict as possible for testing for release.  But as loose as possible for use in development.  And maybe some way to move the slider between them.

oliverklee:
This will allow us to find code which we will need to update to changes in PHP or in our dependencies.

Also order the configuration options in the PHPUnit configuration file.

Fixes #1181

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,12 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.5/phpunit.xsd"
-         bootstrap="vendor/autoload.php"
-         colors="true"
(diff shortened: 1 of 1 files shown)


oliverklee:
It looks like we need to at least bump our Symfony dependencies.

JakeQZ (approved):
Nice job.  (Not much in this PR, but the other stuff that was needed to allow it.)

PR #1183, opened 2022-10-06
[TASK] Switch more CI jobs to PHP 8.1

oliverklee:


JakeQZ (approved):
We seem to have missed this for PHP 8.0, though that doesn't matter now.

This is the sort of thing a tool like Dependabot could actually help with.  I doubt that will ever happen, though.

PR #1184, opened 2022-10-06
[BUGFIX] Remove speculative parent constructor calls in test code

oliverklee:
This fixes a deprecation warning in PHP 8.2.

### DIFF
--- a/tests/Support/Constraint/CssConstraint.php
+++ b/tests/Support/Constraint/CssConstraint.php
@@ -225,23 +225,4 @@ abstract class CssConstraint extends Constraint
 
         return $matcher;
     }
-
-    /**
-     * Invokes the parent class's constructor if there is one.  Since PHPUnit 8.x, `Constraint` (the parent class) does
-     * not have a defined constructor.  But...
-     *
-     * In PHP, a class that may be extended should have a defined constructor, even if it does nothing.
-     *
-     * That way, a child class that needs a constructor can explicitly call its parent's constructor.  If it could not,
-     * and then a newer version of the parent class needed a constructor, the newly added constructor would not be
-     * invoked implicitly during instantiation of the child class.  In that eventuality, the child class would also need
-     * updating, so there would be a compatibility issue, and potential for uncaught bugs if the child class's
-     * constructor was not updated.
-     */
-    protected function __construct()
-    {
-        if (\is_callable('parent::__construct')) {
-            parent::__construct();
-        }
-    }
 }
--- a/tests/Support/Constraint/IsEquivalentCss.php
+++ b/tests/Support/Constraint/IsEquivalentCss.php
@@ -27,8 +27,6 @@ final class IsEquivalentCss extends CssConstraint
      */
     public function __construct(string $css)
     {
-        parent::__construct();
-
         $this->css = $css;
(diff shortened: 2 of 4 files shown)


JakeQZ (commented):
What is the deprecation warning saying?  Is `is_callable` deprecated, or is it the format of the string being passed as a parameter, or something else?  Does it occur at parse time or run time?

I’d like to retain the code conditionally calling the parent constructor depending on its existence, if at all possible, and it can be somehow fixed to avoid the deprecation warning, rather than remove it.

PHP has a language flaw that other OO languages do not.  Parent class constructors are not implicitly called.  Thus, if a 3rd party library has a constructor added to a (non-final) class in a new release, it will not be called by subclasses, meaning important initialization will be skipped, and the software will have bugs introduced, in scenarios that may not be caught by existing unit tests.

I have been bitten by this a couple of times.  The bugs introduced have been obscure and hard to track down.

The problem is avoided by always defining a constructor (in a non-final class), even if it is empty.  That way, users of your class can explicitly call it from their subclasses, and, if, later, you need it to do something, you won’t be making a breaking change (assuming your users implemented their constructor to call yours).

For those reasons, I think it was wrong of the developers of PHPUnit to remove the constructor for `Constraint`.  If in a future release, they need to add one back, people’s tests will not work properly, possibly leading to more bugs going undetected.

This rationale is even mentioned in the DocBlock for the constructor.

However, I've just seen this at https://www.php.net/manual/en/function.is-callable.php :

> `is_callable()` reports constructors as not being callable.

So maybe the code as it was was not functioning as intended.

oliverklee:
This is the error in PHP 8.2 with the switch “convert deprecations to exception” enabled:

> 2) Pelago\Emogrifier\Tests\Unit\Css\CssDocumentTest::parsesSelector with data set "type" ('p')
[133](https://github.com/MyIntervals/emogrifier/actions/runs/3196361353/jobs/5218722180#step:8:134)
Use of "parent" in callables is deprecated

https://github.com/MyIntervals/emogrifier/actions/runs/3196361353/jobs/5218722180

oliverklee:
Maybe we can use a `method_exists` call instead?

TimWolla:
Relevant RFC with background information: https://wiki.php.net/rfc/deprecate_partially_supported_callables

`[parent::class, '__construct']` might work as expected.

JakeQZ:
> Maybe we can use a `method_exists` call instead?

Possibly, if that works.  Or, failing that, reflection - since it's in testing code rather than the actual product, using reflection would be fine I think.

I'd like it be verified that what we do go with actually works as intended.  This would involve knocking up a couple of classes and seeing if the test for existence of the parent constructor actually works in the child constructor, and it's called if it exists.

We don't need this to be added to our tests - it's just to satisfy ourselves of PHP's behaviour, which is not clear from the documentation - it's possible that what I assumed was currently doing the trick was actually not, because I didn't realize `is_callable()` reports constructors as not being callable (though don't know if that's the case when called from within a subclass's constructor, or only when called from outside the class).

JakeQZ:
> `[parent::class, '__construct']` might work as expected.

May be worth a try.  If I find time later this evening I'll see what might work.  However, I don't have PHP 8.2 so will be unable to know if deprecation warnings are triggered.

oliverklee:
Superseded by #1186.

PR #1187, opened 2022-10-07
[BUGFIX] Bump the minimum Symfony 4.4. version

oliverklee:
For PHP 8.1 and 8.2, we need a bug fix that avoids a deprecation warning. (Symfony 5.4.0, 6.0.0 and 6.1.0 already have this fix.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Drop support for PHP 7.2 (#1111)
 
 ### Fixed
-- Bump the minimum Symfony 4.4. version (#1187)
+- Bump the minimum Symfony 4.4 version to avoid PHP deprecation warnings (#1187)
 
 ## 6.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Drop support for PHP 7.2 (#1111)
 
 ### Fixed
+- Bump the minimum Symfony 4.4. version (#1187)
 
 ## 6.0.0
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Drop support for PHP 7.2 (#1111)
 
 ### Fixed
+- Bump the minimum Symfony 4.4. version (#1187)
```
JakeQZ, line 25: It took me a bit of time to understand this.  I think "4.4.x" is much clearer than "4.4." which looks like the end of a sentence.  Also we should give the reason (such as "to avoid a PHP deprecation warning").

composer.json
```diff
@@ -41,7 +41,7 @@
         "ext-dom": "*",
         "ext-libxml": "*",
         "sabberworm/php-css-parser": "^8.4.0",
-        "symfony/css-selector": "^4.4 || ^5.4 || ^6.0"
+        "symfony/css-selector": "^4.4.23 || ^5.4.0 || ^6.0.0"
```
JakeQZ, line 44: I like adding the `0` point release number for clarity.

JakeQZ (changes requested):
Niggly, but I think the changelog entry needs improvement and clarification, since this will go out on the release note.

CHANGELOG.md
oliverklee: Ah, no, the additional period was just a typo. :-)

I'll update the changelog entry in a second.

JakeQZ (approved):
Looks fine now to me.

CHANGELOG.md
JakeQZ: I've fixed it in the commit title too.

PR #1192, opened 2022-10-12
[TASK] Replace deprecated rule in our PHP-CS-Fixer ruleset

### LINKED ISSUE #1190: Stop using the `no_trailing_comma_in_singleline_array` PHP-CS-Fixer rule
```
Using cache file ".php-cs-fixer.cache".
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS                                                                                                                      35 / 35 (100%)
Legend: .-no changes, F-fixed, S-skipped (cached or empty file), I-invalid file syntax (file ignored), E-error

Checked all files in 0.001 seconds, 14.000 MB memory used

Detected deprecations in use:
- Rule "no_trailing_comma_in_singleline_array" is deprecated. Use "no_trailing_comma_in_singleline" instead.
```

oliverklee:
Fixes #1190

### DIFF
--- a/config/php-cs-fixer.php
+++ b/config/php-cs-fixer.php
@@ -28,7 +28,7 @@ return (new \PhpCsFixer\Config())
 
             // array notation
             'array_syntax' => ['syntax' => 'short'],
-            'no_trailing_comma_in_singleline_array' => true,
+            'no_trailing_comma_in_singleline' => true,
             'no_whitespace_before_comma_in_array' => true,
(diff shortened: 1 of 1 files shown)


JakeQZ (approved):
I'm still wondering what else apart from an array can have a trailing comma - I know function calls can't, because I've sometimes converted an array to function call parameters, left the trailing comma in, and got the expected parse error.

Obviously we need to change this to match the new name.

PR #1193, opened 2022-10-12
[TASK] Switch PHP-CS-Fixer from PSR12 to its sucessor PER

### LINKED ISSUE #1189: Switch PHP-CS-Fixer from PSR12 to its sucessor PER
https://www.php-fig.org/per/coding-style/

oliverklee:
https://www.php-fig.org/per/coding-style/

Fixes #1189


JakeQZ (approved):
Interesting - and good - that we already appear to match this standard.  We are ahead of the game :))

PR #1196, opened 2022-11-01
[TASK] Prepare release of version 7.0.0

oliverklee:
We should have version 7.0.0 with PHP 8.2 support out of the door before the final PHP 8.2 will get released later this month.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+
+### Changed
+
+### Deprecated
+
+### Removed
+
+### Fixed
+
+## 7.0.0
+
+### Added
 - Add support for PHP 8.2 (#1155)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,8 +5,21 @@ This project adheres to [Semantic Versioning](https://semver.org/).
...
+
+### Removed
+
+### Fixed
+
+## 7.0.0
+
 ### Added
 - Add support for PHP 8.2 (#1155)
+- Add official support for Symfony 6
```
JakeQZ, line 22: This was added in #1065 which was included in 6.0.0 so I don't think this changelog entry belongs here.

JakeQZ (changes requested):
I think the changelog entry about Symfony 6 doesn't belong.

CHANGELOG.md
oliverklee: Oh, right, I think we missed that. Going to change that back in a second.

JakeQZ (approved):
Looks fine now.

PR #1199, opened 2022-11-30
[TASK] Upgrade to Psalm 5

oliverklee:
Also update the baseline with the newly-found warnings (which we can tackle in future changes).


JakeQZ (approved):
I've added #1200 so we can track resolution of the new issues found.

oliverklee:
Thanks!

PR #1208, opened 2023-01-16
[FEATURE] Re-establish support for PHP 7.2

### LINKED ISSUE #1207: Re-establish support for PHP 7.2
Revert #1111.

oliverklee:
This reverts commit 9cd1e6073290ef4328065d1709a73067465b7fee.

Fixes #1207

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
+          - '7.2'
           - '7.3'
           - '7.4'
           - '8.0'
@@ -172,6 +173,7 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
+          - '7.2'
           - '7.3'
           - '7.4'
           - '8.0'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 ### Changed
+- Re-establish support for PHP 7.2 (#1208)
 
 ### Deprecated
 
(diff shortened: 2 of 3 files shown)


JakeQZ:
There may have been some breaking changes between PHPUnit 8.x and 9.x.  IIRC, some of the assert methods were renamed, with the old names deprecated.  It may be that we do need to drop support for PHP 7.2, if it's not possible for our test code to be compatible with all the versions of PHPUnit that would be needed to support such a wide range range of PHP versions.

The principle was not to drop support for older PHP versions if there's no reason to do so.  But if there is a reason, so be it.

mvorisek:
Yes, please do not drop PHP support too early. This library prevented us to upgrade PHP for Prestashop which uses v5.x which is limited to PHP 8.0 for no big reason.

JakeQZ:
> This library prevented us to upgrade PHP for Prestashop which uses v5.x which is limited to PHP 8.0 for no big reason.

It took me a few minutes to work out your point.  I have now got it.

I also disagree with setting a maximum version number for our software.  If we didn't, we wouldn't have to constantly update it for every new PHP release.

@oliverklee's argument is that otherwise we would be releasing software that has not been tested against a future PHP version, but nontheless claims to work with it.

Prestashop could up the version of Emogrifier to 6.0.  Or use `Composer` to select and install a suitable version.  I think the issue you have is with Prestashop.

oliverklee:
We also could provide maintenance versions (with bug fixes and support for higher PHP versions) of e.g., the previous major release. However, someone would need to put the time into that to maintain these branches (mostly backport PRs, and create PRs for adding support and CI for higher PHP versions). @JakeQZ @mvorisek Would you two be willing to do this?

JakeQZ:
> @JakeQZ @mvorisek Would you two be willing to do this?

(I replied the GitHub notification email with a comment, which I thought would get posted here.  I guess that feature of GitHub doesn't work, or maybe it wasn't intended to work like that.)

Basically, no,  (I don't have that extra time.)

But I think addressing #1210 should solve most of these types of issues for the future, in an automated way, so we wouldn't have to do anything more than that (moving forwards).

@mvorisek, thanks for sharing your experience; useful feedback and good to know.  Feel free to further comment on #1210 there.

Re this PR, I think due to the breaking changes between various major versions of PHPUnit (and maybe other compatibility issues), re-supporting PHP 7.2 would be too much effort to be viable given that the already-small benefit will fade over time.

So I'm happy to close this (as `willnotimplement` or suchlike).  @oliverklee, thanks for giving it a try.

PR #1209, opened 2023-01-16
[TASK] Downgrade from PHPUnit 9.x to 8.x

oliverklee:
This allows us to re-introduce support for PHP 7.2.


JakeQZ:
#1208 is closed, so this can be too.  See the comment there for the rationale/reasoning.

PR #1211, opened 2023-01-23
[TASK] Update to Psalm 5.6

oliverklee:


psalm.baseline.xml
```diff
@@ -1,80 +1,20 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="dev-master@">
+<files psalm-version="5.6.0@e784128902dfe01d489c4123d69918a9f3c1eac5">
```
JakeQZ, line 2: We've seen this weird change before, and found it to be a non-issue, IIRC.

psalm.baseline.xml
```diff
@@ -1,80 +1,20 @@
...
+            $cssRules['inlinable'],
+            /**
+             * @param array{selector: string, line: int} $first
+             * @param array{selector: string, line: int} $second
+             */
+            function (array $first, array $second): int {
+                return $this-&gt;sortBySelectorPrecedence($first, $second);
+            }
+        )</code>
+    </InvalidArgument>
```
JakeQZ, line 15: We should investigate this as a separate issue.

psalm.xml
```diff
@@ -2,6 +2,7 @@
 <psalm
   errorLevel="1"
   strictBinaryOperands="true"
+  findUnusedBaselineEntry="true"
```
JakeQZ, line 5: Looks like a useful addition, though on just reading the [one-line description](https://psalm.dev/docs/running_psalm/configuration/#findunusedbaselineentry), I don't fully understand what it actually does.

JakeQZ (approved):
Interesting that a lot of the warnings have disappeared.  I guess the new version is smarter, and smart enough to determine that these are non-issues.

psalm.baseline.xml
JakeQZ: #1212

psalm.xml
oliverklee: This switch controls whether Psalm will complain about baseline entries listing a problem that does not exist (anymore).

JakeQZ:
Oops; forgot to click squash and merge after approving.  Done now.

psalm.xml
JakeQZ: Ah, so some of the warnings that were removed in this PR should perhaps have been removed previously.

oliverklee: Yes, possibly (though I tried to remember to update the baseline file when updating Psalm).

PR #1218, opened 2023-08-01
Add support for PHP 8.3

TimWolla:


oliverklee (commented):
Thanks for the PR, @TimWolla! :heart:

We should change this to “Add support for PHP 8.3”, allow PHP 8.3 in `composer.json` and add a corresponding feature changelog entry.

Also, is there any particular package that is not compatible with 8.3 yet and that forces us to ignore the platform requirements?

TimWolla:
> We should change this to “Add support for PHP 8.3”, allow PHP 8.3 in composer.json and add a corresponding feature changelog entry.

Okay, I've rewritten the commit history to do this.

> Also, is there any particular package that is not compatible with 8.3 yet and that forces us to ignore the platform requirements?

When testing, CI for PHP 8.3 failed due to Emogrifier's own composer.json.

oliverklee (approved):
Awesome, thanks!

PR #1221, opened 2023-09-07
[BUGFIX] Accurate return type for AbstractHtmlProcessorTest data providers

JakeQZ:
`TRegx\DataProvider\DataProviders::cross` now expects `array<int|string, array<int, mixed>>`, rather than whatever it was previously loosely specified to.

(This does raise the question of whether tightening up the code annotation for Psalm could be considered a breaking change.  Though it only breaks the CI, not the software itself, so I guess not.)

### DIFF
--- a/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
+++ b/tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
@@ -834,7 +834,7 @@ final class AbstractHtmlProcessorTest extends TestCase
     }
 
     /**
-     * @return string[][]
+     * @return array<array-key, array<int, string>>
      */
     public function provideContentTypeMetaTag(): array
     {
@@ -860,7 +860,7 @@ final class AbstractHtmlProcessorTest extends TestCase
     }
 
     /**
-     * @return string[][]
+     * @return array<array-key, array<int, string>>
      */
     public function provideHtmlAroundContentType(): array
     {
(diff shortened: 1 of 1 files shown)


JakeQZ:
Missed that there were two similar changes needed, not just one.

I'm getting lots of errors (i.e. noise) from Psalm, all of which seem to make sense, and none of which are in the `baseline`.  Don't know why you're not.  I'm running PHP 7.4 on Windows, whilst you're running PHP 8.1 on Linux.  That should not matter; Psalm detects from `composer.json` that we're targeting PHP 7.3 as a minimum.  I've had this issue before, and it seemed to be down to PHIVE.  If I got Composer to install Psalm instead of PHIVE, I got the same results as you.  IIRC, I had a hacked version of `composer.json` to do that.

But, for example, it's pointing out that the return value from `preg_replace` is not always a string.  In the error case it is an integer.  We are not handling that situation.  So I think I need to use the extra errors I'm getting to tighten things up.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -860,7 +860,7 @@ public function provideContentTypeMetaTag(): array
     }
 
     /**
-     * @return string[][]
+     * @return array<array-key, array<int, string>>
```
oliverklee, line 863: For both changes, we can be even more specific and set `string` (instead of `array-key`) for the keys of the outer array.

oliverklee (changes requested):

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
JakeQZ: Actually we can be even more specific, since the inner arrays always have the same number of elements.

oliverklee (approved):
Neat!

PR #1224, opened 2023-09-08
[TASK] Annotate regex patterns as non-empty-string

JakeQZ:
A mere `string` type annotation will generate errors with Psalm 5.15 (the latest version at time of writing).


oliverklee (approved):
I like it!

PR #1226, opened 2023-09-10
[BUGFIX] Skip any style elements with null value, but still remove them

JakeQZ:
Psalm indicates that `DOMNode::nodeValue` could be `null`, perhaps for entirely empty `<style>` elements.

If such are encountered during pre-processing, still remove them from the DOM, but don't append their contents to the CSS to be inlined with extra line-breaks.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -492,7 +492,7 @@ class CssInliner extends AbstractHtmlProcessor
 
         $css = '';
         foreach ($styleNodes as $styleNode) {
-            if ($styleNode->nodeValue !== null) {
+            if (\is_string($styleNode->nodeValue)) {
                 $css .= "\n\n" . $styleNode->nodeValue;
             }
             $parentNode = $styleNode->parentNode;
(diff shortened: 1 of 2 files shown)


src/CssInliner.php
```diff
@@ -492,7 +492,9 @@ private function getCssFromAllStyleNodes(): string
 
         $css = '';
         foreach ($styleNodes as $styleNode) {
-            $css .= "\n\n" . $styleNode->nodeValue;
+            if ($styleNode->nodeValue !== null) {
```
oliverklee, line 495: Let's make this more clear to the reader which type is expected (so the reader doesn't need to look up the non-null type of `DOMNode::nodeValue`):

```suggestion
            if (\is_string($styleNode->nodeValue)) {
```

oliverklee (changes requested):

oliverklee (approved):
LGTM.

oliverklee:
@JakeQZ Could you please rebase, resolve the merge conflicts, and repush? Thanks!

oliverklee (approved):

PR #1227, opened 2023-09-10
[TASK] Update the CI tools

oliverklee:
Also update the configuration files.

### DIFF
--- a/config/php-cs-fixer.php
+++ b/config/php-cs-fixer.php
@@ -72,7 +72,7 @@ return (new \PhpCsFixer\Config())
             'yoda_style' => ['equal' => false, 'identical' => false, 'less_and_greater' => false],
 
             // function notation
-            'type_declaration_spaces' => true,
+            'type_declaration_spaces' => ['elements' => ['function', 'property']],
             'native_function_invocation' => ['include' => ['@all']],
             'no_spaces_after_function_name' => true,
             'return_type_declaration' => ['space_before' => 'none'],
@@ -149,7 +149,7 @@ return (new \PhpCsFixer\Config())
             // whitespace
             'compact_nullable_typehint' => true,
             'no_extra_blank_lines' => true,
-            'spaces_inside_parentheses' => true,
+            'spaces_inside_parentheses' => ['space' => 'none'],
             'no_whitespace_in_blank_line' => true,
         ]
     );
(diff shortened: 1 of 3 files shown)


config/php-cs-fixer.php
```diff
@@ -149,7 +149,7 @@
             // whitespace
             'compact_nullable_typehint' => true,
             'no_extra_blank_lines' => true,
-            'no_spaces_inside_parenthesis' => true,
+            'spaces_inside_parentheses' => true,
```
JakeQZ, line 152: [Allowed values for `spaces_inside_parentheses` are `'none'` or `'single'`](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/v3.26.1/doc/rules/whitespace/spaces_inside_parentheses.rst).

JakeQZ (changes requested):
Updates are fine.

I am curious to know why our CI/weekly build seems to keep using the minimum version of the PHIVE Phars, rather than the latest that satisfy the version constraint.  We should have got the errors addressed by #1224 detected within a week of the Psalm release that introduced them, which was made at least three weeks ago.  The build failure addressed by #1221 was picked up by the CI build as expected - but in that case the updated component breaking the build is installed via Composer not PHIVE.  If I run `phive update`, it installs the latest versions satisfying the version constraint.  Are we failing to do this in the GitHub Action?  Beyond the scope of this PR, but if we could sort that out, we wouldn't have to keep bumping the minimum versions in `phars.xml` without a good reason.

The PHP CS Fixer config changes seem to lack RTFM ;;)))  (Sorry, couldn't resist use of that initialism 😏)

oliverklee:
> I am curious to know why our CI/weekly build seems to keep using the minimum version of the PHIVE Phars, rather than the latest that satisfy the version constraint. 

That's because `phive install` (which we're using in the CI pipeline) installs the exact version configured as `installed` in the `phars.xml`, while `phive update <package>` would install the highest version allowed by `version`:

```xml
<phar name="php-cs-fixer" version="^3.13.0" installed="3.13.0" location="./.phive/php-cs-fixer" copy="false"/>
```

`phive update` will also update the `installed` version in the `phars.xml`, but leave the `version` alone.

(I usually update both when updating our CI tools.)

oliverklee:
Ah, it looks like I missed to read part of your comment, @JakeQZ. Sorry for that.

I'd prefer use to have pinned versions for our CI tools: I don't want our CI build to fail if we haven't changed anything (and it's not due to an update to a production dependency).

oliverklee:
> Wasn't immediately obvious to me that the values should be in an array with the keys `'elements'` and `'space'`.

It wasn't to me, either. ;-)

PR #1228, opened 2023-09-10
[TASK] Update to PHPUnit 9.6

oliverklee:


JakeQZ (approved):
Fine, but seems a bit pointless.  Composer will (or at least should) install the very same latest version anyway (unlike PHIVE - see https://github.com/MyIntervals/emogrifier/pull/1227#pullrequestreview-1618943473).  And there are no changes between 9.5 and 9.6 that we specifically require.  Also, the XSD file has not changed apart from the version number being updated in a comment.

oliverklee:
> Composer will (or at least should) install the very same latest version anyway

For a regular `composer install` (without a `composer.lock`) or `composer update`, yes, it will. However, we run our CI pipeline both with the highest and lowest allowed Composer dependencies (to check for problems related to different versions of our dependencies), so for the "lowest" dependency set, we should in  theory have been using PHPUnit 9.5, but for some reason indeed were already using PHPUnit 9.6: https://github.com/MyIntervals/emogrifier/actions/runs/6125941206/job/16628972405

PR #1229, opened 2023-09-10
[CLEANUP] Use more specific type annotations in `CssDocument`

oliverklee:
`list` means `array with numeric continuous keys starting with 0`.


JakeQZ (approved):
I always wondered what `list<>` meant.  You learn something every day.  Thanks.

PR #1230, opened 2023-09-10
[CLEANUP] Use more specific type annotations in `CssInliner`

oliverklee:
`list` means `array with numeric continuous keys starting with 0`.


JakeQZ (approved):
More of the same.

PR #1234, opened 2023-09-11
[TASK] Add rebasing guidance to CONTRIBUTING.md

JakeQZ:
[This procedure](https://github.com/MyIntervals/emogrifier/issues/373#issuecomment-536631092) has had four more years of being tried and tested since originally written.

I keep referring back to it when a rebase is needed.  It never fails.

So I think it needs to be incorporated into the actual docs, rather than just sitting there as a comment.


oliverklee (approved):
Thanks!

PR #1236, opened 2023-09-19
[FEATURE] Add the ability to exclude CSS selectors from being inlined

nlemoine:
Exclude CSS selectors from the stylesheets being inlined. This might for example prevent global CSS rules with universal selector (*) to be inlined on each node.

- Add unit tests
- Normalize selectors spaces/line breaks
- Add CHANGELOG entry
- Add README documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+- Add CSS selectors exclusion feature (#1236)
 
 ### Changed
 
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -629,7 +629,7 @@ class CssInliner extends AbstractHtmlProcessor
             if (\count($this->excludedCssSelectors) > 0) {
                 // Normalize spaces, line breaks & tabs
                 $selectorsNormalized = \array_map(static function (string $selector): string {
-                    return (string) \preg_replace('@\\s++@u', ' ', $selector);
+                    return (string)\preg_replace('@\\s++@u', ' ', $selector);
                 }, $selectors);
 
                 $selectors = \array_filter($selectorsNormalized, function (string $selector): bool {
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -62,6 +62,11 @@ class CssInliner extends AbstractHtmlProcessor
     private $excludedSelectors = [];
 
     /**
+     * @var array<non-empty-string, bool>
+     */
+    private $excludedCssSelectors = [];
+
+    /**
      * @var array<string, bool>
      */
     private $allowedMediaTypes = ['all' => true, 'screen' => true, 'print' => true];
@@ -296,6 +301,36 @@ class CssInliner extends AbstractHtmlProcessor
     {
         if (isset($this->excludedSelectors[$selector])) {
             unset($this->excludedSelectors[$selector]);
+        }
+
+        return $this;
+    }
+
+    /**
+     * Adds a selector to exclude CSS selector from emogrification.
+     *
+     * @param non-empty-string $selector the selector to exclude, e.g., `.editor`
+     *
+     * @return $this
+     */
+    public function addExcludedCssSelector(string $selector): self
+    {
+        $this->excludedCssSelectors[$selector] = true;
+
+        return $this;
+    }
+
+    /**
+     * No longer excludes the CSS selector from emogrification.
+     *
+     * @param non-empty-string $selector the selector to no longer exclude, e.g., `.editor`
+     *
+     * @return $this
+     */
+    public function removeExcludedCssSelector(string $selector): self
+    {
+        if (isset($this->excludedCssSelectors[$selector])) {
+            unset($this->excludedCssSelectors[$selector]);
         }
 
         return $this;
@@ -588,7 +623,21 @@ class CssInliner extends AbstractHtmlProcessor
 
             $mediaQuery = $cssRule->getContainingAtRule();
             $declarationsBlock = $cssRule->getDeclarationAsText();
-            foreach ($cssRule->getSelectors() as $selector) {
+            $selectors = $cssRule->getSelectors();
+
+            // Maybe exclude CSS selectors
+            if (\count($this->excludedCssSelectors) > 0) {
+                // Normalize spaces, line breaks & tabs
+                $selectorsNormalized = \array_map(static function (string $selector): string {
+                    return (string) \preg_replace('@\\s++@u', ' ', $selector);
+                }, $selectors);
+
+                $selectors = \array_filter($selectorsNormalized, function (string $selector): bool {
+                    return !isset($this->excludedCssSelectors[$selector]);
+                });
+            }
+
+            foreach ($selectors as $selector) {
                 // don't process pseudo-elements and behavioral (dynamic) pseudo-classes;
                 // only allow structural pseudo-classes
                 $hasPseudoElement = \strpos($selector, '::') !== false;
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -3142,6 +3142,94 @@ final class CssInlinerTest extends TestCase
 
     /**
      * @test
+     *
+     * @param array<non-empty-string> $exludedSelectors
+     *
+     * @dataProvider excludedCssRuleDataProvider
+     */
+    public function addExcludedCssSelectorCanExcludeCssSelectors(
+        string $css,
+        array $exludedSelectors,
+        string $expectedInlineCss
+    ): void {
+        $subject = $this->buildDebugSubject('<html><body><section><div class="green"></div></section></body></html>');
+
+        foreach ($exludedSelectors as $exludedSelector) {
+            $subject->addExcludedCssSelector($exludedSelector);
+        }
+        $subject->inlineCss($css);
+
+        self::assertStringContainsString(
+            '<section><div class="green" style="' . $expectedInlineCss . '"></div></section>',
+            $subject->renderBodyContent()
+        );
+    }
+
+    /**
+     * @return array<array{non-empty-string, array<non-empty-string>, non-empty-string}>
+     */
+    public function excludedCssRuleDataProvider(): array
+    {
+        return [
+            'simple selector' => [
+                // CSS
+                '* { margin: 0; } .green { color: green; }',
+                // Exclude
+                ['*'],
+                // Expected inline CSS
+                'color: green;',
+            ],
+            'multiple selectors' => [
+                '*, div { border: solid 1px red; } .green { color: green; }',
+                ['*'],
+                'border: solid 1px red; color: green;',
+            ],
+            'descendant selector' => [
+                'section .green { color: green; } .green { border-color: green; }',
+                ['section .green'],
+                'border-color: green;',
+            ],
+            'descendant selector with line break' => [
+                "section\n\n\t.green { color: green; } .green { border-color: green; }",
+                ['section .green'],
+                'border-color: green;',
+            ],
+            'descendant selector with non standard spaces' => [
+                "section\u{a0}.green { color: green; } .green { border-color: green; }",
+                ['section .green'],
+                'border-color: green;',
+            ],
+        ];
+    }
+
+    /**
+     * @test
+     */
+    public function removeExcludedCssSelectorProvidesFluentInterface(): void
+    {
+        $subject = CssInliner::fromHtml('<html></html>');
+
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 - Add support for PHP 8.3 (#1218)
+- Add CSS selectors exclusion feature (#1202)
```
JakeQZ, line 10: This should now be #1236.
```suggestion
- Add CSS selectors exclusion feature (#1236)
```

JakeQZ (changes requested):
The CHANGELOG entry needs updating to reflect the new PR number.  You should be able to accept the suggested change to save time.

Otherwise looks good to me.  Thanks for contributing, and for your patience <3

tests/Unit/CssInlinerTest.php
```diff
@@ -3140,6 +3140,90 @@ public function inlineCssNotInDebugModeIgnoresOnlyInvalidExcludedSelector(): voi
...
+    /**
+     * @test
+     *
+     * @dataProvider excludedCssRuleDataProvider
+     *
+     * @param string $css
+     * @param array<string> $exludedSelectors
+     * @param string $expectedInlineCss
+     */
+    public function addExcludedCssSelectorCanExcludeCssSelectors(string $css, array $exludedSelectors, string $expectedInlineCss): void
```
JakeQZ, line 3152: There's a coding style line length limit of 120 characters.

```suggestion
    public function addExcludedCssSelectorCanExcludeCssSelectors(
        string $css,
        array $exludedSelectors,
        string $expectedInlineCss
    ): void
```

tests/Unit/CssInlinerTest.php
```diff
@@ -3140,6 +3140,90 @@ public function inlineCssNotInDebugModeIgnoresOnlyInvalidExcludedSelector(): voi
...
+    /**
+     * @return array<array{string, array<string>, string}>
+     */
+    public function excludedCssRuleDataProvider(): array
+    {
+        return [
+            'simple selector' => [
+                '* { margin: 0; } .green { color: green; }', // CSS
+                ['*'], // Exclude
+                'color: green;', // Expected inline CSS
```
JakeQZ, line 3176: The coding style checker is configured not to allow comments on the same line as statements.

```suggestion
                // CSS
                '* { margin: 0; } .green { color: green; }',
                // Exclude
                ['*'],
                // Expected inline CSS
                'color: green;',
```

JakeQZ (changes requested):
I've just allowed the CI tools to run ([for some reason](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) they are not run for a new contributor), and they picked up a couple of coding style issues.

Again, I've suggested alterations that may suffice and save you time if you're happy to accept them.

CHANGELOG.md
nlemoine: Yes, I thought the original that contains all discussions was more relevant.

nlemoine:
Hey @JakeQZ, sorry for the very late reply 🙈

I rebased and fixed every changes needed, can you give it a last review?

CHANGELOG.md
```diff
@@ -6,8 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
-
-### Changed
```
JakeQZ, line 10: Could you restore the 'Changed' section here?

We'll remove it when making a release if it is still empty, but for now it acts as a marker/placeholder.

tests/Unit/CssInlinerTest.php
```diff
@@ -3140,6 +3140,97 @@ public function inlineCssNotInDebugModeIgnoresOnlyInvalidExcludedSelector(): voi
...
+     * @param string $css
+     * @param array<string> $exludedSelectors
+     * @param string $expectedInlineCss
+     */
+    public function addExcludedCssSelectorCanExcludeCssSelectors(
+        string $css,
+        array $exludedSelectors,
+        string $expectedInlineCss
+    ): void
+    {
```
JakeQZ, line 3157: I think the `braces_position` error is coming from here - it wants `): void {` all on one line.

JakeQZ (changes requested):
Apologies that the CI checker does not run when you commit.  It found a code layout issue.  Also we'd like the empty 'Changed' section restored in `CHANGELOG.md`.  Otherwise looks good.

tests/Unit/CssInlinerTest.php
JakeQZ: Though TBH, I'd prefer it as you have it - only put the function opening brace on the same line as the closing argument brace if there's no type annotation.  But that's beyond the scope of this.

nlemoine: I ran into issues while trying to run it locally: 
```
  The rules contain unknown fixers: "type_declaration_spaces", "spaces_inside_parentheses" (did you mean "no_spaces_inside_parenthesis"?).
````

**Comment by nlemoine** [comment]
@JakeQZ Thanks you for the prompt review, it should be good now!

File: src/CssInliner.php [hunk 6]
```diff
@@ -61,6 +61,11 @@ class CssInliner extends AbstractHtmlProcessor
      */
     private $excludedSelectors = [];
 
+    /**
+     * @var array<string, bool>
```
oliverklee, line 65: AFAICT, we can even use `non-empty-string` instead of `string` here.

src/CssInliner.php
```diff
@@ -301,6 +306,36 @@ public function removeExcludedSelector(string $selector): self
         return $this;
     }
 
+    /**
+     * Adds a selector to exclude CSS selector from emogrification.
+     *
+     * @param string $selector the selector to exclude, e.g., ".editor"
```
oliverklee, line 312: Ditto.

src/CssInliner.php
```diff
@@ -301,6 +306,36 @@ public function removeExcludedSelector(string $selector): self
...
+    }
+
+    /**
+     * No longer excludes the CSS selector from emogrification.
+     *
+     * @param string $selector the selector to no longer exclude, e.g., ".editor"
+     *
+     * @return $this
+     */
+    public function removeExcludedCssSelector(string $selector): self
```
oliverklee, line 330: Ditto.

src/CssInliner.php
```diff
@@ -301,6 +306,36 @@ public function removeExcludedSelector(string $selector): self
...
+    {
+        $this->excludedCssSelectors[$selector] = true;
+
+        return $this;
+    }
+
+    /**
+     * No longer excludes the CSS selector from emogrification.
+     *
+     * @param string $selector the selector to no longer exclude, e.g., ".editor"
```
oliverklee, line 326: Let's use backticks instead of single quotes to to mark `.editor` as code.

src/CssInliner.php
```diff
@@ -588,7 +623,21 @@ private function collateCssRules(CssDocument $parsedCss): array
 
             $mediaQuery = $cssRule->getContainingAtRule();
             $declarationsBlock = $cssRule->getDeclarationAsText();
-            foreach ($cssRule->getSelectors() as $selector) {
+            $selectors = $cssRule->getSelectors();
+
+            // Maybe exclude CSS selectors
+            if (!empty($this->excludedCssSelectors)) {
+                // Normalize spaces, line breaks & tabs
+                $selectorsNormalized = \array_map(static function ($selector) {
```
oliverklee, line 631: Please let's add native type declarations to the static function.

src/CssInliner.php
```diff
@@ -588,7 +623,21 @@ private function collateCssRules(CssDocument $parsedCss): array
...
+                    return \preg_replace('@\\s++@u', ' ', $selector);
+                }, $selectors);
+
+                $selectors = \array_filter($selectorsNormalized, function ($selector) {
```
oliverklee, line 635: Ditto.

src/CssInliner.php
```diff
@@ -588,7 +623,21 @@ private function collateCssRules(CssDocument $parsedCss): array
 
             $mediaQuery = $cssRule->getContainingAtRule();
             $declarationsBlock = $cssRule->getDeclarationAsText();
-            foreach ($cssRule->getSelectors() as $selector) {
+            $selectors = $cssRule->getSelectors();
+
+            // Maybe exclude CSS selectors
+            if (!empty($this->excludedCssSelectors)) {
```
oliverklee, line 629: Do we really need this condition? Or could we rely on the code below to be a no-op for no excluded selectors?

tests/Unit/CssInlinerTest.php
```diff
@@ -3140,6 +3140,96 @@ public function inlineCssNotInDebugModeIgnoresOnlyInvalidExcludedSelector(): voi
...
     }
 
+    /**
+     * @test
+     *
+     * @dataProvider excludedCssRuleDataProvider
+     *
+     * @param string $css
+     * @param array<string> $exludedSelectors
+     * @param string $expectedInlineCss
```
oliverklee, line 3150: Please let's not duplicate in PHPDoc what we already have in native type declarations.

oliverklee (changes requested):

src/CssInliner.php
nlemoine: I guess we do, otherwise we'll be running `preg_replace` on all selectors for no valid reason.

src/CssInliner.php
nlemoine: Ditto what?

src/CssInliner.php
nlemoine: Actually, no. Because `array<non-empty-string, bool>` is a child of `non-empty-array` and will raise analysis errors.

tests/Unit/CssInlinerTest.php
nlemoine: Consider the https://cs.symfony.com/doc/rules/phpdoc/no_superfluous_phpdoc_tags.html rule then :)

src/CssInliner.php
oliverklee: I'm confused - are you by change thinking of `non-empty-array<string, bool>` instead of `array<non-empty-string, bool>`? (As far as I understand it, the latter would also allow an empty array.)

src/CssInliner.php
oliverklee: Oh, it seems I had clicked on the wrong line, and now I cannot remember what I was referring to in the first place. ;-)

tests/Unit/CssInlinerTest.php
```diff
@@ -3140,6 +3140,92 @@ public function inlineCssNotInDebugModeIgnoresOnlyInvalidExcludedSelector(): voi
...
     }
 
+    /**
+     * @test
+     *
+     * @dataProvider excludedCssRuleDataProvider
+     */
+    public function addExcludedCssSelectorCanExcludeCssSelectors(
+        string $css,
+        array $exludedSelectors,
```
oliverklee, line 3150: We'll need a PHPDoc annotation:

`@param array<string> $exludedSelectors`

oliverklee (changes requested):

tests/Unit/CssInlinerTest.php
JakeQZ: > I ran into issues while trying to run it locally:

Sounds like some of the development tools (e.g. PHP CS Fixer) need updating.  [The `CONTRIBUTING` file is out of date](https://github.com/MyIntervals/emogrifier/issues/373).  Some tools are now installed with PHIVE.  Seems like you have previously installed them with PHIVE (otherwise I wouldn't expect them run at all), and just need to run `phive update` from the project root directory.

src/CssInliner.php
nlemoine: > As far as I understand it, the latter would also allow an empty array

Seems like it doesn't, here's the output of Psalm:
```
ERROR: PropertyTypeCoercion - src/CssInliner.php:318:9 - $this->excludedCssSelectors expects 'array<non-empty-string, bool>',  parent type 'non-empty-array<string, bool>' provided (see https://psalm.dev/198)
        $this->excludedCssSelectors[$selector] = true;
```

nlemoine: Error thrown using:
```php
    /**
     * @var array<non-empty-string, bool>
     */
    private $excludedCssSelectors = [];
```

nlemoine: Ok found it, `add/removeExcludedCssSelector` also has to have `non-empty-string`, not just string (which can be empty).

oliverklee (approved):

JakeQZ (approved):
LGTM now.  Thanks for contributing and perservering to completion. <3

nlemoine:
Thanks @JakeQZ & @oliverklee ! 🥳

PR #1238, opened 2023-10-16
[TASK] Prepare release of version 7.1.0

oliverklee:


JakeQZ (commented):
Totally agree with version number, since to support PHP 8.3, we don't need to drop support for PHP 7.3, there are no other breaking changes, and it's not merely a bugfix release.  The 8.0 release is when we drop PHP 7.3 (or make some other non-backwards-compatible change, which might not actually involve dropping PHP 7.3 as planned).

However, as this is on the main release line, shouldn't the 'Added' and 'Changed' sections be cleared for the next `x.y.z` release?  Won't 8.0.0 include everything in 7.1.0 by definition - I mean, if the next major or minor release is indeed 8.0.0, there will never be a 7.2.0, only ever perpahs a 7.1.1 - ?

I checked the 'Steps to release a new version' and indeed the `branch-alias` does not need changing for a minor release.

oliverklee:
> However, as this is on the main release line, shouldn't the 'Added' and 'Changed' sections be cleared for the next x.y.z release? Won't 8.0.0 include everything in 7.1.0 by definition - I mean, if the next major or minor release is indeed 8.0.0, there will never be a 7.2.0, only ever perpahs a 7.1.1 - ?

I'm confused … :-) Are you referring to the alias in the `composer.json` or the changelog?

(Also, I've noticed that in our release checklist, the "next version" would need a look into the future. I think it's our best guess then.)

JakeQZ:
> I'm confused … :-) Are you referring to the alias in the `composer.json` or the changelog?

Sorry, my waffliness may have muddied what I was saying.

I meant the changelog should not have the changes for 7.1.0 duplicated into the section for the next (x.y.z) release, because the next main (x.y.z) release will include the 7.1.0 changes (by definition according to semver, if I understand correctly).
 
> (Also, I've noticed that in our release checklist, the "next version" would need a look into the future. I think it's our best guess then.)

Yes, that's what I was trying to say.  No changes to `composer.json` are needed, since the next major version will still be 8.0.0.  I was merely just confirming the correctness of there being no changes made for step 1.

oliverklee:
> I meant the changelog should not have the changes for 7.1.0 duplicated into the section for the next (x.y.z) release, because the next main (x.y.z) release will include the 7.1.0 changes (by definition according to semver, if I understand correctly).

Yes, I agree. This also is what we've been doing so far.

Oh, I see that I indeed duplicated things in this PR. Thanks for catching this! I've repushed now.

JakeQZ (approved):

PR #1253, opened 2024-02-13
[TASK] Drop PHP_CodeSniffer

oliverklee:
This tool is not well-maintained anymore, and it has no PER2 coding standard, which is a blocker for us moving to PER2.

PHP-CS-Fixer already has a PER2 coding standard. If we use PHP-CS-Fixer on PER2 and PHP_CodeSniffer on PSR12 (the latest available standard there), both tools would keep overwriting each other's changes endlessly, and they would never agree on the correct formatting.


JakeQZ (approved):
Agree it seems to duplicate functionality with 'The Fixer', and they could end up warring.

PR #1255, opened 2024-02-13
[TASK] Streamline `CONTRIBUTING.md`

oliverklee:
- fix the spelling
- convert to UTF-8

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -136,12 +136,12 @@ and final approval, it may be required that you rebase your changes against the
 latest `main` branch.
 
 There are potential pitfalls here if you follow the suggestions from `git`,
-which could leave your banch in an unrecoverable mess,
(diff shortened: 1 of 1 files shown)


JakeQZ (approved):
Strange how it got in this state.  I've checked that both editors I use have UTF-8 set as default for both new and loaded files (though they may not have at the time).  IIRC I copied the contents from https://github.com/MyIntervals/emogrifier/issues/373#issuecomment-536631092 and just changed `master` to `main`.  The presence of the curly quotes implies that I probably originally wrote it using Word - which has a [spell chequer](http://www.davidpbrown.co.uk/poetry/martha-snow.html).

I originally used single quotes for 'diff' and 'patching base', but that's a personal style to distinguish terminology from things actually said (written or spoken).  Either is fine.  And agree 'feature branch' doesn't need quotes around 'feature'.

PR #1256, opened 2024-02-24

oliverklee:
Also update the Psalm baseline.

Also autoformat the code according to the new PHP-CS-Fixer version.


JakeQZ (approved):
I see theres a few new Psalm issues to look into.

PR #1263, opened 2024-06-06
[TASK] Update the QA tools

oliverklee:
Also reformat the code with the updated version of PHP-CS-Fixer.

Also recreate the Psalm baseline with the additional warnings that get detected now.


JakeQZ (approved):
Would be nice to resolve the Psalm warnings at some point, to get back to an empty `baseline`.  I might find some time one day...

PR #1264, opened 2024-06-06
[TASK] Update our coding standard

### LINKED ISSUE #1262: Switch to the PER2 coding standard and reduce the number of custom rules
Basically, copy the rules from https://github.com/MyIntervals/PHP-CSS-Parser/tree/main/config .

oliverklee:
Switch to the PER2 coding standard and reduce the number of custom rules.

This basically aligns our code style with that used by the PHP-CSS-Parser project, which is very close to the PER-2 coding standard.

Fixes #1262

### DIFF
--- a/config/php-cs-fixer.php
+++ b/config/php-cs-fixer.php
@@ -2,154 +2,36 @@
 
 declare(strict_types=1);
 
-if (PHP_SAPI !== 'cli') {
-    die('This script supports command line usage only. Please check your command.');
-}
-
 return (new \PhpCsFixer\Config())
     ->setRiskyAllowed(true)
     ->setRules(
         [
-            // rule sets
-            '@DoctrineAnnotation' => true,
-            '@PHP56Migration:risky' => true,
-            '@PHP70Migration' => true,
-            '@PHP70Migration:risky' => true,
-            '@PHP71Migration' => true,
-            '@PHP71Migration:risky' => true,
+            '@PER-CS2.0' => true,
+            '@PER-CS2.0:risky' => true,
+
+            '@PHPUnit50Migration:risky' => true,
+            '@PHPUnit52Migration:risky' => true,
+            '@PHPUnit54Migration:risky' => true,
+            '@PHPUnit55Migration:risky' => true,
+            '@PHPUnit56Migration:risky' => true,
             '@PHPUnit57Migration:risky' => true,
             '@PHPUnit60Migration:risky' => true,
             '@PHPUnit75Migration:risky' => true,
             '@PHPUnit84Migration:risky' => true,
-            '@PER' => true,
 
-            // alias
-            'no_alias_functions' => true,
+            // needed for PHP 7.x
+            'trailing_comma_in_multiline' => ['after_heredoc' => true, 'elements' => ['arrays', 'match']],
 
-            // array notation
-            'array_syntax' => ['syntax' => 'short'],
-            'no_trailing_comma_in_singleline' => true,
-            'no_whitespace_before_comma_in_array' => true,
-            'whitespace_after_comma_in_array' => true,
-
-            // basic
-            'encoding' => true,
-            'psr_autoloading' => true,
-
-            // casing
-            'magic_constant_casing' => true,
-            'native_function_casing' => true,
-
-            // cast notation
-            'cast_spaces' => ['space' => 'none'],
-            'lowercase_cast' => true,
-            'modernize_types_casting' => true,
-            'no_short_bool_cast' => true,
-            'short_scalar_cast' => true,
-
-            // class notation
-            'class_attributes_separation' => true,
-            'no_blank_lines_after_class_opening' => true,
-            'no_php4_constructor' => true,
-
-            // class usage
-            // (no rules used from this section)
-
-            // comment
-            'no_empty_comment' => true,
-            'single_line_comment_style' => true,
-
-            // constant notation
-            // (no rules used from this section)
-
-            // control structure
-            'elseif' => true,
-            'no_superfluous_elseif' => true,
-            'no_unneeded_control_parentheses' => true,
-            'no_unneeded_curly_braces' => true,
-            'no_useless_else' => true,
-            'trailing_comma_in_multiline' => true,
-            'yoda_style' => ['equal' => false, 'identical' => false, 'less_and_greater' => false],
-
-            // function notation
-            'type_declaration_spaces' => ['elements' => ['function', 'property']],
-            'native_function_invocation' => ['include' => ['@all']],
-            'no_spaces_after_function_name' => true,
-            'return_type_declaration' => ['space_before' => 'none'],
-
-            // import
-            'no_leading_import_slash' => true,
-            'no_unused_imports' => true,
-            'ordered_imports' => true,
-
-            // language construct
-            'combine_consecutive_issets' => true,
-            'combine_consecutive_unsets' => true,
-            'declare_equal_normalize' => true,
-            'dir_constant' => true,
-            'is_null' => true,
-
-            // list notation
-            // (no rules used from this section)
-
-            // namespace notation
-            'no_leading_namespace_whitespace' => true,
-
-            // naming
-            // (no rules used from this section)
-
-            // operator
-            'concat_space' => ['spacing' => 'one'],
-            'new_with_braces' => true,
-            'standardize_not_equals' => true,
-            'ternary_operator_spaces' => true,
-            'ternary_to_null_coalescing' => true,
-            'unary_operator_spaces' => true,
-
-            // PHP tag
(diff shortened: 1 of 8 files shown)


oliverklee:
It looks like I'll need to tweak the ruleset a bit to retain compatibility with PHP 7.3 and 7.4.

oliverklee:
@JakeQZ friendly ping :-)

JakeQZ:
> @JakeQZ friendly ping :-)

Sorry, I missed this amongst the other 464 emails in the last 5 days.  Not even exaggerating - I just counted them ;)

config/php-cs-fixer.php
```diff
@@ -2,154 +2,36 @@
...
-            'native_function_invocation' => ['include' => ['@all']],
-            'no_spaces_after_function_name' => true,
-            'return_type_declaration' => ['space_before' => 'none'],
 
-            // import
-            'no_leading_import_slash' => true,
-            'no_unused_imports' => true,
-            'ordered_imports' => true,
+            // overwrite the PER2 defaults to restore compatibility with PHP 7.x
+            'trailing_comma_in_multiline' => ['elements' => ['arrays', 'match']],
```
JakeQZ, line 23: As per issue on sister project, we can't include `match` here, until PHP-CS-Fixer fixes the issue.

JakeQZ (changes requested):
I'll need to go through the removed settings in more detail, which will take a bit of time I don't have right now.  But there is the `match` PHP 8 issue.

config/php-cs-fixer.php
oliverklee: It's removed now.

config/php-cs-fixer.php
```diff
@@ -2,154 +2,36 @@
...
             'php_unit_set_up_tear_down_visibility' => true,
+            'php_unit_test_annotation' => ['style' => 'annotation'],
             'php_unit_test_case_static_method_calls' => ['call_type' => 'self'],
-
-            // PHPDoc
-            'no_blank_lines_after_phpdoc' => true,
-            'no_empty_phpdoc' => true,
-            'phpdoc_add_missing_param_annotation' => true,
-            'phpdoc_indent' => true,
-            'phpdoc_no_package' => true,
```
JakeQZ, line 125: Ditto these two.

config/php-cs-fixer.php
```diff
@@ -2,154 +2,36 @@
...
-            'phpdoc_scalar' => true,
-            'phpdoc_separation' => true,
-            'phpdoc_trim' => true,
```
JakeQZ, line 128: And this.

config/php-cs-fixer.php
```diff
@@ -2,154 +2,36 @@
...
-            'phpdoc_types' => true,
-            'phpdoc_types_order' => ['null_adjustment' => 'always_last', 'sort_algorithm' => 'none'],
```
JakeQZ, line 130: And this.

config/php-cs-fixer.php
```diff
@@ -2,154 +2,36 @@
...
-
-            // return notation
-            'no_useless_return' => true,
```
JakeQZ, line 133: And this.

config/php-cs-fixer.php
```diff
@@ -2,154 +2,36 @@
...
-
-            // semicolon
-            'multiline_whitespace_before_semicolons' => true,
-            'no_empty_statement' => true,
-            'no_singleline_whitespace_before_semicolons' => true,
-            'semicolon_after_instruction' => true,
```
JakeQZ, line 139: And these three.

config/php-cs-fixer.php
```diff
@@ -2,154 +2,36 @@
...
-            'space_after_semicolon' => true,
-
-            // strict
-            'declare_strict_types' => true,
```
JakeQZ, line 143: And this.

config/php-cs-fixer.php
```diff
@@ -2,154 +2,36 @@
...
-
-            // string notation
-            'escape_implicit_backslashes' => ['single_quoted' => true],
```
JakeQZ, line 146: This should be replaced with `string_implicit_backslashes => ['single_quoted' => 'escape']`, which isn't in any of the rule sets used (`escape_implicit_backslashes` is deprecated).

JakeQZ (changes requested):
I've been through all the added and removed rules, and checked what the defaults are, and whether they are included in any of the rule sets that will be active following this change.

Quite a few of the removed rules I think we need to keep, because they are not applied by default, nor included in an rule set.

Many of the other removed rules seemed to be incorrectly specified as a Boolean, when in fact what is expected is an array of configuration values.  I've ignored those, assuming they were invalid and would have had no effect anyway.  I don't know if I'm missing something, like if it's possible to provide a Boolean when an array is expected, and that would result in enabling all options (though in some cases the options aren't Boolean either).

Some of the added rules appear unnecessary as they are the default values anyway.  At least one of the added rules is given as a Boolean when an array is expected.

config/php-cs-fixer.php
```diff
@@ -2,154 +2,36 @@
...
-            'no_whitespace_before_comma_in_array' => true,
-            'whitespace_after_comma_in_array' => true,
-
-            // basic
-            'encoding' => true,
-            'psr_autoloading' => true,
-
-            // casing
-            'magic_constant_casing' => true,
-            'native_function_casing' => true,
```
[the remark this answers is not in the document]
JakeQZ: GitHub diff isn't highlighting the lines I selected, so you'll need to check the smallprint, e.g. "Comment on lines -39 to -41" to see how many preceding lines I'm referring to.  If there's nothing there, it means I'm just commenting on a single line.

config/php-cs-fixer.php
```diff
@@ -2,154 +2,36 @@
...
-
-            // constant notation
-            // (no rules used from this section)
-
-            // control structure
-            'elseif' => true,
-            'no_superfluous_elseif' => true,
-            'no_unneeded_control_parentheses' => true,
-            'no_unneeded_curly_braces' => true,
-            'no_useless_else' => true,
```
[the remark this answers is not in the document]
oliverklee: Now that I think of if, I like the first variant more. So I'd prefer to not have this rule for now.

config/php-cs-fixer.php
oliverklee: Oh, it seems `match` was there 2 times. I've removed the other one now as well, and also created https://github.com/MyIntervals/PHP-CSS-Parser/pull/629.

oliverklee:
@JakeQZ Thank your for doing the painstaking work of going through this with a fine-grained comb! :pray: I've now updated the ruleset accordingly.

config/php-cs-fixer.php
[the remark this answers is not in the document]
JakeQZ: I think I agree.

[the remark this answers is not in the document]
JakeQZ: On that basis, we also need to remove the `no_superfluous_elseif`.

config/php-cs-fixer.php
```diff
@@ -2,154 +2,96 @@
...
             // comment
             'no_empty_comment' => true,
-            'single_line_comment_style' => true,
-
-            // constant notation
-            // (no rules used from this section)
 
             // control structure
-            'elseif' => true,
             'no_superfluous_elseif' => true,
```
JakeQZ, line 40: If we remove `no_useless_else`, we should also remove `no_superfluous_elseif`.

JakeQZ (commented):
I think I've checked most if not all the initial review change suggestions.  However, it's tricky since GitHub drops some of the original review comments from the diff if it can't work out where to place them now.  (In some cases that's quite poor - e..g. lines added then removed - the review comment disappears too.)

One thing I have spotted: if we don't want [`no_useless_else`](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/control_structure/no_useless_else.rst), we probably also don't want [`no_superfluous_elseif`](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/control_structure/no_superfluous_elseif.rst), for consistency.

config/php-cs-fixer.php
oliverklee: Done.

JakeQZ (approved):
Having finalized this for now (it will change in future, I'm sure) here, we should probably update the equivalent on the sister project to match.  And like the contrubution guidelines, CoC, and other bits and bobs, keep them in sync.

I wonder if there's a way of using a template to keep cross-project files like these in sync.  You mentioned that the CoC came from a GitHub template (I think).  Can we introduce our own repository-owner templates to help?

PR #1270, opened 2024-07-04
[TASK] Add a PHP-CS-Fixer rule to prefer single-quoted string literals

oliverklee:
As we already are doing this in our code, this added rule does not cause any changes to the code.


JakeQZ (approved):
Thanks for finding this config setting.

PR #1271, opened 2024-07-04
[TASK] Add a fixer rule to ensure a PHP 8.4 compatibility

### LINKED ISSUE #1269: Add a PHP-CS-Fixer rule to avoid deprecation warnings in PHP 8.4
https://cs.symfony.com/doc/rules/function_notation/nullable_type_declaration_for_default_null_value.html

oliverklee:
Ensure that parameters that have a `null` default value also have `null` as part of their type declarations.

Also ensure a uniform notation for nullable types.

https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated

Fixes #1269

### DIFF
--- a/config/php-cs-fixer.php
+++ b/config/php-cs-fixer.php
@@ -41,6 +41,10 @@ return (new \PhpCsFixer\Config())
 
             // function notation
             'native_function_invocation' => ['include' => ['@all']],
+            'nullable_type_declaration' => [
+                'syntax' => 'question_mark',
+            ],
+            'nullable_type_declaration_for_default_null_value' => true,
 
             // import
             'no_unused_imports' => true,
--- a/config/php-cs-fixer.php
+++ b/config/php-cs-fixer.php
@@ -41,9 +41,6 @@ return (new \PhpCsFixer\Config())
 
             // function notation
             'native_function_invocation' => ['include' => ['@all']],
-            'nullable_type_declaration' => [
-                'syntax' => 'question_mark',
-            ],
             'nullable_type_declaration_for_default_null_value' => true,
 
             // import
@@ -54,6 +51,7 @@ return (new \PhpCsFixer\Config())
             'combine_consecutive_unsets' => true,
             'dir_constant' => true,
             'is_null' => true,
+            'nullable_type_declaration' => true,
 
             // namespace notation
             'no_leading_namespace_whitespace' => true,


config/php-cs-fixer.php
```diff
@@ -41,6 +41,10 @@
 
             // function notation
             'native_function_invocation' => ['include' => ['@all']],
+            'nullable_type_declaration' => [
+                'syntax' => 'question_mark',
+            ],
```
JakeQZ, line 46: This seems to be the [default value, at least in the latest version](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/language_construct/nullable_type_declaration.rst), so we don't need to explicitly specify it.  Also, it would belong in the `language_construct` section.

config/php-cs-fixer.php
```diff
@@ -41,6 +41,10 @@
...
+            'nullable_type_declaration_for_default_null_value' => true,
```
JakeQZ, line 47: This [should be an array](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/function_notation/nullable_type_declaration_for_default_null_value.rst) like `['use_nullable_type_declaration' => true]`.  However, it's deprecated and the default value is `true` anyway.

JakeQZ (changes requested):
I don't think we need either of these, as they don't change default values and/or are deprecated.  If we want them, they're either in the wrong section, or with incorrect syntax.

config/php-cs-fixer.php
JakeQZ: I tested this with PHP-CSS-Parser.  It seems that a simple value of `true` does work, presumably for backward-compatibllity.  Also, it seems that in the version of PHP-CS-Fixer we have there (3.59.3, which seems to be the latest), this is NOT the default setting.  So I am a bit mystified.

I am using PHP 7.4.  We have already found that PHP-CS-Fixer behaves differently depending on PHP version in the execution environment.

I think we need to run the fixer with all supported PHP versions and/or get them to sort that problem out.  The execution environment should not affect behaviour.  It's the target environment that matters.

JakeQZ (commented):
There's a significant issue with PHP-CS-Fixer's behaviour (apparently) incorrectly depending on the execution environment.

oliverklee:
> I think we need to run the fixer with all supported PHP versions and/or get them to sort that problem out. The execution environment should not affect behaviour. It's the target environment that matters.

That it out of scope for this PR. So please let's have a separate ticket for that and then have a separate PR for that.

oliverklee:
I'm not completely sure that getting the same results for PHP-CS-Fixer for all supported PHP versions is feasible as the PHP parsing used by the tools seems to depend on the PHP version it runs on: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/2378

JakeQZ:
> I'm not completely sure that getting the same results for PHP-CS-Fixer for all supported PHP versions is feasible as the PHP parsing used by the tools seems to depend on the PHP version it runs on: [PHP-CS-Fixer/PHP-CS-Fixer#2378](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/2378)

Thinking about this, I don't think it's a problem.  If some of the rules require a later PHP version, they just simply won't be run.  So a developer using an older supported PHP version might not get all fixes applied (or be told about things to fix until committing a PR to GitHub).

However, what I don't understand is that the [documentation for `nullable_type_declaration_for_default_null_value`](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/function_notation/nullable_type_declaration_for_default_null_value.rst) says it should be a configutation array with a key `use_nullable_type_declaration` whose default value is `true`, and that it is deprecated, and [that deprecation is included in 3.59.3](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/commit/47530e1750be1e88ab54c2a6d33520827cf10691) which we are using.  But without it, we don't get the default behaviour.  I also don't understand how or why a setting the rule to the primitive value `true` (instead of the configuration array) also achieves the result of setting the configuration parameter to what is suppsed to be its default value anyway.

oliverklee:
My best guesses are this:
- `true` means "enable this rule and use the default configuration".
- The documentation might either not always be up to date, or it might be for the development version, not for the released version.

JakeQZ:
> My best guesses are this:
> 
>     * `true` means "enable this rule and use the default configuration".
> 
>     * The documentation might either not always be up to date, or it might be for the development version, not for the released version.

I think the first seems to be the case (and sometimes the second may be when changes are yet to be released, though the documentation seems to be auto-generated from the actual `Fixer` classes, so I would expect it to be up-to-date).

Some rules have nothing to configure, and can just be enabled or disabled (disabled by default).  Other rules have configuration options, which can be explicitly specified but also have default values.  For such rules, `true` can be used instead of an array of configuration options, to enable the rule with the default options.  (I don't know if any empty array, being 'falsy', would have the same effect as `true`.)  The default configuration options only apply if the rule is enabled, which by default it is not.

I couldn't find this clearly stated anywhere in the docs.

I think I have been conflating default configuration options (when a rule is enabled) and default rule application (which is always 'not applied'), so may need to revisit #1264 - I accepted all removals of rules with configuration options where we simply provided `true`, since their specification didn't seem to be correct; but some of those we actually may want to retain.

In the case of `nullable_type_declaration_for_default_null_value`, I now realize (IIUC) that it's the configration option `use_nullable_type_declaration` that has been deprecated, not the rule itself!

So for this PR, the second change is correct.  However the first change is in the wrong section, and since the configuration option is the default anyway, we could just use `true` instead of a configuration array.

config/php-cs-fixer.php
JakeQZ: The default for `syntax` is `question_mark`, so I think we could just have `'nullable_type_declaration' => true` (based on my renewed understanding of how the config system works).  But maybe we'd like to explicitly state the preference anyway - IDK.

It's definitely in the wrong section though, and should be in `language_construct`.

config/php-cs-fixer.php
JakeQZ: Based on my renewed understanding of how the config system works, this change *is* actually correct :)

JakeQZ (changes requested):
Finally, I think the only thing that needs to be corrected here is that `nullable_type_declaration` is in the wrong section.

The configuration option given is the default, so it might not be necessary to provide it.  However, it's possible (maybe unlikely) that for some previous supported PHP version a developer may be using it is otherwise.  IDK whether we should explicitly specify it.  Your call...

config/php-cs-fixer.php
oliverklee: Done and repushed.

PR #1272, opened 2024-07-05
[CLEANUP] Remove unnecessary `md5()` call in `CssInliner::getCssSelectorPrecedence()`

TimWolla:
The input selector is guaranteed to be a string, so we can directly index the cache array with it. This avoids the overhead of calculating the cryptographic hash and the creation of a fresh string, makes the array more friendly for manual inspection using a debugger and avoids using a function that might be deprecated with PHP 8.4.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -742,7 +742,7 @@ class CssInliner extends AbstractHtmlProcessor
      */
     private function getCssSelectorPrecedence(string $selector): int
     {
-        $selectorKey = \md5($selector);
+        $selectorKey = $selector;
         if (isset($this->caches[self::CACHE_KEY_SELECTOR][$selectorKey])) {
             return $this->caches[self::CACHE_KEY_SELECTOR][$selectorKey];
         }


src/CssInliner.php
```diff
@@ -742,7 +742,7 @@ private function sortBySelectorPrecedence(array $first, array $second): int
      */
     private function getCssSelectorPrecedence(string $selector): int
     {
-        $selectorKey = \md5($selector);
+        $selectorKey = $selector;
         if (isset($this->caches[self::CACHE_KEY_SELECTOR][$selectorKey])) {
             return $this->caches[self::CACHE_KEY_SELECTOR][$selectorKey];
```
oliverklee, line 747: I think we then can directly use `$selector` as array keys and drop the local variable `$selectorKey`.

oliverklee (changes requested):
Thanks!

src/CssInliner.php
TimWolla: The `$selector` variable is reassigned in the loop in line 756, that's why I kept the `$selectorKey`.

oliverklee (approved):
LGTM.

PR #1273, opened 2024-07-06

oliverklee:


oliverklee:
> Should this not be updating the minimum version, as well as the installed version?

For `phive install`, it doesn't make any difference, as this uses the version set as `installed`.

Pros for keeping it like this:
- less work when updating (only `phive update`, but no manual editing of the file)

Cons for keeping it like this:
- `version` and `installed` are out of sync

I'd be fine either way, and would slightly prefer the version where only the `installed` version is increased as this makes the updates for me less of a hassle. Would you be fine with this?

JakeQZ:
> I'd be fine either way, and would slightly prefer the version where only the `installed` version is increased as this makes the updates for me less of a hassle. Would you be okay with this?

Yes, that would be fine.  I hadn't realized the `installed` version was more than just record-keeping for the local development environment.

JakeQZ (approved):

PR #1278, opened 2024-08-23
[FEATURE] Add support for PHP 8.4

oliverklee:


oliverklee:
Currently blocked by https://github.com/MyIntervals/PHP-CSS-Parser/pull/677.

davidrans:
Is this still blocked? https://github.com/MyIntervals/PHP-CSS-Parser/pull/677 was closed because it was unnecessary

oliverklee:
Hi @davidrans, thanks for the ping!

Yes, this is still blocked. We'll need to create a new release of PHP-CSS-Parser first.

oliverklee:
Symfony 4.4 is not compatible with PHP 8.4 due to an explicitly nullable constructor parameter:
https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/CssSelector/XPath/Translator.php

Symfony 4.4 also won't get any fixes anymore as it reached its EOL in November 2023 according to https://symfony.com/releases.

I propose we skip the unit tests on CI for PHP 8.4 with the lowest dependencies (so we can release version 7.3.0) and drop support for Symfony 4.4 in version 8.0.0.

What do you think, @JakeQZ?

JakeQZ:
> I propose we skip the unit tests on CI for PHP 8.4 with the lowest dependencies (so we can release version 7.3.0) and drop support for Symfony 4.4 in version 8.0.0.

I think that makes sense, and is preferable to dropping Symfony 4.4 support in the same release that supports PHP 8.4 as well as including recent features and fixes.  It's possible (though unlikely) that someone might be using PHP 8.4 with a complex set of dependencies that means they need the older Symfony version, but even if so, these are not errors, just deprecation warnings.

JakeQZ:
> I propose we skip the unit tests on CI for PHP 8.4 with the lowest dependencies

PS.  I'm not sure how you'd get the GitHub Action matrix shaped like a Tetяis piece...

JakeQZ (approved):

PR #1280, opened 2024-08-23
[TASK] Switch static analysis for the CI jobs to PHP 8.3

oliverklee:
This brings free performance.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - '8.2'
+          - '8.3'
 
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


oliverklee:
Note: After/before this has been (force-)merged, we'll need to switch the branch protection configuration to the new jobs (as the required jobs are currently set to PHP 8.2).

JakeQZ (approved):
All bases covered.

JakeQZ:
Needs some repo config change to clear out the old "expected" actions.  What is that?

oliverklee:
> Needs some repo config change to clear out the old "expected" actions. What is that?

1. Settings > Branches > edit the `main` branch protection rule, which ends us up at https://github.com/MyIntervals/emogrifier/settings/branch_protection_rules/870902
2. remove the corresponding 8.2 rules from "Require status checks to pass before merging "
3. add the corresponding 8.3 rules there using find-as-you-type

PR #1281, opened 2024-08-23
[TASK] Link the contribution guidelines from the README

oliverklee:


oliverklee:
> I think I've already mentioned elsewhere that the release steps probably belong there rather than here. And perhaps that GitHub doesn't make it easy to find the the contribution guidelines.

Possibly. :wink: I've just created #1284 and #1285 for this.

PR #1283, opened 2024-08-24
[TASK] Switch our type checks from Psalm to PHPStan

oliverklee:
PHPStan does similar things to Psalm. The PHPStan project is developed a lot more actively, though.

(Also, unlike Psalm, PHPStan is compatible with our PHP version requirements, which allows us to drop another package from PHIVE.)

We're directly starting with the highest level (else, where would be the fun in that?) and will fix the listed warnings later.

We'll also add a package with additional extra-strict rules later.


oliverklee:
Note: We'll need to update the CI configuration to expect the PHPStan job instead of the Psalm job after this PR has been merged.

oliverklee:
Another note: The PHPUnit project recently switched from Psalm to PHPStan for the exact same reason: https://github.com/sebastianbergmann/phpunit/commit/ef2df9aab0b4c7e74759a5691cef58f97a5d14a5

JakeQZ (commented):
> The PHPStan project is developed a lot more actively, though.

In my limited eperience, I've found Psalm to be much more thorough (sometimes to the point of frustration).  So perhaps the 'activity' is because PHPStan still has some catching up to do, while Psalm is a mature product that only needs updating for new PHP features.

I also don't see much of a reason not to have both:
  - Pros:
    - More robust code;
  - Cons:
    - More hassle for contributors;
    - More CPU time used for CI builds.
    
I'm happy with the change to switch, just thought it needs a little more consideration before taking the plunge.

> (else, where would be the fun in that?)

🤣

JakeQZ:
> Another note: The PHPUnit project recently switched from Psalm to PHPStan for the exact same reason: [sebastianbergmann/phpunit@ef2df9a](https://github.com/sebastianbergmann/phpunit/commit/ef2df9aab0b4c7e74759a5691cef58f97a5d14a5)

How can I find the PR and issue discussion for this?  I don't see any comments at all, let alone the rationale for it.

oliverklee:
> I also don't see much of a reason not to have both

I'd strongly prefer to have only one. The reason it that both tools do almost the same job, but catering to both with suitable PHPDoc annotation is a hassle I'd like to avoid. (Also, I use PHPStan in all my other projects, and this would allow me to not keep up to date with Psalm development.)

> How can I find the PR and issue discussion for this? I don't see any comments at all, let alone the rationale for it.

I failed to find the PR for that commit, too. Concerning the rationale: I saw the commit and then asked Sebastian (the author of PHPUnit) about it the next time we met for a board gaming session. :slightly_smiling_face:

JakeQZ (approved):
> I'd strongly prefer to have only one. The reason it that both tools do almost the same job, but catering to both with suitable PHPDoc annotation is a hassle I'd like to avoid.

Yeah, I remember that other 'con' from previous discussion – it’s potentially impossible to satisfy both in some situations, or even different versions of Psalm.

Psalm had an insistence that you provided a type annotation for a local variable whenever it could not determine its type.  But then complained that the type annotation was unnecessary when the new version was able to determine the type.

> I failed to find the PR for that commit, too. Concerning the rationale: I saw the commit and then asked Sebastian (the author of PHPUnit) about it the next time we met for a board gaming session. 🙂

Small world.  Was this F2F/OTB or online?  What games?  I play a bit on BGA, mostly Ticket to Ride, Carcassonne, Stone Age and Splendor.

What was his reasoning behind the change?

JakeQZ:
You'll need to change something in the config for this to go through.

oliverklee:
>  Was this F2F/OTB or online? What games? 

This is face-to-face, every few weeks. In that gaming round, we usually play bigger euro games like Brass, Ark Nova or Terraforming Mars.

> What was his reasoning behind the change?

Sebastian's reasoning was that PHPStan is being maintained a lot more actively than Psalm with a lot more contributors. One of the reasons for this is that Vimeo pulled funding for the Psalm project.

- https://github.com/phpstan/phpstan-src/pulse/monthly
- https://github.com/vimeo/psalm/pulse/monthly

oliverklee:
@JakeQZ I’ve found an article about the change in Psalm: https://muglug.medium.com/my-incredible-journey-with-php-df45d72ba2a5

JakeQZ:
> I’ve found an article about the change in Psalm

So, basically, the main person who was working on Psalm at Vimeo moved on in 2021.  Though, at that time, he thought they would find someone to replace him to perform a similar role.  I guess they didn't.  A lot of the heavy lifting PHP used to do server-side (with AJAX requests) is now being pushed client-side, using libraries like React (causing older devices to struggle).

PR #1287, opened 2024-08-27
[TASK] Add extra-strict PHPStan rules

oliverklee:
Because we want our code to be as clean and reliable as possible.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -11,6 +11,11 @@ parameters:
 			path: src/CssInliner.php
 
 		-
+			message: "#^Casting to int something that's already int\\<0, max\\>\\.$#"
+			count: 1
+			path: src/CssInliner.php
+
+		-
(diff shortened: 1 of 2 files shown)


phpstan-baseline.neon
```diff
@@ -10,6 +10,11 @@ parameters:
 			count: 2
 			path: src/CssInliner.php
 
+		-
+			message: "#^Casting to int something that's already int\\<0, max\\>\\.$#"
```
JakeQZ, line 14: Does this mean we'd need to do something like `is_int($var) ? $var : (int)$var`.  I don't like that.  But we can review that section of code and see how it could be improved.

JakeQZ (approved):
More fun :)

phpstan-baseline.neon
oliverklee: No, more like this: #1290

PR #1288, opened 2024-08-27
[TASK] Be less strict about the `@covers` annotation

oliverklee:
We still want to require the `@covers` annotation to be used for testcase (`forceCoversAnnotation="true"`), but we also want to be able to indirectly execute code that is not referenced via an `@covers` annotation.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.6/phpunit.xsd"
-         beStrictAboutCoversAnnotation="true"
          beStrictAboutOutputDuringTests="true"
          beStrictAboutTodoAnnotatedTests="true"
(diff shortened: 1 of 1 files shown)


JakeQZ (approved):
I guess this means classes that happen to be used, but are fully covered, and not intended to be covered.

oliverklee:
> I guess this means classes that happen to be used, but are fully covered, and not intended to be covered.

I don't quite understand this sentence. :-)

I'll try to use my own words:

With `forceCoversAnnotation="true"`, PHPUnit complains if a test calls code that is not referenced via the `@covers` annotation of the corresponding testcase. Removing this configuration option allows this without PHPUnit complaining.

JakeQZ:
> > I guess this means classes that happen to be used, but are fully covered, and not intended to be covered.
> 
> I don't quite understand this sentence. :-)

I think i missed out a "not".  I'll try again.

I guess this means classes that happen to be used, but are not fully covered, and not intended to be covered by this `TestCase`, because they have (or should have) their own dedicated `TestCase`.  E.g. `ATest` `@covers` `A`.  `A` happens to use some functionality from `B`.  But `ATest` does not `@covers` `B`.  `BTest` `@covers` `B`.

PR #1291, opened 2024-08-28
[BUGFIX] Fix the image URL of the CI status badge

oliverklee:
https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/adding-a-workflow-status-badge


JakeQZ (approved):
It seemed to be working as it was, though the old URL may stop working at some point.  This is the now recommended URL, and also works.

PR #1292, opened 2024-08-28
[CLEANUP] Improve a misleading variable name

oliverklee:


JakeQZ (approved):
I guess `$number` was short for `$numberOfMatches`, but is actually no more meaningful than `$n`.

PR #1300, opened 2024-09-05
[TASK] Rein in Dependabot a bit

oliverklee:
We want to stay at PHPUnit 9.x to stay compatible with PHP 7.3.

Also, `rawr/cross-data-providers` has breaking changes in newer releases. As that package has been abandonded anyway, we won't bother with updating it and switch to the successor instead later.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,4 +13,8 @@ updates:
       interval: "daily"
     allow:
       - dependency-type: "development"
+    ignore:
+      - dependency-name: "phpunit/phpunit"
+        versions: [ ">= 10.0.0" ]
+      - dependency-name: "rawr/cross-data-providers"
     versioning-strategy: "increase"


JakeQZ (approved):
Looks like we may need to do this with several dependencies.  Dependabot doesn't seem to have a versioning strategy to increase minor or patch version only.

oliverklee:
> Dependabot doesn't seem to have a versioning strategy to increase minor or patch version only.

It actually does, but it uses the configuration in `.github/dependabot.yml` as a guideline, not the `^` or  `~` in the `composer.json`.

oliverklee:
… and within those limits, it tries to update to the highest version possible.

JakeQZ:
> It actually does, but it uses the configuration in `.github/dependabot.yml` as a guideline, not the `^` or `~` in the `composer.json`.

It seems that you have to manually specify to limit each component with an upper version bound.  So when we want to update to PHPUnit 10, we'd have to change both `composer.json` and `dependabot.yml`.  I was looking for a general setting for [`versioning-strategy`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy) that would cover all tools (though we probably want to allow major updates for some tools anyway).

PR #1307, opened 2024-09-11
[TASK] Add pregReplaceCallback method to handle internal regex errors

JakeQZ:
If an internal PCRE error occurs,
an exception will be thrown or `trigger_error` will be called, depending on the 'debug' setting.
In the latter case,
the original string passed to `preg_replace_callback` will be returned.

Resolves one PHPStan issue.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -51,11 +51,6 @@ parameters:
 			path: src/CssInliner.php
 
 		-
-			message: "#^Parameter \\#2 \\$value of method DOMElement\\:\\:setAttribute\\(\\) expects string, string\\|null given\\.$#"
-			count: 1
-			path: src/CssInliner.php
-
-		-
 			message: "#^Parameter \\#3 \\$subject of function preg_replace expects array\\|string, string\\|null given\\.$#"
 			count: 1
 			path: src/CssInliner.php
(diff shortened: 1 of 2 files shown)


JakeQZ:
PHPStan behaves differently depending on which PHP version and OS it is run on.  Why?!  It's a PITA as it is.

JakeQZ:
The decrease in coverage cannot be fixed here.  It's an untestable private method that's been added.  I give up.

JakeQZ:
Maybe the `preg_` wrapper(s) need moving to a separate class.  But that is beyond the scope of what I'm trying to do at the moment.  I'm just trying to fix PHPStan errors with what I am doing.  I am currently using PHP 7.4 on Windows, which produces a different set of PHPStan errors from those in the list.  Yes, I can set up PHP 8.3 or whatever, but it's a PITA painstakingly going through the config files to make the required changes for it to actually work.

JakeQZ:
> Maybe the `preg_` wrapper(s) need moving to a separate class.

Abandoning this in favour of that.

> PHPStan behaves differently depending on which PHP version and OS it is run on.

Maybe not OS, but definitely PHP version.  It's impossible to continue working with PHP 7.4 so I have now set up 8.3.

PR #1308, opened 2024-09-12
[CLEANUP] Add class for `preg_ function` error handling

JakeQZ:
The new class uses functionality that previously existed in `CssInliner`. The `preg_replace` wrapper is extended to also support arrays as the `$pattern` and `$replacement` parameters, and to support the `$limit` and `$count` parameters.  A `preg_replace_callback` wrapper is also added.

With these methods in a separate class, they can be independently tested, which is duly done.

All calls to `preg_replace` and `preg_replace_callback` are changed to use the new class.  This resolves a number of PHPStan errors.

### DIFF
--- a/tests/Unit/Utilities/PregTest.php
+++ b/tests/Unit/Utilities/PregTest.php
@@ -1,0 +1,219 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\Tests\Unit\Utilities;
+
+use Pelago\Emogrifier\Utilities\Preg;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers \Pelago\Emogrifier\Utilities\Preg
+ */
+final class PregTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function throwExceptionsProvidesFluentInterface(): void
+    {
+        $subject = new Preg();
+
+        $result = $subject->throwExceptions(false);
+
+        self::assertSame($subject, $result);
+    }
+
+    /**
+     * @test
+     */
+    public function exceptionNotThrownByDefault(): void
+    {
+        $this->expectNotToPerformAssertions();
+        $subject = new Preg();
+
+        // Version 6 of PHPUnit allowed setting `PHPUnit\Framework\Error\Notice::$enabled` to `false`
+        // to prevent its error handler throwing exceptions upon `trigger_error`.
+        // This functionality has been removed by version 9.
+        // Version 10 allows use of the #[WithoutErrorHandler] attribute, but this requires PHP 8.0 anyway.
+        // So we must `@` to suppress the triggered error.
+        // This also means we can't test for our own triggered error, since `preg_*` also triggers an error.
+        @$subject->replace('/', '', '');
+    }
+
+    /**
+     * @test
+     */
+    public function exceptionThrownAfterThrowExceptionsTurnedOn(): void
+    {
+        $this->expectException(\RuntimeException::class);
+        $subject = new Preg();
+
+        $subject->throwExceptions(true);
+        @$subject->replace('/', '', '');
+    }
+
+    /**
+     * @test
+     */
+    public function exceptionNotThrownAfterThrowExceptionsTurnedOff(): void
+    {
+        $this->expectNotToPerformAssertions();
+        $subject = new Preg();
+
+        $subject->throwExceptions(true);
+        $subject->throwExceptions(false);
+        @$subject->replace('/', '', '');
+    }
+
+    /**
+     * @return array<array<array-key, string|non-empty-array<string>|int>>
+     */
+    public function providePregReplaceArgumentsAndExpectedResult(): array
+    {
+        return [
+            'string arguments' => [
+                'pattern' => '/a/',
+                'replacement' => 'fab',
+                'subject' => 'abba',
+                'limit' => -1,
+                'expect' => 'fabbbfab',
+            ],
+            'array pattern' => [
+                'pattern' => ['/a/', '/b/'],
+                'replacement' => 'fab',
+                'subject' => 'abba',
+                'limit' => -1,
+                'expect' => 'fafabfabfabfafab',
+            ],
+            'array pattern and replacement' => [
+                'pattern' => ['/a/', '/b/'],
+                'replacement' => ['fab', 'z'],
+                'subject' => 'abba',
+                'limit' => -1,
+                'expect' => 'fazzzfaz',
+            ],
+            'with limit' => [
+                'pattern' => '/a/',
+                'replacement' => 'fab',
+                'subject' => 'abba',
+                'limit' => 1,
+                'expect' => 'fabbba',
+            ],
+        ];
+    }
+
+    /**
+     * @test
+     *
+     * @param non-empty-string|non-empty-array<non-empty-string> $pattern
+     * @param string|non-empty-array<string> $replacement
+     *
+     * @dataProvider providePregReplaceArgumentsAndExpectedResult
+     */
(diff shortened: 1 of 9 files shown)


JakeQZ:
The new class and method names correlate with the built-in PHP functions which are wrapped.  However, if it's deemed that more descriptive names are preferred, they can be changed.

tests/Unit/Utilities/PregTest.php
```diff
@@ -0,0 +1,219 @@
...
+
+        $result = $subject->throwExceptions(false);
+
+        self::assertSame($subject, $result);
+    }
+
+    /**
+     * @test
+     */
+    public function exceptionNotThrownByDefault(): void
```
oliverklee, line 30: Let's make the test name into a full statement:

```suggestion
    public function replaceByDefaultDoesNotThrowAnException(): void
```

tests/Unit/Utilities/PregTest.php
```diff
@@ -0,0 +1,219 @@
...
+    {
+        $this->expectNotToPerformAssertions();
+        $subject = new Preg();
+
+        // Version 6 of PHPUnit allowed setting `PHPUnit\Framework\Error\Notice::$enabled` to `false`
+        // to prevent its error handler throwing exceptions upon `trigger_error`.
+        // This functionality has been removed by version 9.
+        // Version 10 allows use of the #[WithoutErrorHandler] attribute, but this requires PHP 8.0 anyway.
+        // So we must `@` to suppress the triggered error.
+        // This also means we can't test for our own triggered error, since `preg_*` also triggers an error.
```
oliverklee, line 40: I really love this explanation!

tests/Unit/Utilities/PregTest.php
```diff
@@ -0,0 +1,219 @@
...
+        @$subject->replace('/', '', '');
+    }
+
+    /**
+     * @test
+     */
+    public function exceptionThrownAfterThrowExceptionsTurnedOn(): void
```
oliverklee, line 47:
```suggestion
    public function exceptionIsThrownAfterThrowExceptionsTurnedOn(): void
```

tests/Unit/Utilities/PregTest.php
```diff
@@ -0,0 +1,219 @@
...
+        $subject = new Preg();
+
+        $subject->throwExceptions(true);
+        @$subject->replace('/', '', '');
+    }
+
+    /**
+     * @test
+     */
+    public function exceptionNotThrownAfterThrowExceptionsTurnedOff(): void
```
oliverklee, line 59:
```suggestion
    public function exceptionIsNotThrownAfterThrowExceptionsTurnedOff(): void
```

tests/Unit/Utilities/PregTest.php
```diff
@@ -0,0 +1,219 @@
...
+        $this->expectNotToPerformAssertions();
+        $subject = new Preg();
+
+        $subject->throwExceptions(true);
+        $subject->throwExceptions(false);
+        @$subject->replace('/', '', '');
+    }
+
+    /**
+     * @return array<array<array-key, string|non-empty-array<string>|int>>
```
oliverklee, line 70:
```suggestion
     * @return array<non-empty-string, array{pattern: string|list<non-empty-string>, replacement: non-empty-string|list<non-empty-string>, subject: non-empty-string , limit: int, expect: non-empty-string}>
```

tests/Unit/Utilities/PregTest.php
```diff
@@ -0,0 +1,219 @@
...
+     * @dataProvider providePregReplaceArgumentsAndExpectedResult
+     */
+    public function replaceCallbackReplaces(
+        $pattern,
+        $replacement,
+        string $subject,
+        int $limit,
+        string $expectedResult
+    ): void {
+        $callback = static function (array $matches) use ($replacement): string {
```
oliverklee, line 138: I'd strongly prefer this static function to be a class method in order to make it easier to grok. Then we could also use class properties instead of static local variables.

oliverklee (changes requested):
Love this! I've suggested some renamings and one refactoring.

tests/Unit/Utilities/PregTest.php
JakeQZ: I've not renamed this to explicitly state the method that will be tested for not throwing an exception.  I realized that these tests need to cover that the exceptions on/off switch applies to both methods.  So have extended the tests thusly.

tests/Unit/Utilities/PregTest.php
JakeQZ: Done.

tests/Unit/Utilities/PregTest.php
JakeQZ: Done.

tests/Unit/Utilities/PregTest.php
JakeQZ: I did consider this, but found that the array keys have to be explicitly stated in the type, whereas we only care about the types of the values, in order.  However, I'm fine either way so have changed it (which does mean the method must now use the specific key names).

tests/Unit/Utilities/PregTest.php
JakeQZ: Done.

tests/Unit/Utilities/PregTest.php
```diff
@@ -0,0 +1,282 @@
...
+ */
+final class PregTest extends TestCase
+{
+    /**
+     * @var string|array<string>
+     */
+    private $replaceCallbackReplacement;
+
+    /**
+     * @var ?string
```
oliverklee, line 21: The `?` notation only works for native type declarations. For PHPDoc type annotations, we need to use `|null`:

```suggestion
     * @var string|null
```

tests/Unit/Utilities/PregTest.php
```diff
@@ -0,0 +1,282 @@
...
+     */
+    private $lastReplaceCallbackMatch;
+
+    /**
+     * @var int
+     */
+    private $replaceCallbackReplacementIndex;
```
oliverklee, line 28: As this is non-nullable, let's initialize it with an int:

```suggestion
    private $replaceCallbackReplacementIndex = 0;
```

tests/Unit/Utilities/PregTest.php
```diff
@@ -0,0 +1,282 @@
...
+        string $expectedResult
+    ): void {
+        $this->replaceCallbackReplacement = $replacement;
+        $this->lastReplaceCallbackMatch = null;
+        $this->replaceCallbackReplacementIndex = -1;
+        $testSubject = new Preg();
+
+        $result = $testSubject->replaceCallback(
+            $pattern,
+            \Closure::fromCallable([$this, 'callbackForReplaceCallback']),
```
oliverklee, line 197: Isn't the array also a callable, e.g. like this, so that we don't need the `Closure::fromCallable`?

`[$this, 'callbackForReplaceCallback']`

oliverklee (approved):

tests/Unit/Utilities/PregTest.php
JakeQZ: The `?` notation works with PHPStan - without it, it generates an error when the property is assigned to `null`.

tests/Unit/Utilities/PregTest.php
JakeQZ: It's not accessible to the method `replaceCallback` as it is not `public`, resulting in "[Argument passed to function must be callable, array given](https://stackoverflow.com/questions/43170785/argument-passed-to-function-must-be-callable-array-given)".  (The built in function `preg_replace_callback` circumvents the restriction in some internal way.)

tests/Unit/Utilities/PregTest.php
JakeQZ: Done.

Issue #1276, opened 2024-08-08
Support to convert CSS Variables?


njt1982:
Hi,

Is it already possible to convert CSS variables into inline values?

For example...

- The HTML email being sent needs per-site brand colours used (eg button background).
- Have the button with a `bg-brand` class on it.
- I have some CSS text being dynamically generated
- The CSS currently contains something like:

```
body {
  --brand: #ff0000;
}
.bg-brand {
  background: var(--brand);
}
.link-color {
  color: var(--brand);
}
td.container {
  border-color: var(--brand);
}
```

(this is a simplified example!).

For now, I'm going to change my code to replace `var(--brand)` with something like `[BRAND-COLOR]` and then string replace that with the hex code...

I'm only asking as I already have CSS elsewhere that defines these brand colour variables and it would be nice to reuse that rather than repeat the same variables in different languages in different places :)

(PS Thank for all the work on this package!)

oliverklee:
Hi! No, at the moment, this library does not convert CSS variables into inline values (yet). I’m also not sure this should happen at the parser level, though, or whether this should be the responsibility of the code taking the syntax tree (the result from parsing the CSS).

@sabberworm @JakeQZ What are your thoughts on this?

JakeQZ:
The evaluation of CSS variables is based on the DOM tree.  The value is taken from self-or-closest-ancestor where the variable is effectively a property defined according to normal CSS precedence rules.

So I think to support this, we need
- An initial Emogrification pass to determine the variables specifically assigned at each element node, and some way of storing the results so they can be looked up by element (using the XPath path is risky, since it changes upon manipulation of the DOM tree - I think we would need to set a custom ID property on each DOM element);
- In the regular Emogrification pass, whenever a property value with a CSS variable is encountered, we traverse the DOM tree to find its value from the previous step;
- Support for `:root` (commonly used for variable definitions, though is just another way of matching the `html` element).

Then there's non-inlinable CSS (like `@media` and `:hover`).  The best we can do for that is to include all the variable definitions that might apply (to begin with we can include all, then, as an enhancement, filter out those that aren't needed).

Some refactoring might not go amiss first, though I'm not sure where or how.  Alternatively, we go ahead, and review later where there seems to be common duplicated functionality.

JakeQZ:
There will be caveats if we do this.  E.g.

```css
:root {
  --font-size: 16px;
}
@media (max-width: 640px) {
  :root {
    --font-size: 14px;
  }
[...]

PR #1311, opened 2024-09-14
[CLEANUP] Add DeclarationBlockParser class

JakeQZ:
This is essentially a refactoring to move the `parseCssDeclarationsBlock` method that was duplicated in both `CssInliner` and `CssToAttributeConverter` to a separate class.

As well as eliminating duplicate code/functionality, there are other benefits:
- The method can be tested independently, which is duly done;
- The results cache is now shared;
- The functionality is readily available to other classes, e.g. for the solution to #1276 (as currently proposed).

PHPStan errors from the original code have been resolved as follows:
- Use `Preg::split` to wrap `preg_split` to handle unexpected errors;
- Check the return value from `preg_match` precisely with `!== 1` instead of `!`.
  
Longer-term, we should be able to use functionality from PHP-CSS-Parser for more accurate parsing (e.g. strings containing semicolons), whereupon this class can become a wrapper for that, but currently that functionality is not available in a standalone class.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6,18 +6,13 @@ parameters:
 			path: src/CssInliner.php
 
 		-
-			message: "#^Argument of an invalid type array\\<int, string\\>\\|false supplied for foreach, only iterables are supported\\.$#"
-			count: 1
-			path: src/CssInliner.php
-
-		-
 			message: "#^Method Pelago\\\\Emogrifier\\\\CssInliner\\:\\:getAllNodesWithStyleAttribute\\(\\) return type with generic class DOMNodeList does not specify its types\\: TNode$#"
 			count: 1
 			path: src/CssInliner.php
 
 		-
 			message: "#^Only booleans are allowed in a negated boolean, int\\|false given\\.$#"
-			count: 3
+			count: 2
 			path: src/CssInliner.php
 
 		-
@@ -41,11 +36,6 @@ parameters:
 			path: src/HtmlProcessor/AbstractHtmlProcessor.php
 
 		-
-			message: "#^Argument of an invalid type array\\<int, string\\>\\|false supplied for foreach, only iterables are supported\\.$#"
-			count: 1
-			path: src/HtmlProcessor/CssToAttributeConverter.php
-
-		-
 			message: "#^Method Pelago\\\\Emogrifier\\\\HtmlProcessor\\\\CssToAttributeConverter\\:\\:getAllNodesWithStyleAttribute\\(\\) return type with generic class DOMNodeList does not specify its types\\: TNode$#"
 			count: 1
 			path: src/HtmlProcessor/CssToAttributeConverter.php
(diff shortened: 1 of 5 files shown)


JakeQZ:
I can only think that coverage has gone down because the code path in which `preg_split` is never followed.

However, it is probably impossible to pass a declaration block that would cause it to fail (and would depend on PHP environment configuration), because the regex is fairly simple.

Maybe I should add a `Preg::split` wrapper before putting this in...

JakeQZ:
> Maybe I should add a `Preg::split` wrapper before putting this in...

This has eliminated an unexercised statement.  I also realized that the cache functionality wasn't being tested (I had previously thought to put in a test for that, then forgot).

Now I am pretty sure all statements/constructs in the new class are covered by the test.

But still coverage is down 0.02%.

Note that there is a conditional `?:` in which the RHS is never evaluated.  But even without that (and allowing the PHPStan error instead), coverage is still down by the same amount.

I think the issue with the Coveralls test is that it is too simplistic.  It just looks at the percentage of code units (statements?) that are covered.  Here, I've replaced two completely-covered methods with one completely-covered method.  Thus the amount of uncovered code, as a percentage, will inevitably increase.  The metric that should be used for CI failure should be the actual amount of uncovered code, not the percentage of it as a whole.  I would expect similar problems whenever any functionality is removed.

Unless you can spot any peice of code in the new class that is not covered by tests, I suggest that the Coveralls CI failure be overridden in this case.  And we should see if there is an alternative metric that can be used for CI failure (as suggested above), and if not create a ticket.

Obviously review the code as usual :)

JakeQZ:
Changing to draft again, as now dependent on #1313.

Though the coverage metrics issue described above is still active - @oliverklee?

oliverklee:
> Unless you can spot any peice of code in the new class that is not covered by tests, I suggest that the Coveralls CI failure be overridden in this case.

Yes, I agree.

I honestly don't know how the `coverage/coveralls` ended up twice in the list of actions. For my other projects, I only have the first of the two checks.

oliverklee:
> I honestly don't know how the `coverage/coveralls` ended up twice in the list of actions. For my other projects, I only have the first of the two checks.

I've now checked the "leave PR comments" checkbox (which, for some reason) was not checked in the Coveralls configuration. Let's see.

oliverklee:
I've also removed access to the Coveralls Pro 3rd party app now.

JakeQZ:
This can be reviewed now.

JakeQZ:
> I've now checked the "leave PR comments" checkbox (which, for some reason) was not checked in the Coveralls configuration. Let's see.

Seems to have worked.  However, there was no comment made when I made the final repush to this PR branch.  That may have been because the PR was still marked as draft at the time (and chaning PR status from 'draft' to 'ready for review' does not trigger the comment).

oliverklee:
> That may have been because the PR was still marked as draft at the time (and chaning PR status from 'draft' to 'ready for review' does not trigger the comment).

Yes, that's what I think as well. (The CI for PRs only runs after a push.)

oliverklee (approved):
LGTM. I like it!

PR #1312, opened 2024-09-15
[TASK] Add Preg::split to wrap preg_split with error handling

JakeQZ:


JakeQZ:
This is a prerequisite for #1311 so that that can be completed without decreasing code coverage.

I have not updated any existing code using `preg_split` to use the new method.  That can be done as a separate PR.  In particular, I don't want to update `CssInliner::parseDeclarationsBlock` which will be removed by #1311.  This currently ignores the `$debug` setting, so I don't want to make a change that would apply that via `Preg::throwExceptions()`, which would then require #1311 to add a similar setting to the class extracted there - which I think is unnecessary for now.  (It can always be added later as a separate PR.)

(Do I get the prize for the longest test method name? ;))

oliverklee (approved):
Neat!

PR #1313, opened 2024-09-16
[TASK] Remove support for PREG_SPLIT_OFFSET_CAPTURE from Preg::split

JakeQZ:
This flag is not used anywhere in the current code base.  It causes complications with static analysis due to its introduction of a dynamic return type for `preg_split`, which would require a PHPStan extension to properly solve (and which assumes PHPStan would always be used as the static analysis tool).

### DIFF
--- a/src/Utilities/Preg.php
+++ b/src/Utilities/Preg.php
@@ -101,7 +101,7 @@ final class Preg
     public function split(string $pattern, string $subject, int $limit = -1, int $flags = 0): array
     {
         if (($flags & PREG_SPLIT_OFFSET_CAPTURE) !== 0) {
-            throw new \RuntimeException('PREG_SPLIT_OFFSET_CAPTURE is not supported by Preg::split');
+            throw new \RuntimeException('PREG_SPLIT_OFFSET_CAPTURE is not supported by Preg::split', 1726506348);
         }
 
         $result = \preg_split($pattern, $subject, $limit, $flags);
--- a/src/Utilities/Preg.php
+++ b/src/Utilities/Preg.php
@@ -89,28 +89,26 @@ final class Preg
     /**
      * Wraps `preg_split`.
      * If an error occurs, and exceptions are not being thrown,
-     * a single-element array containing the original `$subject` is returned -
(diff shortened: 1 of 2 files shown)


JakeQZ:
> a PHPStan extension

Or perhaps [conditional return types](https://phpstan.org/writing-php-code/phpdoc-types#conditional-return-types) and finding some proper documentation about them.

src/Utilities/Preg.php
```diff
@@ -89,28 +89,26 @@ public function replaceCallback(
...
      *
-     * @return array<int, string>|array<int, array{0: string, 1: int}>
+     * @return array<int, string>
      *
      * @throws \RuntimeException
      */
     public function split(string $pattern, string $subject, int $limit = -1, int $flags = 0): array
     {
+        if (($flags & PREG_SPLIT_OFFSET_CAPTURE) !== 0) {
+            throw new \RuntimeException('PREG_SPLIT_OFFSET_CAPTURE is not supported by Preg::split');
```
oliverklee, line 104: Please let's always provide an exception code (the current timestamp). This allows us to search the code for an exception code that is displayed.

```suggestion
            throw new \RuntimeException('PREG_SPLIT_OFFSET_CAPTURE is not supported by Preg::split', 1726473096);
```

tests/Unit/Utilities/PregTest.php
```diff
@@ -346,13 +338,12 @@ public function splitReturnsArrayContainingSubjectOnError(): void
     /**
      * @test
      */
-    public function splitWithOffsetCaptureReturnsArrayContainingArrayWithSubjectAndZeroOffsetOnError(): void
+    public function splitWithOffsetCaptureIsNotSupported(): void
     {
+        $this->expectException(\RuntimeException::class);
```
oliverklee, line 343: Please let's also test for the exception message and the code.

src/Utilities/Preg.php
JakeQZ: Done..

tests/Unit/Utilities/PregTest.php
JakeQZ: Done.

PR #1314, opened 2024-09-16
[CLEANUP] Use Preg::split to handle unexpected preg_split errors

JakeQZ:
This resolves three PHPStan errors.

Instances of calls to `preg_split` in the two (identical) `parseCssDeclarationsBlock` methods are not updated - those are covered separately by #1311.


JakeQZ:
Might have merge conflicts with #1311 in `phpstan-baseline.neon`, so this or that PR might need a rebase after the other is merged to `main` - we'll see.

oliverklee (approved):

PR #1315, opened 2024-09-16
[CLEANUP] Remove an impossible `throw`

JakeQZ:
PHPStan is able to determine that `array_combine` when passed the results of `array_keys` and `array_values` for the same array can never fail (in the sense of returning `false`), because the two passed arrays will always be the same length.

Psalm could not, which was why the return value check and `throw` was originally put in.


JakeQZ:
On the basis of this, and a few other things, PHPStan impresses me much.  It keeps me warm in the middle of the night.  I wonder if [Shania Twain](https://www.youtube.com/watch?v=mqFLXayD6e8) agrees.

oliverklee:
@JakeQZ And it warms my heart to see you so motivated by increasing code coverage and killing PHPStan warnings! :heart:

(It's the same for me. I find PHPStan quite addictive.)

oliverklee (approved):

PR #1316, opened 2024-09-17
[CLEANUP] Add `CssInliner::querySelectorAll` method

JakeQZ:
This `private` method provides functionality that is used in three separate places (albeit a one-liner), and includes taking care of exception handling according to the debug mode setting.

The new method and parameter names match the equivalent in the [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll).

There will be a tiny performance impact due to additional method calls.  But the benefits include code re-use and common error handling.

Also add `ensureNodeIsElement` method for use with the result from `querySelectorAll` to ensure type safety.  `querySelectorAll` should always return a `DOMNodeList` of `DOMElement`s.  But this cannot be guaranteed if there is a bug in CssSelector whereby it returns an XPath expression which may match non-element nodes.  And it would (probably) be sub-optimal for `querySelectorAll` to have an extra loop checking this, when it can be done in the caller's loop over the returned data.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -533,18 +533,20 @@ class CssInliner extends AbstractHtmlProcessor
             }
 
             return $result;
-        } catch (ParseException|\RuntimeException $exception) {
+        } catch (ParseException $exception) {
+            $alwaysThrowParseException = $options[self::QSA_ALWAYS_THROW_PARSE_EXCEPTION] ?? false;
+            if ($this->debug || $alwaysThrowParseException) {
+                throw $exception;
+            }
+            return new \DOMNodeList();
+        } catch (\RuntimeException $exception) {
             if (
                 $this->debug
-                || ($options[self::QSA_ALWAYS_THROW_PARSE_EXCEPTION] ?? false) && $exception instanceof ParseException
             ) {
                 throw $exception;
             }
-            // In non-debug mode, `ParseException`s are silently ignored, whereas `RuntimeException`s (indicating a bug
-            // in CssSelector) have their message passed to the error handler (for logging or custom handling).
-            if ($exception instanceof \RuntimeException) {
-                \trigger_error($exception->getMessage());
-            }
+            // `RuntimeException` indicates a bug in CssSelector so pass the message to the error handler.
+            \trigger_error($exception->getMessage());
             return new \DOMNodeList();
         }
     }
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -54,6 +54,13 @@ class CssInliner extends AbstractHtmlProcessor
     private const COMBINATOR_MATCHER = '(?:\\s++|\\s*+[>+~]\\s*+)(?=[[:alpha:]_\\-.#*:\\[])';
 
     /**
+     * options array key for `querySelectorAll`
+     *
+     * @var string
+     */
+    private const QSA_ALWAYS_THROW_PARSE_EXCEPTION = 'alwaysThrowParseException';
+
+    /**
      * @var array<string, bool>
      */
     private $excludedSelectors = [];
@@ -164,7 +171,11 @@ class CssInliner extends AbstractHtmlProcessor
      * @return $this
      *
      * @throws ParseException in debug mode, if an invalid selector is encountered
-     * @throws \RuntimeException in debug mode, if an internal PCRE error occurs
+     * @throws \RuntimeException
+     *         in debug mode, if an internal PCRE error occurs
+     *         or `CssSelectorConverter::toXPath` returns an invalid XPath expression
+     * @throws \UnexpectedValueException
+     *         if a selector query result includes a node which is not a `DOMElement`
      */
     public function inlineCss(string $css = ''): self
     {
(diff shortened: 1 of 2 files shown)


JakeQZ:
> [Coverage Status](https://coveralls.io/builds/69830368)
> 
> coverage: 96.688% (-0.6%) from 97.311% when pulling **[5c03b33](https://github.com/MyIntervals/emogrifier/commit/5c03b337b50148c69d58cd6c58f968164a483159) on cleanup/query-selector-all** into **[fe0bb4d](https://github.com/MyIntervals/emogrifier/commit/fe0bb4d99ac5ff596807d7f6263cca039bc29150) on main**.

> [If the expression is malformed or the contextNode is invalid, DOMXPath::query() returns false.](https://www.php.net/manual/en/domxpath.query.php#refsect1-domxpath.query-returnvalues)

That will never happen unless there is a bug in `CssSelectorConverter`.  We are not using the `$contextNode` parameter.  So the error handling code path cannot be exercised in tests (for coverage or otherwise).

JakeQZ:
Can improve this.  One instance that could call the new method is missed.  And excpetion handling logic (and/or its definition) can be made more precise.

JakeQZ:
Will be ready for review when I repush.  I'm marking as ready for review now, to see if we get a Coveralls comment on repush when it's not in draft status.

oliverklee:
> Oh. I've just noticed that Coveralls updates its comment, instead of providing a new one.

Yeah. I had the same facial expression some time ago when I noticed the same thing … ;-)

src/CssInliner.php
```diff
@@ -496,34 +496,72 @@ private function getCssFromAllStyleNodes(): string
...
+     *        This is an array of option values to control behaviour:
+     *        - `QSA_ALWAYS_THROW_PARSE_EXCEPTION` - `bool` - throw any `ParseException` regardless of debug setting.
+     *
+     * @return \DOMNodeList<\DOMNode> the HTML elements that match the provided CSS `$selectors`
+     *
+     * @throws ParseException
+     *         in debug mode (or with `QSA_ALWAYS_THROW_PARSE_EXCEPTION` option), if an invalid selector is encountered
+     * @throws \RuntimeException in debug mode, if `CssSelectorConverter::toXPath` returns an invalid XPath expression
+     */
+    private function querySelectorAll(string $selectors, array $options = []): \DOMNodeList
```
oliverklee, line 526: Can/should we add an `@param non-empty-string $selectors` annotation?

src/CssInliner.php
```diff
@@ -496,34 +496,72 @@ private function getCssFromAllStyleNodes(): string
...
+        } catch (ParseException|\RuntimeException $exception) {
+            if (
+                $this->debug
+                || ($options[self::QSA_ALWAYS_THROW_PARSE_EXCEPTION] ?? false) && $exception instanceof ParseException
+            ) {
+                throw $exception;
+            }
+            // In non-debug mode, `ParseException`s are silently ignored, whereas `RuntimeException`s (indicating a bug
+            // in CssSelector) have their message passed to the error handler (for logging or custom handling).
+            if ($exception instanceof \RuntimeException) {
```
oliverklee, line 545: I find this construct hard to grok. Maybe we can use multiple `catch`es instead to make the code flow more obvious to the reader.

src/CssInliner.php
```diff
@@ -496,34 +496,72 @@ private function getCssFromAllStyleNodes(): string
...
+
+            if ($result === false) {
+                throw new \RuntimeException('query failed with selector \'' . $selectors . '\'', 1726533051);
+            }
+
+            return $result;
+        } catch (ParseException|\RuntimeException $exception) {
+            if (
+                $this->debug
+                || ($options[self::QSA_ALWAYS_THROW_PARSE_EXCEPTION] ?? false) && $exception instanceof ParseException
```
oliverklee, line 539: I propose we add at least one local, well-named variable to make this easier to understand.

src/CssInliner.php
```diff
@@ -496,34 +496,72 @@ private function getCssFromAllStyleNodes(): string
...
+                \trigger_error($exception->getMessage());
+            }
+            return new \DOMNodeList();
+        }
+    }
+
+    /**
+     * @throws \UnexpectedValueException if `$node` is not a `DOMElement`
+     */
+    private function ensureNodeIsElement(\DOMNode $node): \DOMElement
```
oliverklee, line 555: We can add this line to make this obvious to PHPStan:

```
@phpstan-assert \DomElement $node
```

Then we don't need to return `$node.

https://phpstan.org/writing-php-code/narrowing-types#custom-type-checking-functions-and-methods

oliverklee: I'm not really sure, though, whether that would really improve this piece of code. What do you think?

oliverklee (commented):
I really like this refactoring! :heart:

src/CssInliner.php
JakeQZ: Not without changing various code throughout the class to enfore that selectors are always non-empty strings, which I think is beyond the scope of this PR.

oliverklee (approved):

src/CssInliner.php
JakeQZ: Have done this.  (When I originally wrote this, I hadn't added the option to always throw `ParseExceptions`, so it was simpler, but one of the callers needs to handle that exception itself.)

src/CssInliner.php
JakeQZ: Done.

src/CssInliner.php
JakeQZ: The return value is needed for 'fluent' calls like `doSomethingWithElement($this->ensureNodeIsElement($node))`, without needing a separate statement.  Also, I'm not sure that PHPStan can see inside the method when analyzing code like the following:

```php
private function doSomething(\DOMNode $node)
{
    $this->ensureNodeIsElement($node);
    doSomethingWithElement($node);
}
```

PR #1319, opened 2024-09-19
[TASK] Add `Preg::match` to wrap `preg_match` with error handling

JakeQZ:
This will allow code using `preg_match` to be written more cleanly.

### DIFF
--- a/src/Utilities/Preg.php
+++ b/src/Utilities/Preg.php
@@ -114,6 +114,31 @@ final class Preg
         return $result;
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


JakeQZ:
I'll update the code to use it as a separate PR.

I also think the methods in `PregTest` should be ordered by method tested (apart from the tests for throwing or not throwing exceptions), rather than having all the tests for failure fallback at the end.  Again, as a separate PR.

oliverklee (approved):
Neat!

PR #1323, opened 2024-09-19
[DOCS] Add an API and deprecation policy

oliverklee:
This is exactly the same change as in our sister project, PHP-CSS-Parser:

https://github.com/MyIntervals/PHP-CSS-Parser/pull/720


JakeQZ (approved):
Matches word for word.

PR #1327, opened 2024-09-19
[CLEANUP] Reorder methods in PregTest

JakeQZ:
They are now grouped by the method tested, apart from tests for the exception switch, which cover all methods at once.

### DIFF
--- a/tests/Unit/Utilities/PregTest.php
+++ b/tests/Unit/Utilities/PregTest.php
@@ -184,6 +184,46 @@ final class PregTest extends TestCase
 
     /**
      * @test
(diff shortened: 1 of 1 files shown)


JakeQZ:
@oliverklee, I recall a discussion about method ordering (though not where on when), in which we kind-of agreed:
- methods should be grouped by functionality, regardless of whether they are `private` or `public`;
- supporting methods (such as data providers and callbacks, in this case) should be presented before the higher-level method that uses them.

I've followed that principle here.

oliverklee (approved):

PR #1332, opened 2024-09-22
[BUGFIX] Preserve case of CSS variable names

JakeQZ:
Unlike standard CSS properties, custom properties have case-sensitive names.

Relates to #1276.


oliverklee (approved):
Looks good to me!

PR #1336, opened 2024-09-23
[FEATURE] Support CSS custom properties

### LINKED ISSUE #1276: Support to convert CSS Variables?

JakeQZ:
Add a new `HtmlProcessor` class that can evaluate CSS custom properties (variables) after CSS which uses them has been inlined.

Resolves #1276, though an additional feature to remove unused variable definitions after evaluation could be provided.

### DIFF
--- a/src/HtmlProcessor/CssVariableEvaluator.php
+++ b/src/HtmlProcessor/CssVariableEvaluator.php
@@ -186,8 +186,10 @@ class CssVariableEvaluator extends AbstractHtmlProcessor
             $variableDefinitions = $ancestorVariableDefinitions;
         }
 
-        for ($child = $element->firstElementChild; $child !== null; $child = $child->nextElementSibling) {
-            $this->evaluateVaraiblesInElementAndDescendants($child, $variableDefinitions);
+        foreach ($element->childNodes as $child) {
+            if ($child instanceof \DOMElement) {
+                $this->evaluateVaraiblesInElementAndDescendants($child, $variableDefinitions);
+            }
         }
 
         return $this;
--- a/src/HtmlProcessor/CssVariableEvaluator.php
+++ b/src/HtmlProcessor/CssVariableEvaluator.php
@@ -1,0 +1,195 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pelago\Emogrifier\HtmlProcessor;
+
+use Pelago\Emogrifier\Utilities\DeclarationBlockParser;
+use Pelago\Emogrifier\Utilities\Preg;
+
+/**
+ * This class can evaluate CSS custom properties that are defined and used in inline style attributes.
+ */
+class CssVariableEvaluator extends AbstractHtmlProcessor
+{
+    /**
+     * temporary collection used by {@see replaceVariablesInDeclarations} and callee methods
+     *
+     * @var array<non-empty-string, string>
+     */
+    private $currentVariableDefinitions = [];
+
+    /**
+     * Replaces all CSS custom property references in inline style attributes with their corresponding values where
+     * defined in inline style attributes (either from the element itself or the nearest ancestor).
+     *
+     * @throws \UnexpectedValueException
+     *
+     * @return $this
+     */
+    public function evaluateVariables(): self
+    {
+        return $this->evaluateVaraiblesInElementAndDescendants($this->getHtmlElement(), []);
+    }
+
+    /**
+     * @param array<non-empty-string, string> $declarations
+     *
+     * @return array<non-empty-string, string>
+     */
+    private function getVaraibleDefinitionsFromDeclarations(array $declarations): array
+    {
+        return \array_filter(
+            $declarations,
+            static function (string $key): bool {
+                return \substr($key, 0, 2) === '--';
+            },
+            ARRAY_FILTER_USE_KEY
+        );
+    }
+
+    /**
+     * Callback function for {@see replaceVariablesInPropertyValue} performing regular expression replacement.
+     *
+     * @param array<int, string> $matches
+     */
+    private function getPropertyValueReplacement(array $matches): string
+    {
+        $variableName = $matches[1];
+        if (isset($this->currentVariableDefinitions[$variableName])) {
+            return $this->currentVariableDefinitions[$variableName];
+        } else {
+            $fallbackValueSeparator = $matches[2] ?? '';
+            if ($fallbackValueSeparator !== '') {
+                $fallbackValue = $matches[3];
+                // The fallback value may use other CSS variables, so recurse
+                return $this->replaceVariablesInPropertyValue($fallbackValue);
+            } else {
+                return $matches[0];
+            }
+        }
+    }
+
+    /**
+     * Regular expression based on {@see https://stackoverflow.com/a/54143883/2511031 a StackOverflow answer}.
+     */
+    private function replaceVariablesInPropertyValue(string $propertyValue): string
+    {
+        return (new Preg())->replaceCallback(
+            '/
+                var\\(
+                    \\s*+
+                    # capture variable name including `--` prefix
+                    (
+                        --[^\\s\\),]++
+                    )
+                    \\s*+
+                    # capture optional fallback value
+                    (?:
+                        # capture separator to confirm there is a fallback value
+                        (,)\\s*
+                        # begin capture with named group that can be used recursively
+                        (?<recursable>
+                            # begin named group to match sequence without parentheses, except in strings
+                            (?<noparentheses>
+                                # repeated zero or more times:
+                                (?:
+                                    # sequence without parentheses or quotes
+                                    [^\\(\\)\'"]++
+                                    |
+                                    # string in double quotes
+                                    "(?>[^"\\\\]++|\\\\.)*"
+                                    |
+                                    # string in single quotes
+                                    \'(?>[^\'\\\\]++|\\\\.)*\'
+                                )*+
+                            )
+                            # repeated zero or more times:
+                            (?:
+                                # sequence in parentheses
+                                \\(
+                                    # using the named recursable pattern
+                                    (?&recursable)
+                                \\)
+                                # sequence without parentheses, except in strings
+                                (?&noparentheses)
+                            )*+
+                        )
+                    )?+
+                \\)
+            /x',
+            \Closure::fromCallable([$this, 'getPropertyValueReplacement']),
+            $propertyValue
+        );
+    }
+
+    /**
+     * @param array<non-empty-string, string> $declarations
+     *
+     * @return array<non-empty-string, string>|false `false` is returned if no substitutions were made.
+     */
+    private function replaceVariablesInDeclarations(array $declarations)
+    {
+        $substitutionsMade = false;
+        $result = \array_map(
+            function (string $propertyValue) use (&$substitutionsMade): string {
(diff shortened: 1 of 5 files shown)


JakeQZ:
This uses some `DOMElement` properties not available in PHP 7, so will need some changes there.

JakeQZ:
> This uses some `DOMElement` properties not available in PHP 7, so will need some changes there.

Now fixed (as a separate commit, so it can still be seen what the PHP 8 code would look like).

src/HtmlProcessor/CssVariableEvaluator.php
```diff
@@ -0,0 +1,197 @@
...
+            /x',
+            \Closure::fromCallable([$this, 'getPropertyValueReplacement']),
+            $propertyValue
+        );
+    }
+
+    /**
+     * @param array<non-empty-string, string> $declarations
+     *
+     * @return array<non-empty-string, string>|false `false` is returned if no substitutions were made.
```
oliverklee, line 129: Would it make sense to return `null` instead of `false`? Then we can make this a nullable type instead of a union type.

src/HtmlProcessor/CssVariableEvaluator.php
```diff
@@ -0,0 +1,197 @@
...
+
+        return \implode('; ', $declarationStrings) . ';';
+    }
+
+    /**
+     * @param array<non-empty-string, string> $ancestorVariableDefinitions
+     *
+     * @return $this
+     */
+    private function evaluateVaraiblesInElementAndDescendants(
```
oliverklee, line 169: Typo: `Varaibles` -> `Variables`

tests/Unit/HtmlProcessor/CssVariableEvaluatorTest.php
```diff
@@ -0,0 +1,394 @@
...
+                    }
+                    p {
+                        --text-color: green;
+                    }
+                    p {
+                        color: var(--text-color);
+                    }
+                ',
+                // The variable definition is not removed, but its value should nonetheless be applied
+                'expect' => '<p style="--text-color: green; color: green;">',
```
oliverklee, line 132: This example tends to boggle my mind … is this the way CSS custom properties are expected to work?

tests/Unit/HtmlProcessor/CssVariableEvaluatorTest.php
```diff
@@ -0,0 +1,394 @@
...
+                    body {
+                        width: var(--page-width, calc(100vw - 2 * calc(1rem + 10px)));
+                    }
+                ',
+                'expect' => '<body style="width: calc(100vw - 2 * calc(1rem + 10px));">',
+            ],
+            // Processing through `CssInliner` may result in the quotes being changed.
+            // A direct test with HTML is also performed for the following four scenarios.
+            'fallback value with single-quoted string containing opening parenthesis' => [
+                'css' => '
```
oliverklee, line 276: I propose we switch to double quotes for those strings in PHP that contain (only) single quotes. Then we don't need to escape the single quotes.

tests/Unit/HtmlProcessor/CssVariableEvaluatorTest.php
```diff
@@ -0,0 +1,394 @@
...
+            ],
+            'fallback value with double-quoted string containing opening parenthesis' => [
+                'css' => '
+                    h1 {
+                        content: var(--main-heading, "Missing heading :(");
+                    }
+                ',
+                'expect' => '<h1 style=\'content: "Missing heading :(";\'>',
+            ],
+            'fallback value with double-quoted string containing closing parenthesis' => [
```
oliverklee, line 299: I really love this thorough set of test cases! :heart:

tests/Unit/HtmlProcessor/CssVariableEvaluatorTest.php
```diff
@@ -0,0 +1,394 @@
...
+
+        $htmlResult = $subject->evaluateVariables()->render();
+
+        self::assertStringContainsString($expectedHtmlFragment, $htmlResult);
+    }
+
+    /**
+     * @return array<non-empty-string, array{html: non-empty-string, expect: non-empty-string}>
+     */
+    public function provideHtmlUsingVaraiblesAndExpectedHtmlFragmentAfterInliningAndEvaluation(): array
```
oliverklee, line 333: Varaibles -> Variables (maybe do a global search'n'replace)

src/HtmlProcessor/CssVariableEvaluator.php
```diff
@@ -0,0 +1,197 @@
...
+     * Callback function for {@see replaceVariablesInPropertyValue} performing regular expression replacement.
+     *
+     * @param array<int, string> $matches
+     */
+    private function getPropertyValueReplacement(array $matches): string
+    {
+        $variableName = $matches[1];
+        if (isset($this->currentVariableDefinitions[$variableName])) {
+            return $this->currentVariableDefinitions[$variableName];
+        } else {
```
oliverklee, line 61: As we have a `return` in the guard clause, we don't need the `else`, which will allow us to lose one level of indentation.

src/HtmlProcessor/CssVariableEvaluator.php
```diff
@@ -0,0 +1,197 @@
...
+            $fallbackValueSeparator = $matches[2] ?? '';
+            if ($fallbackValueSeparator !== '') {
+                $fallbackValue = $matches[3];
+                // The fallback value may use other CSS variables, so recurse
+                return $this->replaceVariablesInPropertyValue($fallbackValue);
+            } else {
```
oliverklee, line 67: Ditto. (I'm not a big fan of early returns that are not guard clauses, though, and would prefer to make this if/else construct not have early returns.)

src/HtmlProcessor/CssVariableEvaluator.php
```diff
@@ -0,0 +1,197 @@
...
+                                \\(
+                                    # using the named recursable pattern
+                                    (?&recursable)
+                                \\)
+                                # sequence without parentheses, except in strings
+                                (?&noparentheses)
+                            )*+
+                        )
+                    )?+
+                \\)
```
oliverklee, line 119: Wow … see the master of regexp at work! :exploding_head: :+1:

src/HtmlProcessor/CssVariableEvaluator.php
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * @param array<non-empty-string, string> $declarations
+     *
+     * @return array<non-empty-string, string>|false `false` is returned if no substitutions were made.
+     */
+    private function replaceVariablesInDeclarations(array $declarations)
+    {
+        $substitutionsMade = false;
+        $result = \array_map(
+            function (string $propertyValue) use (&$substitutionsMade): string {
```
oliverklee, line 135: Using all these used array functions feels to me like we maybe could use another class structure to have more type safety and to make the code more communicative. What do you think?

oliverklee (commented):
Amazing work! I've added some comments, both regarding the general architecture as well as some nitpicks.

I've also rebased the branch and autoformatted the README. So you'll probably need to hard-reset your local branch to the remote one (or delete the local copy and then use the remote one).

tests/Unit/HtmlProcessor/CssVariableEvaluatorTest.php
JakeQZ: Yes, the value comes from the a definition for the element itself or the closest ancestor.  Or, in other words, custom properties are inherited.

tests/Unit/HtmlProcessor/CssVariableEvaluatorTest.php
JakeQZ: All of the strings contain both double and single quotes (one type for the `style` attribute value, and the other type for the `content` property value within).

src/HtmlProcessor/CssVariableEvaluator.php
JakeQZ: Yes, changed.

src/HtmlProcessor/CssVariableEvaluator.php
JakeQZ: I've added a local variable for the return value.

src/HtmlProcessor/CssVariableEvaluator.php
JakeQZ: Ditto.

tests/Unit/HtmlProcessor/CssVariableEvaluatorTest.php
JakeQZ: Can't believe how many times I made the same typo.  Search and replace found it in three method names (though not any comments or documentation, etc.).

src/HtmlProcessor/CssVariableEvaluator.php
JakeQZ: Not sure what you mean or are suggesting.  The method performs a transformation on each element of an array, for which `array_map` does the job.  The reason to keep track of whether this transformation made any changes is an optimization.  PHP's copy-on-write handling of arrays helps, but if a change is made, comparing `$newArray !== $oldArray` means checking each element (until the difference is found).  (If the array is unmodified, the test is instant, since internally they'll still be referencing the same array.)

JakeQZ:
> autoformatted the README.

Many of the other code examples (PHP, CSS or HTML) in the README still have only two spaces indentation.  (Those that are CSS or HTML are in bullet-points, though the PHP ones are also at main-content level.)

I've spotted I mistake in the sample HTML I added, which I'll fix.

src/HtmlProcessor/CssVariableEvaluator.php
JakeQZ: To be fair, this was mostly based on an [Stack Overflow answer](https://stackoverflow.com/questions/54143257/recursive-regex-for-matching-everything-in-parenthesis-pcre/54143883#54143883) by someone who is clearly much more expert, though I did fix a couple of issues with that.  The trick for handling quoted strings could certainly be useful elsewhere in the project where we are not yet using (or able to use) CssSelector or PHP-CSS-Parser - there are certainly some constructs that will currently fail.

tests/Unit/HtmlProcessor/CssVariableEvaluatorTest.php
JakeQZ: The `@property` rule allows changing of inheritance behaviour.  It would be possible to support it (somehow), but I don't see any great need at the moment.  I don't think it is commonly used (I wasn't even aware of it) - whereas CSS variables are everywhere nowadays (hence my keenness to support them).

oliverklee (approved):
:tada:

PR #1349, opened 2024-10-25
[TASK] Don't cache the PHPUnit results

oliverklee:
PHPUnit result caching is not related to performance, but only to running failed tests only (which is a feature we don't use).

Disabling the PHPUnit result cache keeps the result cache files from littering the project directory.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -5,6 +5,7 @@
          beStrictAboutOutputDuringTests="true"
          beStrictAboutTodoAnnotatedTests="true"
          bootstrap="vendor/autoload.php"
+         cacheResult="false"
          colors="true"
          convertDeprecationsToExceptions="true"
          forceCoversAnnotation="true"


JakeQZ (approved):
I remember well an issue with PHPUnit running failed tests first, but in a completely different order from the original test order.  So if you had a lot of failing tests (possibly all for the same reason), and were working on fixing the first failure, it was a headache because the test result you were focusing on was then buried in a long list of failures (or not there at all, if you'd actually fixed it, but not many of the others, so you were then looking for a possibly-non-existent needle in a haystack).

Perhaps the out-of-orderedness of PHPUnit running failed tests first has been fixed in the version we now use.  If so, it would be useful to save a bit of time on the test run.  But that's beyond the scope of this PR.  For now, we don't use the cache so should not create it.

Will need a rebase :/

JakeQZ:
> Will need a rebase :/

I took the plunge and used GitHub's visual merge editor to 'rebase', which if nothing goes wrong achieves the same result (though does a 'merge' rather than a 'rebase'), given that there was only one clear and obvious conflict :)

PR #1350, opened 2024-10-26
[CLEANUP] Reformat the PHPUnit configuration file

oliverklee:
This gets rid of the 9-space indentation and also will make diffs cleaner in the future.


JakeQZ (approved):
+1

PR #1353, opened 2024-10-26
[CLEANUP] Autoformat `CONTRIBUTING.md`

oliverklee:
This will allow autoformatting when we do changes without introducing unrelated formatting changes.


JakeQZ (approved):
Whitespace-only changes, without removing or adding an indentation per se.

PR #1354, opened 2024-10-27
[DOCS] Rework the development deps section in `CONTRIBUTING.md`

oliverklee:


CONTRIBUTING.md
```diff
@@ -53,18 +53,23 @@ first before investing a lot of time in writing code.
...
 composer install
-composer require --dev slevomat/coding-standard:^4.0
 ```
 
-Note that the development dependencies (in particular, for PHP_CodeSniffer)
-require PHP 7.0 or later. The second command installs the PHP_CodeSniffer
-dependencies and should be omitted if specifically testing against an earlier
-version of PHP, however you will not be able to run the static code analysis.
+We also have some optional development dependencies that require higher PHP
+versions that the lowest PHP version this project supports. Hence they are not
```
JakeQZ, line 64: Typo ("that" should be "than").
```suggestion
versions than the lowest PHP version this project supports. Hence they are not
```

JakeQZ (changes requested):
Theres a typo.  Otherwise fine (and long overdue).

JakeQZ (approved):

PR #1361, opened 2024-11-03
[TASK] Drop support for Symfony versions without PHP 8.4 support

### LINKED ISSUE #1359: Raise the Symfony bugfix versions to ensure PHP 8.4 compatibility

oliverklee:
These Symfony versions are PHP-8.4-compatible:

- 7.1.0
- 7.0.3
- 6.4.3
- 6.3.12
- 6.2 (none)
- 6.1 (none)
- 6.0 (none)
- 5.4.35

The relevant thing is whether the `CssSelector/XPath/Translator.php` class does not use implicit nullable parameters anymore.

So we'll need to drop support for Symfony 6.0 through 6.2 altogether and raise the minimum required bugfix versions of the other minor versions.

Fixes #1359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,13 @@ Please also have a look at our
 
 ### Changed
 
+- Raise the minimum required Symfony bugfix versions (#1361)
+
 ### Deprecated
 
 ### Removed
 
-- Drop support for Symfony 4.4 (#1358)
+- Drop support for Symfony 4.4, 6.0., 6.1, 6.2 (#1358, #1361)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ Please also have a look at our
 
 ### Removed
 
-- Drop support for Symfony 4.4, 6.0., 6.1, 6.2 (#1358, #1361)
+- Drop support for Symfony 4.4, 6.0, 6.1, 6.2 (#1358, #1361)
 
 ### Fixed
 
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "ext-dom": "*",
         "ext-libxml": "*",
         "sabberworm/php-css-parser": "^8.7.0",
-        "symfony/css-selector": "^5.4.35 || ~6.3.12 || ~6.4.3 || ^7.0.3"
+        "symfony/css-selector": "^5.4.35 || ~6.3.12 || ^6.4.3 || ^7.0.3"
     },
     "require-dev": {
         "php-parallel-lint/php-parallel-lint": "1.4.0",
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -41,7 +41,7 @@
         "ext-dom": "*",
         "ext-libxml": "*",
         "sabberworm/php-css-parser": "^8.7.0",
-        "symfony/css-selector": "^5.4.0 || ^6.0.0 || ^7.0.0"
+        "symfony/css-selector": "^5.4.35 || ~6.3.12 || ~6.4.3 || ^7.0.3"
```
JakeQZ, line 44: ["~1.2 is equivalent to >=1.2 <2.0.0, while ~1.2.3 is equivalent to >=1.2.3 <1.3.0."](https://getcomposer.org/doc/articles/versions.md#tilde-version-range-).  So this would not permit version 6.5.

If Symfony adheres to semver, we could just require `^6.4.3` and not even allow `~6.3.12`, since there should be no breaking changes between 6.3 and 6.4.  But I think it's still better to also allow 6.3 (since we have no reason not to), just in case there is an edge case where a (misconfigured) package permits 6.3 but not 6.4.

CHANGELOG.md
```diff
@@ -12,11 +12,13 @@ Please also have a look at our
...
 ### Changed
 
+- Raise the minimum required Symfony bugfix versions (#1361)
+
 ### Deprecated
 
 ### Removed
 
-- Drop support for Symfony 4.4 (#1358)
+- Drop support for Symfony 4.4, 6.0., 6.1, 6.2 (#1358, #1361)
```
JakeQZ, line 21: Dropping 6.0, 6.1 and 6.2 shouldn't be of any consequence, if Symfony adheres to semver, as we still support 6.x.  But OTOH [on the other hand] there's no reason not to be explicit about the change.

composer.json
JakeQZ:
```suggestion
        "symfony/css-selector": "^5.4.35 || ~6.3.12 || ^6.4.3 || ^7.0.3"
```

JakeQZ (changes requested):
This needs altering to allow Symfony 6.5+.  Otherwise I think is fine - my other comments are just thoughts (I'm running with `--verbose`).

CHANGELOG.md
JakeQZ, line 21: Sorry, I missed that there should not be a dot after "6.0".
```suggestion
- Drop support for Symfony 4.4, 6.0, 6.1, 6.2 (#1358, #1361)
```

JakeQZ (changes requested):
Sorry, I missed I typo in previous review, which I've now spotted :/

JakeQZ (approved):

PR #1362, opened 2024-11-03
[TASK] Make the Composer scripts for fixing things more consistent

oliverklee:
- always use the `fix:` prefix for all fixing-related Composer scripts
- add a `fix` script to conveniently run all fixers
- add a missing script description

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -105,7 +105,7 @@ code documentation.
 You can autoformat your code using the following command:
 
 ```shell
-composer php:fix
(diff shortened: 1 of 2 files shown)


JakeQZ (approved):
I agree this is a more logical structure.

These scripts aren't used by any GitHub Actions so there are no corresponding changes needed there.

JakeQZ:
There wasn't previously a missing script description, so I removed that bullet-point from the commit message.  (The added description was for the newly added `fix` script.)

PR #1363, opened 2024-11-05
[TASK] Integrate composer-normalize in the static analysis CI job

oliverklee:


JakeQZ (approved):
Neat.  Does no harm I think to also install the Composer dependencies even though the normalize command uses a PHIVE dependency, and would also make it easier to change the normalize command to use a Composer dependency instead, should we ever want to.

PR #1365, opened 2024-11-07
[FEATURE] Add Rector to our development toolchain

### LINKED ISSUE #890: Keep the code in shape with Rector
https://github.com/rectorphp/rector is a tool that automatically can do PHP language level upgrades. The TYPO3 project already uses it with great success.

oliverklee:
Also rename the `fix:php` script to be more specific now that we have two PHP-related fixing scripts.

Fixes #890

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,6 +96,7 @@ jobs:
           - composer:normalize
           - php:fixer
           - php:md
+          - php:rector
           - php:stan
         php-version:
           - '8.3'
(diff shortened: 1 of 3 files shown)


JakeQZ (approved):
The `config.php` matches that in PHP-CSS-Parser.

JakeQZ:
> Could you please add tests to make sure this change works as expected?

The newly added PHP file happens to be a config file, so does not need tests added for it.

JakeQZ:
> > Could you please add tests to make sure this change works as expected?
> 
> The newly added PHP file happens to be a config file, so does not need tests added for it.

How do you override the github-actions bot's request for changes?

oliverklee:
> How do you override the github-actions bot's request for changes?

As an admin on this repository (which we both are, as far as I can tell), there is a checkbox "merge anyway" that allows us to merge PRs even if some requirements are not fulfilled.

PR #1366, opened 2024-11-07
[CLEANUP] Make the Markdown code highlighting consistent

oliverklee:
So far, we had a mixture of `bash` and `shell`. Let's unify this.


JakeQZ (approved):
[There does not seem to be a definitive best practice](https://stackoverflow.com/questions/20303826/how-to-highlight-bash-shell-commands-in-markdown).  At least if we're consistent, it would be easier to change all with search and replace if needed.

PR #1372, opened 2024-11-10
[TASK] Add more Rector rules

oliverklee:


config/rector.php
```diff
@@ -11,9 +16,39 @@
             __DIR__ . '/../tests',
         ]
     )
-    ->withPhpSets()
-    ->withRules(
-        [
-            // AddVoidReturnTypeWhereNoReturnRector::class,
-        ]
-    );
+    ->withPhpVersion(PhpVersion::PHP_73)
```
JakeQZ, line 19: [The best practise is to let [it] pickup the PHP version from composer.json](https://getrector.com/documentation/php-version-features).  So I don't think we need this (as it will be something more to update), unless it doesn't work as expected and desired otherwise.

config/rector.php
```diff
@@ -11,9 +16,39 @@
...
+    ->withPhpSets(
+        true
+    )
```
JakeQZ, line 22: I can't find any documentation for this having a simple parameter of `true`.  All I found was [this](https://getrector.com/documentation/set-lists).  So it doesn't seem right.

In any case, the parameter should be on the same line as the method call.

config/rector.php
```diff
@@ -11,9 +16,39 @@
...
+    ->withSets([
+        // Rector sets
+
+        LevelSetList::UP_TO_PHP_73,
```
JakeQZ, line 26: Where are these documented?  I can't find it.

config/rector.php
```diff
@@ -11,9 +16,39 @@
...
+        // LevelSetList::UP_TO_PHP_74,
+        // LevelSetList::UP_TO_PHP_80,
+        // LevelSetList::UP_TO_PHP_81,
+        // LevelSetList::UP_TO_PHP_82,
+        // LevelSetList::UP_TO_PHP_83,
```
JakeQZ, line 31: I don't think we need these as comments; it should be obvious by deduction what the next ones will be.

config/rector.php
```diff
@@ -11,9 +16,39 @@
...
+        // SetList::INSTANCEOF,
+        // SetList::NAMING,
+        // SetList::PRIVATIZATION,
+        SetList::STRICT_BOOLEANS,
+        SetList::TYPE_DECLARATION,
+
+        // PHPUnit sets
+
+        PHPUnitSetList::PHPUNIT_90,
+        // PHPUnitSetList::PHPUNIT_100,
```
JakeQZ, line 46: Ditto.

config/rector.php
```diff
@@ -11,9 +16,39 @@
...
+        // PHPUnitSetList::PHPUNIT_CODE_QUALITY,
+    ])
+    ->withRules([
+        AddVoidReturnTypeWhereNoReturnRector::class,
+    ])
+    ->withImportNames(true, true, false)
```
JakeQZ, line 52: [This is all I found by way of documentation about `withImportNames`](https://getrector.com/documentation/import-names), and is insufficient to complete this review.

config/rector.php
```diff
@@ -11,9 +16,39 @@
...
+    ->withSkip([
+    ]);
```
JakeQZ, line 54: [I don't think we need this if we're not skipping anything](https://getrector.com/documentation/ignoring-rules-or-paths).

JakeQZ (changes requested):
I've struggled to find decent documentation.  What I've found has been very sparse.  Where are you getting your info from?

oliverklee:
> I've struggled to find decent documentation. What I've found has been very sparse. Where are you getting your info from?

Mostly from reading the source code after clicking on the method/constant/class in the configuration file.

Also, I copy from the configuration files from other projects, e.g. https://github.com/FriendsOfTYPO3/tea/blob/main/rector.php (that also uses some TYPO3-specific rulesets).

config/rector.php
oliverklee: Ah, I wasn't aware if of this behavior. Removed now.

config/rector.php
oliverklee: Removed.

config/rector.php
oliverklee: https://github.com/rectorphp/rector/blob/main/src/Set/ValueObject/LevelSetList.php and https://github.com/rectorphp/rector/blob/main/config/set/level/up-to-php73.php etc. Basically, the `LevelSetList::UP_TO_PHP_*` set lists are about raising the PHP language level to the corresponding PHP version.

config/rector.php
oliverklee: Done.

config/rector.php
oliverklee: Done.

config/rector.php
oliverklee: Done.

config/rector.php
oliverklee: This set of arguments means:

1. `import` classes instead of using the FQCM.
2. Also apply this to class names in doc blocks.
3. Do not `import` top level classes like `RuntimeException`

JakeQZ (approved):

PR #1373, opened 2024-11-10
[TASK] Use more native types in the tests

oliverklee:
All glory to the Rector!


JakeQZ (approved):
Some of these are quite sharp finds.  I wonder if internally it uses the same analysis engine as PHPStan or Psalm - both of which seem like they could be enhanced to apply suggested fixes in some cases.

oliverklee:
>  I wonder if internally it uses the same analysis engine as PHPStan or Psalm

Yes, it does - Rector has PHPStan as a dependency: https://packagist.org/packages/rector/rector

Issue #1219, opened 2023-08-09
Upgrade to phpunit-data-provider 3.0.0


oliverklee:
https://github.com/t-regx/phpunit-data-provider/releases/tag/3.0.0

JakeQZ:
Note:

> Package `rawr/cross-data-providers` is abandoned, you should avoid using it. Use `rawr/phpunit-data-provider` instead.

But I think you knew this already.

PR #1374, opened 2024-11-16
[TASK] Add `rawr/phpunit-data-provider` as a dev dependency

oliverklee:
This is the first step of migrating from the abandoned `rawr/cross-data-providers` package to this package.

Part of #1219


JakeQZ (approved):
Looks like this package combines the functionality that was in the obsolete package, into a more general set of data provider utilities.

PR #1376, opened 2024-12-12
[TASK] Upgrade to PHPStan 2.x and Rector 2.x

oliverklee:
We'll allow `assertInstanceOf` checks in the tests again in a later change.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,9 +17,11 @@ updates:
     allow:
       - dependency-type: "development"
     ignore:
+      - dependency-name: "phpstan/*"
       - dependency-name: "phpunit/phpunit"
         versions: [ ">= 10.0.0" ]
       - dependency-name: "rawr/cross-data-providers"
+      - dependency-name: "rector/rector"
     versioning-strategy: "increase"
     commit-message:
       prefix: "[Dependabot] "
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -7,6 +7,3 @@ parameters:
   paths:
     - src
     - tests
-
-  # Allow instanceof checks, particularly in tests
-  checkAlwaysTrueCheckTypeFunctionCall: false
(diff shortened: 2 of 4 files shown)


oliverklee:
For this change, we'll need to drop support for PHP < 7.4 first.

phpstan.neon
```diff
@@ -7,6 +7,3 @@ parameters:
   paths:
     - src
     - tests
-
-  # Allow instanceof checks, particularly in tests
-  checkAlwaysTrueCheckTypeFunctionCall: false
```
JakeQZ, line 12: I agree this check should not be disabled per se, and if possible the setting could be applied on a per directory basis: i.e. for `tests` which can by static analysis be proven to pass.  Though I'd quite like the double confirmation.

JakeQZ (commented):
It looks like we'd need have a minimum dev requirement of PHP 7.4 for this upgrade, whilst there would no such requirement for users of the package.  Is such a configuration possible?

oliverklee:
> It looks like we'd need have a minimum dev requirement of PHP 7.4 for this upgrade, whilst there would no such requirement for users of the package. Is such a configuration possible?

No, not as far as I know.

oliverklee:
We could allow both versions, though.

JakeQZ:
> We could allow both versions, though.

Yes, I think that's the solution, with `||` versions in Composer.

.github/dependabot.yml
```diff
@@ -17,9 +17,11 @@ updates:
     allow:
       - dependency-type: "development"
     ignore:
+      - dependency-name: "phpstan/*"
       - dependency-name: "phpunit/phpunit"
         versions: [ ">= 10.0.0" ]
       - dependency-name: "rawr/cross-data-providers"
+      - dependency-name: "rector/rector"
```
JakeQZ, line 24: I presume that the GitHub action is unable to cope with `OR` depencies.  This can be revisted later if and when it is able.

JakeQZ (approved):
This works, I think.

PR #1377, opened 2024-12-12
[TASK] Allow `assertInstanceOf` calls in tests

oliverklee:
We find those checks useful.


JakeQZ (approved):
Agree.  That a powerful static analysis tool can determine a test will always pass does not mean it shouldn't be in place.  (By extension, as a powerful human, I can determine that all the tests will pass, but that does not mean we should not have any unit tests.)

PR #1378, opened 2024-12-17
[TASK] Move and polish the PHPStan configuration and scripts

oliverklee:
Move the configuration and baseline files to `config/` in order to declutter the project main directory (following our sister project).

Stop using the GitHub formatter as this makes it impossible for humans to see any errors if the CI pipeline fails.


JakeQZ (approved):
This makes the PHPStan configuration consistent with that in the sister project PHP-CSS-Parser.

PR #1383, opened 2024-12-31

oliverklee:


config/php-cs-fixer.php
```diff
@@ -71,7 +71,9 @@
             'php_unit_mock_short_will_return' => true,
             'php_unit_set_up_tear_down_visibility' => true,
             'php_unit_test_annotation' => ['style' => 'annotation'],
-            'php_unit_test_case_static_method_calls' => ['call_type' => 'self'],
+            // `'methods' => ['createStub' => 'this']` is required as long as we are not on PHPUnit 10.x yet
+            // (where that method was made static).
+            'php_unit_test_case_static_method_calls' => ['call_type' => 'self', 'methods' => ['createStub' => 'this']],
```
JakeQZ, line 76: I'm surprised that we need this, given that we do not call `createStub` AFAICS, and neither does PHPUnit itself.  I'm also mystified as to what change between 3.64 and 3.66 of the Fixer means that we would now need it.

JakeQZ (commented):
I have a couple of questions about `createStub`.

config/php-cs-fixer.php
oliverklee: Some background on this:

PHPUnit provides [two kinds of test doubles](https://docs.phpunit.de/en/11.5/test-doubles.html): Mocks and stubs. They work in similar ways, and using the correct type helps both the reader recognize the purpose of the corresponding test double more easily as well as PHPStan to recognize incorrect usage.

With PHPUnit version 10, the method `createStub` [was made `static`](https://github.com/sebastianbergmann/phpunit/commit/6d98a507127c215d70a0e108bd72b05fcde58733).

PHP-CS-Fixer 3.66.0 introduced the [feature of (by default) making calls to `createStub` static as well](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/8319), which breaks tests that are on PHPUnit 9.x. (Up to and including version 3.65.0, PHP-CS-Fixer left calls to `createStub` alone AFAIK.)

As we're currently using neither `getMock` nor `getStub`, I'll remove the corresponding line from the PHP-CS-Fixer configuration to keep things simple.

JakeQZ (approved):

PR #1385, opened 2025-01-03
[TASK] Switch one testcase to the new data provider package

oliverklee:
Part of #1219


JakeQZ (approved):
The change is consistent with the migration guide at https://github.com/t-regx/phpunit-data-provider/releases/tag/3.0.0.

Issue #1360, opened 2024-11-03
Get `CONTRIBUTING.md` up to date


oliverklee:
- [x] Contributor Code of Conduct
- [x] General workflow
- [x] About code reviews
- [x] Install the development dependencies
- [x] Unit-test your changes
- [x] Coding Style
- [x] Git commits
- [x] Creating pull requests (PRs)
- [x] Rebasing
- [x] Move "Steps to release a new version" from README

oliverklee:
All done.

PR #1387, opened 2025-01-06
[TASK] Polish the git rebasing section in `CONTRIBUTING.md`

oliverklee:
Part of #1360

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,20 +142,20 @@ To rebase a feature branch to the latest `main`:
 1. Make sure that your local copy of the repository has the most up-to-date
    revisions of `main` (this is important, otherwise you may end up rebasing to
    an older base point):
-   ```sh
-   git checkout main
+   ```bash
+   git switch main
    git pull
    ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -168,8 +168,8 @@ To rebase a feature branch to the latest `main`:
 
    If there were no conflicts, skip this and the next step.
 1. Mark the conflicting files as resolved and continue the rebase
-   ```sh
-   git add *.*
+   ```bash
+   git add .
```
JakeQZ, line 172: From what I understand, this means that `git` will to the file system traversal, rather than the shell, which is probably better due to being OS-independent.

JakeQZ (approved):
All changes seem like improvements to me.

I wrote this section based on rebasing as a collaborator.  I think the instructions for contributors (with PRs on a fork, particularly the examples) would vary from this, and that this section really needs to be written with that target audience in mind.  I'll create a ticket for it...

JakeQZ:
> This section really needs to be written with [contributors as the] target audience in mind. I'll create a ticket for it...

#1388

PR #1390, opened 2025-01-19

oliverklee:


JakeQZ (approved):

JakeQZ (changes requested):
Ah, I see, it breaks something, and this PR is now 'draft'.

Personally (YMMV), I'm not super-keen to update things if there's no need to.  If it ain't broke, don't fix it (because you'll just end up breaking more stuff).

The PHIVE PHARs will be installed with the latest permissable version locally upon running `phive update`.  Do we not also run that during CI?  If so, we don't need to keep updating these files.  If not, maybe we should - ?

We concluded that we needed to restrict versions of the static analysis tools in Composer to avoid breaking updates, though I wonder if we might be able to use [Tilde Version Range](https://getcomposer.org/doc/articles/versions.md#tilde-version-range-) - if the tools really adhere to semver, then point releases should not break anything.

No specific changes requested beyond fixing the CI.

JakeQZ:
We should probably use `~` version in the PHIVE config, so that contributors don't pick up breaking changes.

oliverklee:
> Personally (YMMV), I'm not super-keen to update things if there's no need to. If it ain't broke, don't fix it (because you'll just end up breaking more stuff).

Okay, then let's only update the tools when they either bring features of bug fixes we need/want, or when we'll need to update our code for the new version.

> The PHIVE PHARs will be installed with the latest permissable version locally upon running phive update. Do we not also run that during CI? If so, we don't need to keep updating these files. If not, maybe we should - ?

CI is running `phive install` (not `phive update`) to get the exact versions specified as `installed` in `phars.xml`. That's what contributors are expected to use and what's documented in `CONTRIBUTING.md`. This ensures that tests results are reproducible (except for cases like removed/expired GPG key).

`phive update` is intended to update the tools, and it changes `phars.xml`.

JakeQZ (commented):
> Now that the new PHPStan version comes with changes to the baseline, I think we should get this merged.

I think so too.  Though does this also apply to 1.12.13?  Are we testing with that version in the CI build?

oliverklee:
> Though does this also apply to 1.12.13? Are we testing with that version in the CI build?

No, we're only testing with PHPStan 2.x on CI. We only have PHPStan 1.x in the requirements so we can install the Composer packages with lower PHP versions for running the unit tests.

JakeQZ (approved):
> We only have PHPStan 1.x in the requirements so we can install the Composer packages with lower PHP versions for running the unit tests.

Ah, I see.  PHP >= 7.4 is required to run all the dev tools, though the unit tests will also run under PHP 7.3, for testing PHP 7.3 support.

I tested with PHP 7.3 and 7.4.  Composer will find a set of dependencies to satisfy 7.3.  But `composer-normalize` seems to throw a false error: 'Your Composer dependencies require a PHP version ">= 7.4.0"'.  Is there anything in the config telling `composer-normalize` that PHP 7.4 is required?  If not, it looks like a bug in the tool - I didn't even know it did version constraint checking.

(Also noted that PHPStan version 1.x won't run anyway because it does not support level 10.)

JakeQZ:
> `composer-normalize` seems to throw a false error: 'Your Composer dependencies require a PHP version ">= 7.4.0"'.

The message is misleading, since we install `composer-normalize` via PHIVE.  `composer-normalize` does not follow semver: version 2.21.0 dropped support for PHP 7.3, which is why the error occurs.

Adjusting the `require-dev` section of `composer.json` to require PHP >= 7.4 would not be viable, since that would prevent running the unit tests on 7.3.  But an explanatory note in the contributing guidlines would probably not go amiss...

oliverklee:
I've created #1411 to track the documentation.

PR #1394, opened 2025-01-25
[TASK] Move PHPMD from PHIVE to Composer

oliverklee:
The PHIVE version is no longer installable due to the GPG key not being available any more. Also, we can now install it via Composer just fine.


JakeQZ (approved):
I prefer Composer, whenever possible :)

PR #1395, opened 2025-02-16
[TASK] Add PHPStan rules from Rector Type Perfect

oliverklee:
This is an opiniated set of PHPStan rules to make code more explicit and intuitive to read.

https://github.com/rectorphp/type-perfect

As our code already is in pretty good shape, this new set of rules does not add any new warnings to our baseline.


JakeQZ (approved):
> As our code already is in pretty good shape, this new set of rules does not add any new warnings to our baseline.

Nice to know 🙂

PR #1398, opened 2025-03-23
[CLEANUP] Make a type annotation in a test more specific

oliverklee:


JakeQZ (commented):
Could `value` also be a `non-empty-string`?

tests/Unit/Css/StyleRuleTest.php
```diff
@@ -65,7 +65,9 @@ public function hasAtLeastOneDeclarationForDeclarationBlockWithOneRuleReturnsTru
     }
 
     /**
-     * @return array<string, array{0: array<int, array{property: string, value: string}>, 1: string}>
+     * @return array<string, array{
+     *     0: array<int, array{property: non-empty-string, value: non-empty-string}>, 1: string
```
JakeQZ, line 69: With the array type definition now spanning multiple lines, I think it would be clearer if each array element was on a separate line, as we would do for PHP multiline array definitions.  Though I suspect a trailing comma on the last line wouldn't work here (have not tested), so the analogy ends there.

```suggestion
     *     0: array<int, array{property: non-empty-string, value: non-empty-string}>,
     *     1: string
```

JakeQZ (changes requested):
I'd suggest that the multiline array type definition has each element on a separate line.

tests/Unit/Css/StyleRuleTest.php
JakeQZ: Don't commit this suggestion - see below.

JakeQZ (approved):

PR #1351, opened 2024-10-26
[CLEANUP] Autoformat the code

oliverklee:
This will help keep unrelated formatting changes out of future changes where the developer autoformats the code.

Apart from the formatting changes, also drop a nonsensical `;` after a PHPDoc type annotation that got moved by the autoformat.

PR #1401, opened 2025-03-31
[CLEANUP] Autoformat the code

oliverklee:
This avoids unrelated formatting changes when upcoming changes are introduced (where the code should be in an autoformatted state).

### DIFF
--- a/src/Css/CssDocument.php
+++ b/src/Css/CssDocument.php
@@ -63,7 +63,7 @@ final class CssDocument
     private function hasNestedAtRule(string $css): bool
     {
         return (new Preg())
-            ->match('/@(?:media|supports|(?:-webkit-|-moz-|-ms-|-o-)?+(keyframes|document))\\b/', $css) !== 0;
+                ->match('/@(?:media|supports|(?:-webkit-|-moz-|-ms-|-o-)?+(keyframes|document))\\b/', $css) !== 0;
     }
 
     /**
--- a/tests/Support/Traits/CssDataProviders.php
+++ b/tests/Support/Traits/CssDataProviders.php
@@ -29,19 +29,19 @@ trait CssDataProviders
     {
         $datasetsWithAtMediaRuleSelectorListAndPropertyDeclaration = self::crossDatasetWithItself([
             'unminified CSS with `@media` rule, selector list, and property declaration'
-                => ['@media screen { html, body { color: green; } }'],
+            => ['@media screen { html, body { color: green; } }'],
             'minified CSS with `@media` rule, selector list, and property declaration'
-                => ['@media screen{html,body{color:green}}'],
+            => ['@media screen{html,body{color:green}}'],
             'CSS with `@media` rule, selector list, and property declaration, with extra spaces'
-                => ['  @media  screen  {  html  ,  body  {  color  :  green  ;  }  }  '],
+            => ['  @media  screen  {  html  ,  body  {  color  :  green  ;  }  }  '],
             'CSS with `@media` rule, selector list, and property declaration, with linefeeds'
-                => ["\n@media\nscreen\n{\nhtml\n,\nbody\n{\ncolor\n:\ngreen\n;\n}\n}\n"],
+            => ["\n@media\nscreen\n{\nhtml\n,\nbody\n{\ncolor\n:\ngreen\n;\n}\n}\n"],
             'CSS with `@media` rule, selector list, and property declaration, with Windows line endings'
-                => ["\r\n@media\r\nscreen\r\n{\r\nhtml\r\n,\r\nbody\r\n{\r\ncolor\r\n:\r\ngreen\r\n;\r\n}\r\n}\r\n"],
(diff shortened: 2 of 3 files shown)


oliverklee:
This is a pre-patch to a pre-patch. This is the patch order:

1. autoformat
2. get rid of `array_map` and `array_combine` stuff with callables in tests which I fail to understand when reading
3. get rid of the abandoned `rawr/cross-data-providers`

src/Css/CssDocument.php
```diff
@@ -63,7 +63,7 @@ public function __construct(string $css, bool $debug)
     private function hasNestedAtRule(string $css): bool
     {
         return (new Preg())
-            ->match('/@(?:media|supports|(?:-webkit-|-moz-|-ms-|-o-)?+(keyframes|document))\\b/', $css) !== 0;
+                ->match('/@(?:media|supports|(?:-webkit-|-moz-|-ms-|-o-)?+(keyframes|document))\\b/', $css) !== 0;
```
JakeQZ, line 66: Why the double indentation?

tests/Support/Traits/CssDataProviders.php
```diff
@@ -29,19 +29,19 @@ public function provideEquivalentCompleteCss(): array
     {
         $datasetsWithAtMediaRuleSelectorListAndPropertyDeclaration = self::crossDatasetWithItself([
             'unminified CSS with `@media` rule, selector list, and property declaration'
-                => ['@media screen { html, body { color: green; } }'],
+            => ['@media screen { html, body { color: green; } }'],
```
JakeQZ, line 32: I think this is more readable with the indentation.

We would indent a long variable assignment, and not write

```php
$longVariableName
= $this->longExpression($someValue);
```

I think defining an array value is the same kind of thing.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -224,9 +224,9 @@ public function provideHtmlWithOptionalTagsOmitted(): array
             // broken: OPTGROUP end tag omission with OPTGROUP element following
             'OPTGROUP end tag omission at end of list' => [
                 '<select><optgroup><option> 1 </option><option> 2 </option></optgroup>'
-                    . '<optgroup><option> A </option><option> B </option></select>',
+                . '<optgroup><option> A </option><option> B </option></select>',
```
JakeQZ, line 227: Again, I prefer the indentation.  It makes it easier to see where each element value begins.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -1260,8 +1260,8 @@ private static function assertContainsHtml(string $needle, string $haystack, str
         $needleMatcher = \preg_quote($needle, '%');
         $needleMatcherWithNewlines = (new Preg())->replace(
             '%\\\\<(?:body|ul|dl|optgroup|table|tr|hr'
-                . '|/(?:li|dd|dt|option|optgroup|caption|colgroup|thead|tbody|tfoot|tr|td|th'
-                . '|p|dl|h[1-6]|menu|ol|pre|table|ul|address|blockquote|div|fieldset|form))\\\\>%',
+            . '|/(?:li|dd|dt|option|optgroup|caption|colgroup|thead|tbody|tfoot|tr|td|th'
+            . '|p|dl|h[1-6]|menu|ol|pre|table|ul|address|blockquote|div|fieldset|form))\\\\>%',
```
JakeQZ, line 1264: The indentation makes it easier to see where each parameter expression begins.

JakeQZ (commented):
I'm not convinced your autoformat settings are the best...

JakeQZ:
> I'm not convinced your autoformat settings are the best...

Thought I had a sense of déjà vu: the autoformatting in #1351 originally included these odd indentation changes, which were ultimately taken out of that PR.

oliverklee:
My workflow usually goes like this:

1. code/refactor something
2. autoformat in PhpStorm every now and then
3. before creating the PR, run `composer fix`

If we keep formatting in the code that gets changed by PhpStorm's autoformat (and does not get changed back by PHP-CS-Fixer), I need to either stop using autoformat when coding (and manually format my code for the things that PHP-CS-Fixer does not do) or manually remove the changes that the PhpStorm autoformat does. Both options cost me time, and also it's a kind of work that feels unnecessary to me and that I detest.

So this PR is not about "Which formatting do we like best?", but about "Can we live with this formatting for now?".

These changes are with the default "PSR2+" code formatting preset in PhpStorm, which is the setting that is closest to the coding standard we use.

I'd propose like to propose this:

1. We get this PR merged.
2. If there are any style issues that we think needs improving, we add a corresponding PHP-CS-Fixer rule to our configuration.

Would you be okay with that, @JakeQZ ?

JakeQZ:
> So this PR is not about "Which formatting do we like best?", but about "Can we live with this formatting for now?".

OK.  I can live with this formatting for now.

I don't use PHPStorm.  Does it have individually configurable formatting rules like PHP-CS-Fixer?  If so, perhaps there are some that can be tweaked to get the desired formatting.  Though preferably we would find some options in the fixer that will achieve the desired result, as that would be available to all developers.

JakeQZ (approved):
Added #1402 to cover the formatting issues.

PR #1403, opened 2025-04-11
[CLEANUP] Reduce callbacks for data providers in the tests

oliverklee:
Let's reduce cognitive complexity to make the lives of the poor developers a bit easier.

This is in preparation to migrating more data providers to the new `rawr/phpunit-data-provider` package.

### DIFF
--- a/config/phpstan-baseline.neon
+++ b/config/phpstan-baseline.neon
@@ -243,24 +243,6 @@ parameters:
 			path: ../tests/Unit/HtmlProcessor/HtmlPrunerTest.php
 
 		-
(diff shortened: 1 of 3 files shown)


JakeQZ (approved):
I guess it's as broad as it's long.  I personally prefer a functional programming approach to make use of built-in functions like `array_map` (perhaps extending the offering), rather than a `for` loop, so that the operation on each element is compartmentalized into a well-named method (I'd accept that using a closure is lazy, though does keep the code all in one place if it's not used elsewhere).

Maybe the `for` loop approach makes it easier to see what is going on, and `array_map` has little use.  (Mileage may vary.)

The previous warnings from PHPStan look an area for improvement in the tool: it could in theory determine that the arrays passed to `array_combine` are of equal length and thus `false` would never be returned.  (Or the method itself could have been modified to throw an exception in the impossible case.)

PR #1404, opened 2025-04-27
[BUGFIX] Avoid deprecated string casts

oliverklee:
This paves the way towards PHP-CSS-Parser V9 compatibility.


JakeQZ (approved):
> @JakeQZ friendly ping

Sorry, I missed the email about this.

Issue #1317, opened 2024-09-17
Switch the tests from the abandoned `rawr/cross-data-providers` to `rawr/phpunit-data-provider`


oliverklee:
We should do this in small, self-contained PRs.

PR #1406, opened 2025-05-04
[TASK] Use `DataProvider::join()` to combine data providers

oliverklee:
Part of #1317


JakeQZ (approved):
Looks like it does what the original code (which should perhaps have used `array_merge` rather than `+`) did, as well as checking that the joined arrays have the same number of elements.

PR #1414, opened 2025-06-29
[TASK] Migrate another data provider to `DataProvider::cross()`

oliverklee:
Part of #1317

### DIFF
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -3690,7 +3690,7 @@ final class CssInlinerTest extends TestCase
             ['alone' => ['', '']]
         );
 
-        return DataProvider::cross($datasets1, $datasets2);
+        return DataProvider::join($datasets1, $datasets2);
     }
 
     /**
--- a/tests/Unit/CssInlinerTest.php
+++ b/tests/Unit/CssInlinerTest.php
@@ -3650,7 +3650,7 @@ final class CssInlinerTest extends TestCase
     }
 
     /**
-     * @return DataProvider<string, array<int, string>>
+     * @return DataProvider<string, array{0: string, 1: string, 2: string}>
      */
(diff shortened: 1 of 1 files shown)


tests/Unit/CssInlinerTest.php
```diff
@@ -3694,7 +3694,7 @@ public function provideValidAtRulesWithSurroundingCss(): array
             ['alone' => ['', '']]
         );
 
-        return $datasets;
+        return DataProvider::cross($datasets1, $datasets2);
```
JakeQZ: GitHub ate my comment.  This should be concatenating arrays, not creating a matrix of them.

Return value could be tightened to array of 3 strings, which would have detected this mistake.

oliverklee: Thanks, fixed.

tests/Unit/CssInlinerTest.php
```diff
@@ -3651,12 +3650,11 @@ public function inlineCssNotCopiesInlinableRuleAfterImportRuleToStyleElement():
     }
 
     /**
-     * @return array<string, array<int, string>>
+     * @return DataProvider<string, array<int, string>>
```
JakeQZ: As we're changing the type anyway, can the nested type be changed to specifiy a `list` of exactly three strings?

JakeQZ (changes requested):
`join()` looks like it does the job, but it would be good to also tighten the return type, so that PHPStan would tell us if otherwise.

tests/Unit/CssInlinerTest.php
oliverklee: Done and repushed.

JakeQZ (approved):
I was wondering if there was a neater way of specifying an array of exactly N elements of type M.  I guess there isn't, though didn't look into it.

PR #1416, opened 2025-06-29
[CLEANUP] Drop redundantly-registered Rector rule

oliverklee:
The removed rule already is part of one of the rulesets we use.

### DIFF
--- a/config/rector.php
+++ b/config/rector.php
@@ -6,7 +6,6 @@ use Rector\Config\RectorConfig;
 use Rector\PHPUnit\Set\PHPUnitSetList;
 use Rector\Set\ValueObject\LevelSetList;
 use Rector\Set\ValueObject\SetList;
-use Rector\TypeDeclaration\Rector\ClassMethod\AddVoidReturnTypeWhereNoReturnRector;
 
 return RectorConfig::configure()
     ->withPaths(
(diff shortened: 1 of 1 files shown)


JakeQZ (commented):
> The removed rule already is part of one of the rulesets we use.

I am sure you are right, but may I ask which ruleset?

oliverklee:
This is part of https://github.com/rectorphp/rector/blob/main/src/Config/Level/TypeDeclarationLevel.php, which by https://github.com/rectorphp/rector/blob/main/src/Configuration/RectorConfigBuilder.php gets included in the `SetList::TYPE_DECLARATION` set list.

JakeQZ:
> gets included in the `SetList::TYPE_DECLARATION` set list.

I see that we have this.  But cannot determine what the removed rule might have been doing.

I cannot find any documentation for Rector.  I am not happy about using a library with no documentation.  I do not want to waste time digging through its souce code to figure out how it works.  If there really is no documentation, then it is unusable, and should be dropped :/

oliverklee:
> Thanks for the links, but I still can't grok this. Can you point out more precisely the redundant rule?

Sorry, my explanation was not on point. The removed ruled is included in the `SetList::TYPE_DECLARATION` set list.

Rector provides a documentation page with a documentation of all rules (AFAICS). The corresponding rule is documented here:

https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#addvoidreturntypewherenoreturnrector

oliverklee:
In addition the the list of rules, there's also a comprehensive documentation here:
https://getrector.com/documentation/

Both are linked from the [project README](https://github.com/rectorphp/rector/?tab=readme-ov-file#documentation).

oliverklee:
… and then there's the [book on Rector](https://leanpub.com/rector-the-power-of-automated-refactoring).

PR #1419, opened 2025-07-18
[BUGFIX] Provide the authentication token for PHIVE on CI

oliverklee:
This will hopefully avoid the 403 errors when installing the PHIVE packages.


JakeQZ (approved):
🤞

PR #1425, opened 2025-07-29
[BUGFIX] Improve some type annotations

oliverklee:


src/CssInliner.php
```diff
@@ -424,7 +424,7 @@ private function normalizeStyleAttributesOfAllNodes(): void
     /**
      * Returns a list with all DOM nodes that have a style attribute.
      *
-     * @return \DOMNodeList
+     * @return \DOMNodeList<\DOMNode>
```
JakeQZ, line 427: The values will all be `DOMElement`s, but it would be asking a bit much of Stan to determine that from the XPath query, and an unnecessary performance hit to assert so otherwise.

However, if we can specify this as `\DOMNodeList<\DOMElement>` and do whatever else necessary to tell Stan that we know better, that would allow elimination of various `@var` annotations in code that iterates through the return value.

Also, the same method exists in `CssToAttributeConverter`, though doesn't have the error handling (to throw an exception if the `$expression` is malformed or the `$contextNode` is invalid, which will never actually be the case) that was added in #1027.

I think, before these methods diverge further, we should first add a trait which provides the `getAllNodesWithStyleAttribute` method.  And then rename it to `getAllElementsWithStyleAttribute`.

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ Please also have a look at our
 
 ### Fixed
 
+- Improve some type annotations (#1424)
```
JakeQZ: I don't think this should have a changelog entry, as the changes are only to an internal class and a private method.

JakeQZ (changes requested):
The change to `SimpleStringCache` is fine.

The change to `getAllNodesWithStyleAttribute()` I think needs one or two pre-PRs, and perhaps should be taken out of this PR to be addressed separately.

Neither change affects the public, so I think a changelog entry is superfluous.

src/CssInliner.php
JakeQZ: > ... we should first add a trait which provides the `getAllNodesWithStyleAttribute` method.

Perhaps `Pelago\Emogrifier\Utilities\XPath`.  It can be `use`d as `XPathUtilities`.  If it evolves into providing functionality beyond XPath, it can be renamed later.

JakeQZ:
> The change to `SimpleStringCache` is fine.

This class doesn't seem to be used yet in the classes that would like it, so I've added #1426 to capture that.

oliverklee:
I have just split out the change to the cache class to #1427.

PR #1428, opened 2025-07-30
[TASK] Make some type annotations more specific

oliverklee:


JakeQZ (approved):
Agree some of the DocBlock descriptions were not adding any value.

PR #1432, opened 2025-08-03
[CLEANUP] Drop redundant type annotations in `AbstractHtmlProcessor`

oliverklee:


src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -268,10 +247,6 @@ private function createRawDomDocument(string $html): void
     /**
      * Returns the HTML with added document type, Content-Type meta tag, and self-closing slashes, if needed,
      * ensuring that the HTML will be good for creating a DOM document from it.
-     *
-     * @param string $html
-     *
-     * @return string the unified HTML
```
JakeQZ, line 274: "Unified" was not an appropriate description here (it means 'made into a unit'), so I agree it's best to simply remove the `@return` annotation entirely.

src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -320,10 +291,6 @@ private function normalizeDocumentType(string $html): string
      * Adds a Content-Type meta tag for the charset.
      *
      * This method also ensures that there is a HEAD element.
-     *
-     * @param string $html
-     *
-     * @return string the HTML with the meta tag added
```
JakeQZ, line 326: The main description says that the method will add the `meta` tag, so the description for the `@return` annotation is superfluous repetition.

src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -449,10 +400,6 @@ private function removeHtmlTemplateElements(string $html): string
     /**
      * Makes sure that any self-closing tags not recognized as such by PHP's DOMDocument implementation have a
      * self-closing slash.
-     *
-     * @param string $html
-     *
-     * @return string HTML with problematic tags converted.
```
JakeQZ, line 455: Ditto.

JakeQZ (approved):

PR #1433, opened 2025-08-03
[CLEANUP] Drop redundant type annotations in `CssToAttributeConverter`

oliverklee:


src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -64,22 +64,19 @@ public function convertCssToVisualAttributes(): self
...
 
     /**
-     * Applies $styles to $node.
+     * Applies `$styles` to `$node`.
      *
      * This method maps CSS styles to HTML attributes and adds those to the
      * node.
      *
      * @param array<string, string> $styles the new CSS styles taken from the global styles to be applied to this node
-     * @param \DOMElement $node node to apply styles to
```
JakeQZ, line 82: Agree the `@param` description here is redundant.

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -94,10 +91,6 @@ private function mapCssToHtmlAttributes(array $styles, \DOMElement $node): void
      * Tries to apply the CSS style to $node as an attribute.
```
JakeQZ: As this change is adding backticks for parameter names in the main description elsewhere, I suggest we change this too:

```suggestion
     * Tries to apply the CSS style to `$node` as an attribute.
```

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -94,10 +91,6 @@ private function mapCssToHtmlAttributes(array $styles, \DOMElement $node): void
...
      *
      * This method maps a CSS rule to HTML attributes and adds those to the node.
-     *
-     * @param string $property the name of the CSS property to map
-     * @param string $value the value of the style rule to map
-     * @param \DOMElement $node node to apply styles to
```
JakeQZ, line 100: Agree these are self-explanatory from the parameter names.

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -109,11 +102,7 @@ private function mapCssToHtmlAttribute(string $property, string $value, \DOMElem
     /**
      * Looks up the CSS property in the mapping table and maps it if it matches the conditions.
      *
-     * @param string $property the name of the CSS property to map
-     * @param string $value the value of the style rule to map
-     * @param \DOMElement $node node to apply styles to
```
JakeQZ, line 114: Ditto.

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -109,11 +102,7 @@ private function mapCssToHtmlAttribute(string $property, string $value, \DOMElem
...
-     *
-     * @return bool true if the property can be mapped using the simple mapping table
+     * @return bool whether the property can be mapped using the simple mapping table
```
JakeQZ, line 105: Agree this is an improvement.

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -134,10 +123,6 @@ private function mapSimpleCssProperty(string $property, string $value, \DOMEleme
 
     /**
      * Maps CSS properties that need special transformation to an HTML attribute.
-     *
-     * @param string $property the name of the CSS property to map
-     * @param string $value the value of the style rule to map
-     * @param \DOMElement $node node to apply styles to
```
JakeQZ, line 140: Agree; these are also self-explanatory.

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -160,10 +145,6 @@ private function mapComplexCssProperty(string $property, string $value, \DOMElem
         }
     }
 
-    /**
-     * @param \DOMElement $node node to apply styles to
-     * @param string $value the value of the style rule to map
-     */
```
JakeQZ: Both these contain information which is not obvious from the method or parameter names.  So I think either they should be retained, or a description for the method should be added.

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -196,10 +172,6 @@ private function mapWidthOrHeightProperty(\DOMElement $node, string $value, stri
         $node->setAttribute($property, $number);
     }
 
-    /**
-     * @param \DOMElement $node node to apply styles to
-     * @param string $value the value of the style rule to map
-     */
```
JakeQZ: And these.

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -178,11 +159,6 @@ private function mapBackgroundProperty(\DOMElement $node, string $value): void
         $node->setAttribute('bgcolor', $first);
     }
 
-    /**
-     * @param \DOMElement $node node to apply styles to
-     * @param string $value the value of the style rule to map
-     * @param string $property the name of the CSS property to map
-     */
```
JakeQZ: Ditto for these.

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -212,10 +184,6 @@ private function mapMarginProperty(\DOMElement $node, string $value): void
         }
     }
 
-    /**
-     * @param \DOMElement $node node to apply styles to
-     * @param string $value the value of the style rule to map
-     */
```
JakeQZ: And these.

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -241,8 +204,6 @@ private function isTableOrImageNode(\DOMElement $node): bool
      * Parses a shorthand CSS value and splits it into individual values.  For example: `padding: 0 auto;` - `0 auto` is
      * split into top: 0, left: auto, bottom: 0, right: auto.
      *
-     * @param string $value a CSS property value with 1, 2, 3 or 4 sizes
```
JakeQZ: This `@param` description also has information which is not given elsewhere.

JakeQZ (changes requested):
There are quite a few `@param` descriptions that should be retained, as they provide information which is not otherwise given.  (Alternatively, a method description could be added as a replacement, to provide that information, but it would be simpler to just keep the `@param` descriptions as they are.)

Note that some of the `@param` descriptions are redundant, and should be removed.

Also, as this change is adding backticks in the general description where the DocBlock is being modified, there is an additional pair of backticks that should be added.

src/HtmlProcessor/CssToAttributeConverter.php
oliverklee: Done in #1435.

PR #1434, opened 2025-08-03
[CLEANUP] Drop redundant type annotations in `HtmlPruner`

oliverklee:


src/HtmlProcessor/HtmlPruner.php
```diff
@@ -98,11 +97,6 @@ private function removeClassesFromElements(\DOMNodeList $elements, array $classe
         }
     }
 
-    /**
-     * Removes the `class` attribute from each element in `$elements`.
```
JakeQZ, line 102: Agree the method name is self-explanatory,

src/HtmlProcessor/HtmlPruner.php
```diff
@@ -118,8 +112,6 @@ private function removeClassAttributeFromElements(\DOMNodeList $elements): void
      * Note that it does not inspect the remaining CSS, but uses information readily available from the `CssInliner`
      * instance about the CSS rules that could not be inlined.
      *
-     * @param CssInliner $cssInliner object instance that performed the CSS inlining
```
JakeQZ: I think this parameter description provides information which is not given elsewhere, and should be retained.

oliverklee: Re-added.

JakeQZ (approved):

PR #1435, opened 2025-08-04
[CLEANUP] Improve some PHPDoc comments with backticks

oliverklee:


JakeQZ (approved):
This breaks down #1433 nicely.

PR #1436, opened 2025-08-10
[CLEANUP] Improve some code comments

oliverklee:
Use more backticks and fix a typo.


src/Utilities/DeclarationBlockParser.php
```diff
@@ -43,12 +43,12 @@ public function normalizePropertyName(string $name): string
...
      *
-     *   "color: #000; font-weight: bold;"
+     *   `color: #000; font-weight: bold;`
      *
      * will be parsed into the following array:
      *
-     *   "color" => "#000"
-     *   "font-weight" => "bold"
+     *   `"color" => "#000"`
+     *   `"font-weight" => "bold"`
```
JakeQZ: As we are changing this, I think we should make the example into an actual PHP array, and use single quotes for the strings:

```suggestion
     *   `['color' => '#000', 'font-weight' => 'bold']`
```

It might be tricky to put a multiline array in backticks in the DocBlock - it certainly is as a GitHub suggestion - but if you know a good way to do so, that would be better.

JakeQZ (changes requested):
Agree we should use backticks for both CSS and PHP code samples.

I think one of the examples could be further improved while we're here.

Also, is it possible to use the `@example` PHPDoc tag for some of these?

oliverklee:
> Also, is it possible to use the `@example` PHPDoc tag for some of these?

I propose we move this to a separate PR so we can get the improvements from this PR merged quickly. Would you be willing to tackle that task?

Docs: https://docs.phpdoc.org/guide/references/phpdoc/tags/example.html

src/Utilities/DeclarationBlockParser.php
oliverklee: I'll copy this the syntax used here: https://github.com/TYPO3/typo3/blob/main/typo3/sysext/fluid/Classes/ViewHelpers/CObjectViewHelper.php

JakeQZ:
> > Also, is it possible to use the `@example` PHPDoc tag for some of these?
> 
> I propose we move this to a separate PR so we can get the improvements from this PR merged quickly. Would you be willing to tackle that task?
> 
> Docs: https://docs.phpdoc.org/guide/references/phpdoc/tags/example.html

It seems that with the `@example` tag, the example needs to be in a separate file.  I think it's better that the example is included in the PHPDoc for the method so users don't have to hunt around for it when looking at the source code.

JakeQZ (approved):

PR #1437, opened 2025-08-10
[CLEANUP] Drop redundant type annotations in `Utilities/`

oliverklee:


JakeQZ (approved):
Agree the return value descriptions are superfluous, since the methods are well-named.

PR #1442, opened 2025-08-11
[CLEANUP] Make type annotations in `CssDocument` more specific

oliverklee:
Also adapt the type annotations in `CssInliner` as needed for this.


src/Css/CssDocument.php
```diff
@@ -68,7 +68,7 @@ private function hasNestedAtRule(string $css): bool
     /**
      * Collates the media query, selectors and declarations for individual rules from the parsed CSS, in order.
      *
-     * @param array<array-key, string> $allowedMediaTypes
+     * @param array<array-key, non-empty-string> $allowedMediaTypes
```
JakeQZ: As this is an internal class, and we only call the method with the result of `array_keys()`, could this be simply `list<non-empty-string`?

oliverklee: Good catch, it can indeed! Changed.

src/Css/CssDocument.php
```diff
@@ -115,7 +115,7 @@ public function renderNonConditionalAtRules(): string
     }
 
     /**
-     * @param array<array-key, string> $allowedMediaTypes
+     * @param array<array-key, non-empty-string> $allowedMediaTypes
```
JakeQZ: We can also change this one, which is only called by `getStyleRulesData()`.

```suggestion
     * @param list<non-empty-string> $allowedMediaTypes
```

JakeQZ (changes requested):
We can also change the other `$allowedMediaTypes` `@param` to a `list`.

oliverklee:
Note: The GitHub Action failures are related to some technical problems: https://www.githubstatus.com/

JakeQZ (approved):

PR #1443, opened 2025-08-11
[CLEANUP] Simplify `CssDocument` using a guard clause

oliverklee:


JakeQZ (approved):
I concur.  An early return or exception based on unsuitable arguments works well, but otherwise the return statement should normally be at the end of the method.

PR #1445, opened 2025-08-12
[CLEANUP] Improve type annotations in `StyleRule`

oliverklee:


src/Css/StyleRule.php
```diff
@@ -35,11 +35,10 @@ public function __construct(DeclarationBlock $declarationBlock, string $containi
     }
 
     /**
-     * @return array<int, string> the selectors, e.g. `["h1", "p"]`
+     * @return array<array-key, string> the selectors, e.g. `["h1", "p"]`
```
JakeQZ: This now matches the annotation for the PHP-CSS-Parser method, which it is dependent on, but is the `array-key` redundant?

oliverklee: Yes, as far as I know: `array<array-key, string>` and `array<string>` are semantically equivalent, while the former also communicates that the array keys are not limited to only `string` or `int`.

I'd prefer to have only one of these two notations across our two projects, would be fine with either notation, with a slight preference for the shorter version.

Should I create PRs for harmonizing both our projects to the shorter version?

JakeQZ: > the former also communicates that the array keys are not limited to only `string` or `int`.

They are limited to `string` or `int` by PHP itself, unless I'm mistaken or there has been a recent language change I'm not aware of.

> I'd prefer to have only one of these two notations across our two projects, would be fine with either notation, with a slight preference for the shorter version.

The benefit of the longer version is that it implies that the programmer has considered what the array keys may be, and determined that they may be either.  The benefit of the shorter version is that, um, it's shorter, and easier to quickly grok.  I'm happy either way.

> Should I create PRs for harmonizing both our projects to the shorter version?

If, after considering the paragraph above, you are still favouring the shorter version, and would like to do so, then that would be fine by me.  However, I think it would be best to change them on a case-by-case basis, rather than do a global search-and-replace, which would lose the fine-grained information (as currently presented) of whether they key types may actually have been considered (without having to go back through the history or use `git blame`).

Another possibility (which I think I prefer most) would be to use the short form, but add a comment explaining why the keys are mixed, e.g.

```php
/**
 * @param array<Selector|string>|string $selectors
 *        As this is a public method, both `int` and `string` keys are allowed when an array is passed.
 *        The keys are ignored anyway; only the values are used (in the order that `array_values()` would return).
 */
```

WDYT?

oliverklee:
> while the former also communicates that the array keys are not limited to only `string` or `int`.

Ah, that was nonsense. My bad. What I probably meant to say was:

> while the former also communicates that the array keys are can bei either `string` or `int`, and are not limit to one of those.

src/Css/StyleRule.php
oliverklee: I've now simplified this one instance and repushed.

PR #1446, opened 2025-08-13
[CLEANUP] Tighten types and comments in `AbstractHtmlProcessor`

oliverklee:


src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -272,9 +272,9 @@ private function ensureDocumentType(string $html): string
     }
 
     /**
-     * Makes sure the document type in the passed HTML has lowercase "html".
+     * Makes sure the document type in the passed HTML has lowercase `html`.
      *
-     * @return string HTML with normalized document type
+     * @return ($html is non-empty-string ? non-empty-string : string) HTML with normalized document type
```
JakeQZ: The conditional is a neat trick I wasn't aware of.

However, this method is only called when a document type has been detected in the HTML, so it will never be passed an empty string.  Thus, I think we should add an `@param non-empty-string $html` annotation, and have `@return non-empty-string` (with the description retained).

src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -291,10 +291,13 @@ private function normalizeDocumentType(string $html): string
      * Adds a Content-Type meta tag for the charset.
      *
      * This method also ensures that there is a HEAD element.
+     *
+     * @return non-empty-string
      */
     private function addContentTypeMetaTag(string $html): string
     {
         if ($this->hasContentTypeMetaTagInHead($html)) {
+            \assert($html !== '');
```
JakeQZ: This method is only called after the presence of a `<!DOCTYPE>` has been ensured, so will never be passed an empty string.  So I think we can add an `@param non-empty-string $html` annotation, and then probably won't need the `assert()`s.  Perhaps a description for the `$html` parameter to explain why an empty string is unexpected wouldn't go amiss.

src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -318,6 +321,7 @@ private function addContentTypeMetaTag(string $html): string
         } else {
             $reworkedHtml = self::CONTENT_TYPE_META_TAG . $html;
         }
+        \assert($reworkedHtml !== '');
```
JakeQZ, line 330: If `@param non-empty-string $html` is added, this may not be needed.

JakeQZ (changes requested):
I think we can be even tighter on the `non-empty-string`s...

src/HtmlProcessor/AbstractHtmlProcessor.php
oliverklee: Done. (And added an assert which seemed to be necessary.)

src/HtmlProcessor/AbstractHtmlProcessor.php
oliverklee: Done.

src/HtmlProcessor/AbstractHtmlProcessor.php
oliverklee: This still is necessary as PHPStan currently does not know how our preg class works.

JakeQZ: I think it might struggle even if we used `preg_replace()` directly.

PR #1447, opened 2025-08-19
[CLEANUP] Tighten types in `CssToAttributeConverter`

oliverklee:


src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -64,19 +69,25 @@ public function convertCssToVisualAttributes(): self
...
     /**
      * Applies `$styles` to `$node`.
      *
-     * This method maps CSS styles to HTML attributes and adds those to the
-     * node.
+     * This method maps CSS styles to HTML attributes and adds those to the node.
      *
-     * @param array<string, string> $styles the new CSS styles taken from the global styles to be applied to this node
+     * @param array<non-empty-string, string> $styles the new CSS styles taken from the global styles
+     *        to be applied to this node
```
JakeQZ: The description could be all on its own separate line, for readability:

```suggestion
     * @param array<non-empty-string, string> $styles
     *        the new CSS styles taken from the global styles to be applied to this node
```

src/HtmlProcessor/CssToAttributeConverter.php
```diff
@@ -23,7 +23,13 @@ final class CssToAttributeConverter extends AbstractHtmlProcessor
...
      * of nodes and values.
      *
-     * @var array<string, array{attribute: string, nodes?: array<int, string>, values?: array<int, string>}>
+     * @var array<
+     *        non-empty-string, array{
+     *          attribute: non-empty-string,
+     *          nodes?: list<non-empty-string>,
+     *          values?: list<non-empty-string>
+     *        }
+     *      >
```
JakeQZ: I think it would be more readable with `array{` on a separate line:


```suggestion
     * @var array<
     *        non-empty-string,
     *        array{
     *          attribute: non-empty-string,
     *          nodes?: list<non-empty-string>,
     *          values?: list<non-empty-string>
     *        }
     *      >
```

JakeQZ (changes requested):
I've made a couple of suggestions to improve readability in the DocBlocks, but otherwise looks good.

PR #1452, opened 2025-08-21
[TASK] Streamline `DeclarationBlockParser`

oliverklee:
- simplify a condition to use a guard clause
- make some types more specific
- add a special case for parsing an empty declaration block to improve performance and avoid empty cache keys
- reformat some code


JakeQZ (approved):
Nice.

Issue #919, opened 2020-08-14
Sometimes use safe versions of PHP-internal functions


oliverklee:
When we need the function to throw an exception if a problem has occurred, we can use those.

https://github.com/thecodingmachine/safe

This should make it easier for use to make the static type checkers happy.

oliverklee:
All done now.

PR #1453, opened 2025-08-21
[TASK] Add `thecodingmachine/safe`

oliverklee:
Safe-PHP https://github.com/thecodingmachine/safe provides rewrites of PHP functions to throw an exception instead of returning `false` when an error is encountered.

This will allow us to drop out custom `preg_*` wrapper class and to increase type safety in our codebase.

Also drop the PHP-CS-Fixer rule that adds a trailing backslash to calls to native PHP functions (as this would change the Safe-PHP calls back to their unsafe versions).

The actual code changes will come in subsequent commits.

Part of #919


config/php-cs-fixer.php
```diff
@@ -40,7 +40,6 @@
             'yoda_style' => ['equal' => false, 'identical' => false, 'less_and_greater' => false],
 
             // function notation
-            'native_function_invocation' => ['include' => ['@all']],
```
JakeQZ, line 43: I assume that PHP-CS-Fixer is not smart enough to see where functions are being `use`d.

JakeQZ (approved):

config/php-cs-fixer.php
oliverklee: Correct.

PR #1456, opened 2025-08-21
[TASK] Streamline `CssInliner`

oliverklee:
Closes #1425

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -151,7 +151,7 @@ final class CssInliner extends AbstractHtmlProcessor
      *          selector: string,
      *          hasUnmatchablePseudo: bool,
      *          declarationsBlock: string,
-     *          line: int
+     *          line: int<0, max>
      *      }>|null
      */
     private $matchingUninlinableCssRules = null;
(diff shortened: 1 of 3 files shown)


src/CssInliner.php
```diff
@@ -643,8 +650,8 @@ private function collateCssRules(CssDocument $parsedCss): array
         \usort(
             $cssRules['inlinable'],
             /**
-             * @param array{selector: string, line: int, ...} $first
-             * @param array{selector: string, line: int, ...} $second
+             * @param array{selector: non-empty-string, line: int} $first
+             * @param array{selector: non-empty-string, line: int} $second
```
JakeQZ: Shouldn't these have `line: int<0, max>`?  Also I'd have expected Stan to give a warning when they are passed to `sortBySelectorPrecedence()` which has that further restriction.

oliverklee: Indeed, they should. I've fixed those (and also some other occurrences).

oliverklee:
Hold on, I think `selector` should generally also be a non-empty-string.

oliverklee:
> Hold on, I think `selector` should generally also be a non-empty-string.

Done.

JakeQZ (approved):

PR #1457, opened 2025-08-21
[TASK] Use Safe-PHP functions for regexp in `CssInliner`

oliverklee:
Part of #919


src/CssInliner.php
```diff
@@ -451,8 +454,14 @@ private function normalizeStyleAttributes(\DOMElement $node): void
         $callback = static function (array $propertyNameMatches) use ($declarationBlockParser): string {
             return $declarationBlockParser->normalizePropertyName($propertyNameMatches[0]);
         };
-        $normalizedOriginalStyle = (new Preg())->throwExceptions($this->debug)
-            ->replaceCallback($pattern, $callback, $node->getAttribute('style'));
+        // The safe version is only available in "thecodingmachine/safe" for PHP >= 8.1.
+        if (\function_exists('Safe\\preg_replace_callback')) {
+            $normalizedOriginalStyle = preg_replace_callback($pattern, $callback, $node->getAttribute('style'));
+            \assert(\is_string($normalizedOriginalStyle));
```
JakeQZ, line 460: The `assert` is surely redundant when using the `Safe` version.

src/CssInliner.php
```diff
@@ -618,8 +626,10 @@ private function collateCssRules(CssDocument $parsedCss): array
             // Maybe exclude CSS selectors
             if (\count($this->excludedCssSelectors) > 0) {
                 // Normalize spaces, line breaks & tabs
-                $selectorsNormalized = \array_map(static function (string $selector) use ($preg): string {
-                    return $preg->replace('@\\s++@u', ' ', $selector);
+                $selectorsNormalized = \array_map(static function (string $selector): string {
+                    $result = preg_replace('@\\s++@u', ' ', $selector);
+                    \assert(\is_string($result));
```
JakeQZ, line 631: Ditto.

src/CssInliner.php
```diff
@@ -669,17 +679,16 @@ function (array $first, array $second): int {
...
         }
 
-        if ($preg->match('/:(?:' . self::OF_TYPE_PSEUDO_CLASS_MATCHER . ')/i', $selector) === 0) {
+        if (preg_match('/:(?:' . self::OF_TYPE_PSEUDO_CLASS_MATCHER . ')/i', $selector) === 0) {
             return false;
         }
 
-        foreach ($preg->split('/' . self::COMBINATOR_MATCHER . '/', $selector) as $selectorPart) {
+        foreach (preg_split('/' . self::COMBINATOR_MATCHER . '/', $selector) as $selectorPart) {
+            \assert(\is_string($selectorPart));
```
JakeQZ, line 691: Ditto.

src/CssInliner.php
```diff
@@ -735,14 +742,14 @@ private function getCssSelectorPrecedence(string $selector): int
...
-        $preg = (new Preg())->throwExceptions($this->debug);
         $precedence = 0;
         foreach ($this->selectorPrecedenceMatchers as $matcher => $value) {
             if (\trim($selector) === '') {
                 break;
             }
             $count = 0;
-            $selector = $preg->replace('/' . $matcher . '\\w+/', '', $selector, -1, $count);
+            $selector = preg_replace('/' . $matcher . '\\w+/', '', $selector, -1, $count);
+            \assert(\is_string($selector));
```
JakeQZ, line 752: Ditto.

src/CssInliner.php
```diff
@@ -902,8 +909,9 @@ private function removeImportantAnnotationFromNodeInlineStyle(\DOMElement $node)
         $importantStyleDeclarations = [];
         foreach ($inlineStyleDeclarations as $property => $value) {
             if ($this->attributeValueIsImportant($value)) {
-                $importantStyleDeclarations[$property]
-                    = (new Preg())->throwExceptions($this->debug)->replace('/\\s*+!\\s*+important$/i', '', $value);
+                $declaration = preg_replace('/\\s*+!\\s*+important$/i', '', $value);
+                \assert(\is_string($declaration));
```
JakeQZ, line 913: Ditto.

src/CssInliner.php
```diff
@@ -1007,37 +1015,42 @@ private function existsMatchForCssSelector(string $cssSelector): bool
...
         /** @param array<array-key, string> $matches */
         $callback = function (array $matches): string {
             return $this->replaceUnmatchableNotComponent($matches);
         };
-        $untrimmedSelectorWithoutNots = (new Preg())->throwExceptions($this->debug)
-            ->replaceCallback($pattern, $callback, ' ' . $selector);
+        // The safe version is only available in "thecodingmachine/safe" for PHP >= 8.1.
+        if (\function_exists('Safe\\preg_replace_callback')) {
+            $untrimmedSelectorWithoutNots = preg_replace_callback($pattern, $callback, ' ' . $selector);
+            \assert(\is_string($untrimmedSelectorWithoutNots));
```
JakeQZ, line 1028: Ditto.

src/CssInliner.php
```diff
@@ -1007,37 +1015,42 @@ private function existsMatchForCssSelector(string $cssSelector): bool
...
         }
 
-        $selectorParts = $preg->split(
+        $selectorParts = preg_split(
             '/(' . self::COMBINATOR_MATCHER . ')/',
             $selectorWithoutUnmatchablePseudoComponents,
             -1,
             PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY
         );
+        /** @var list<string> $selectorParts */
```
JakeQZ, line 1053: Would be nice if `Safe` used some type templating to specify return type based on argument type, but I guess that is for the future...

src/CssInliner.php
```diff
@@ -1078,11 +1091,14 @@ private function replaceUnmatchableNotComponent(array $matches): string
      */
     private function removeSelectorComponents(string $matcher, string $selector): string
     {
-        return (new Preg())->throwExceptions($this->debug)->replace(
+        $result = preg_replace(
             ['/([\\s>+~]|^)' . $matcher . '/i', '/' . $matcher . '/i'],
             ['$1*', ''],
             $selector
         );
+        \assert(\is_string($result));
```
JakeQZ, line 1099: This should not be necessary.  The `Safe` function should either return a string or throw an exception.

JakeQZ (changes requested):
The `assert`s should not be necessary.  Surely the `Safe` versions will throw an exception rather than returning something other than the asserted type - otherwise what is the point of them?

JakeQZ (changes requested):
The `assert`s are surely not needed if `Safe` does what it's supposed to do...

JakeQZ:
> The `assert`s are surely not needed if `Safe` does what it's supposed to do...

Sorry, this is a duplicate.  Had multiple tabs open and thought I hadn't finished review.  Can't remove the duplicate.

src/CssInliner.php
oliverklee: No, it's not.

The `Safe` functions make sure that errors throw exceptions instead of returning `false`. However, those functions do not provide PHPStan-compatible annotations like "if a string is provided, a string will be returned, and if an array is provided, an array will be returned".

With our own wrapper, we could do without the `assert` as our wrapper does not support arrays, but only strings.

oliverklee: So this assert is about the value being a `string` instead of an `array`, not about it not being `false`.

JakeQZ: Ah, I see.  I've seen it possible to template the type annotations, or put conditionals into them, so maybe a feature request for 'Safe'...

oliverklee: Yes, absolutely. Would you be willing to create the issue (or provide a PR)?

JakeQZ: > Yes, absolutely. Would you be willing to create the issue (or provide a PR)?

I am not sure of the syntax required, so may need some help.  I have seen such syntax (templates and conditionals), and have used the template syntax once, so am aware it's possible, but not 100% sure how, or what's best to propose...

oliverklee: I'd say we start with the methods we use in our code.

This is the documentation on conditional return types: https://phpstan.org/writing-php-code/phpdoc-types#conditional-return-types

JakeQZ: > I'd say we start with the methods we use in our code.

Agree: those for which we currently need to add an `assert`.

JakeQZ: https://github.com/thecodingmachine/safe/issues/703

PR #1462, opened 2025-08-24
[CLEANUP] Tighten type annotations for `AssertCss`

oliverklee:


oliverklee:
There are two instances where data provider items are empty strings which are used as the needle for assertions, which doesn't seem to make sense to me. I've annotated them in the change.

@JakeQZ Do you think it would sense to remove those?

tests/Unit/CssInlinerTest.php
```diff
@@ -2309,13 +2311,13 @@ private function getCssRuleDatasetsWithSelectorPseudoComponents(array $preceding
...
     /**
-     * @return DataProvider<string, array{0: string}>
+     * @return DataProvider<non-empty-string, array{0: non-empty-string}>
      */
     public function matchingSelectorWithPseudoComponentCssRuleDataProvider(): DataProvider
     {
         $datasetsWithSelectorPseudoComponents = $this->getCssRuleDatasetsWithSelectorPseudoComponents(
             [
-                'lone' => '',
+                'lone' => '', // Does this make sense? This is an empty string.
```
JakeQZ: Yes, it makes sense.  The items in this list not used directly, but are each crossed and combined with a selection of pseudo-components by `getCssRuleDatasetsWithSelectorPseudoComponents()`.

For example `lone` from this array when combined with `pseudo-class` from the data in that method results in `'lone pseudo-class' => ':hover { color: green; }'` in the final dataset, whilst `type &` when combined with `pseudo-class` results in `'type & pseudo-class' => 'a:hover { color: green; }'`.

You should be able to change the return type of `getCssRuleDatasetsWithSelectorPseudoComponents()` to `array<string, array{0: non-empty-string}>`.

It may be possible to used `DataProviders::cross()` to do the crossing, but that would result in three separate strings that would still need combining to make a single CSS string (of the form `$precedingSelectorComponent . $selectorPseudoComponent . ' { ' . $declarationsBlock . ' }'`) for the tests.

tests/Unit/CssInlinerTest.php
```diff
@@ -3650,14 +3652,14 @@ public function inlineCssNotCopiesInlinableRuleAfterImportRuleToStyleElement():
...
-     * @return DataProvider<string, array{0: string, 1: string, 2: string}>
+     * @return DataProvider<non-empty-string, array{0: string, 1: string, 2: string}>
      */
     public function provideValidAtRulesWithSurroundingCss(): DataProvider
     {
         $datasets1 = DataProvider::cross(
             $this->provideValidAtRules(),
             [
-                'alone' => ['', ''],
+                'alone' => ['', ''], // Does this make sense? This is an empty string.
```
JakeQZ: Yes, these are optional extra CSS to go before and/or after the at-rule, to confirm that the rule is still included when there is other CSS.

tests/Unit/CssInlinerTest.php
```diff
@@ -3686,7 +3688,7 @@ public function provideValidAtRulesWithSurroundingCss(): DataProvider
                     . '@font-face{font-family:"Bar Sans";src:url(/bar-sans.woff2) format("woff2")}',
                 ],
             ],
-            ['alone' => ['', '']]
+            ['alone' => ['', '']] // Does this make sense? This is an empty string.
```
JakeQZ: Ditto.  It seems when these test data were added it was felt there was already sufficient testing with surrounding CSS for other at-rules, so these only get tested on their own.  But they use the same test method which has 'before' and 'after' parameters.

JakeQZ (commented):
> There are two instances where data provider items are empty strings which are used as the needle for assertions, which doesn't seem to make sense to me. I've annotated them in the change.
> 
> @JakeQZ Do you think it would sense to remove those?

I don't see them being used as needles in assertions.  They get concatenated with other strings which are non-empty.  I've commented on each.

tests/Unit/Css/CssDocumentTest.php
```diff
@@ -365,7 +365,7 @@ public function provideCssThatMustPrecedeStyleRules(): array
     /**
      * @test
      *
-     * @param string $atRuleCss
+     * @param non-empty-string $atRuleCss
      * @param string $cssBefore
```
JakeQZ: This could be removed.

tests/Unit/CssInlinerTest.php
```diff
@@ -1750,7 +1750,7 @@ public function inlineCssMatchesRuleAfterUnneededCssThing(string $unneededCss):
     /**
      * Data provider for media rules.
      *
-     * @return string[][]
+     * @return array<non-empty-string, list<non-empty-string>>
```
JakeQZ:
```suggestion
     * @return array<non-empty-string, array{0: non-empty-string}>
```

JakeQZ (changes requested):
I think a type specs can be further tightended, and a redundant one can be removed.

PR #1469, opened 2025-09-14
[CLEANUP] Refactor preg usage in `CssInliner`

oliverklee:
This is in preparation for using the safe versions of the preg functions.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -1016,13 +1016,13 @@ final class CssInliner extends AbstractHtmlProcessor
         $callback = function (array $matches): string {
             return $this->replaceUnmatchableNotComponent($matches);
         };
-        $selectorWithoutNots = (new Preg())->throwExceptions($this->debug)
+        $untrimmedSelectorWithoutNots = (new Preg())->throwExceptions($this->debug)
             ->replaceCallback($pattern, $callback, ' ' . $selector);
-        $trimmedSelectorWithoutNots = \ltrim($selectorWithoutNots);
+        $selectorWithoutNots = \ltrim($untrimmedSelectorWithoutNots);
 
         $selectorWithoutUnmatchablePseudoComponents = $this->removeSelectorComponents(
             ':(?!' . self::PSEUDO_CLASS_MATCHER . '):?+[\\w\\-]++(?:\\([^\\)]*+\\))?+',
-            $trimmedSelectorWithoutNots
+            $selectorWithoutNots
         );
 
         if ($preg->match(
@@ -1038,7 +1038,6 @@ final class CssInliner extends AbstractHtmlProcessor
             -1,
             PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY
         );
-        /** @var list<string> $selectorParts */
 
         return \implode('', \array_map(
             function (string $selectorPart): string {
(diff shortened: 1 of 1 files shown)


src/CssInliner.php
```diff
@@ -1012,39 +1011,40 @@ private function removeUnmatchablePseudoComponents(string $selector): string
...
-            ' ' . $selector
-        ));
+        $pattern = '/([\\s>+~]?+):not(\\([^()]*+(?:(?2)[^()]*+)*+\\))/i';
+        /** @param array<array-key, string> $matches */
+        $callback = function (array $matches): string {
+            return $this->replaceUnmatchableNotComponent($matches);
+        };
+        $selectorWithoutNots = (new Preg())->throwExceptions($this->debug)
+            ->replaceCallback($pattern, $callback, ' ' . $selector);
+        $trimmedSelectorWithoutNots = \ltrim($selectorWithoutNots);
```
JakeQZ: Would it be better to name the new intermediate variable `$untrimmedSelectorWithoutNots`, keeping the same name for the final result variable, so that the code which uses it doesn't need to be changed?

src/CssInliner.php
```diff
@@ -1012,39 +1011,40 @@ private function removeUnmatchablePseudoComponents(string $selector): string
...
-            $preg->match(
-                '/:(?:' . self::OF_TYPE_PSEUDO_CLASS_MATCHER . ')/i',
-                $selectorWithoutUnmatchablePseudoComponents
-            )
-            === 0
-        ) {
+        if ($preg->match(
+            '/:(?:' . self::OF_TYPE_PSEUDO_CLASS_MATCHER . ')/i',
+            $selectorWithoutUnmatchablePseudoComponents
+        ) === 0) {
```
JakeQZ, line 1031: Would autoformatting the code keep it like this?

src/CssInliner.php
```diff
@@ -1012,39 +1011,40 @@ private function removeUnmatchablePseudoComponents(string $selector): string
...
             return $selectorWithoutUnmatchablePseudoComponents;
         }
+
+        $selectorParts = $preg->split(
+            '/(' . self::COMBINATOR_MATCHER . ')/',
+            $selectorWithoutUnmatchablePseudoComponents,
+            -1,
+            PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY
+        );
+        /** @var list<string> $selectorParts */
```
JakeQZ: Is this needed?  `Preg::split()` is declared as returning `array<int, string>` (and could be tightened to `list<string>`), though `array_map()` does not require a sequential array.

JakeQZ (commented):
I've checked that the functionality is unchanged, but have a few questions.

src/CssInliner.php
oliverklee: I've removed it for now.

src/CssInliner.php
oliverklee: Makes sense. I've changed the code accordingly.

src/CssInliner.php
oliverklee: Yes, indirectly: PhpStorm autoformatting changes it, and PHP-CS-Fixer then changes it back. (So the code formatting stays the same after a PhpStorm + PHP-CS-Fixer cycle).

PR #1472, opened 2025-09-16

oliverklee:


oliverklee:
> I presume the Stan baseline changes are due to the update.

Yes, exactly.

>  I wonder if the 1.x branch produces the same.

No, it will not.

> If not, perhaps we would need to advise that it won't work with earlier PHP versions, but still need to allow it so that `composer update` succeeds.

Maybe. Or we rely on the fact that if people develop on lower PHP versions, they'll see the CI failure, look at the build and see that it uses a more recent PHP version.

PR #1473, opened 2025-09-17
[TASK] Use safe preg functions in `HtmlPrunerTest`

oliverklee:
Part of #919


tests/Unit/HtmlProcessor/HtmlPrunerTest.php
```diff
@@ -394,8 +395,11 @@ public function removeRedundantClassesMinifiesClassAttributes(string $html, arra
 
         $subject->removeRedundantClasses($classesToKeep);
 
-        (new Preg())->matchAll('/class="([^"]*+)"/', $subject->renderBodyContent(), $classAttributeMatches);
+        preg_match_all('/class="([^"]*+)"/', $subject->renderBodyContent(), $classAttributeMatches);
+        self::assertIsArray($classAttributeMatches);
+        self::assertIsArray($classAttributeMatches[1]);
```
JakeQZ, line 400: I am surprised that this satisfies Stan.  Surely if offset 1 doesn't exist, a PHP notice will be generated before `assertIsArray()` is called.  I would have expected an additional `isset($classAttributeMatches[1])` test to be required.

PR #1475, opened 2025-09-19
[TASK] Use safe preg functions in `CssDocumentTest`

oliverklee:
Part of #919

### DIFF
--- a/config/phpstan-baseline.neon
+++ b/config/phpstan-baseline.neon
@@ -159,12 +159,6 @@ parameters:
 			path: ../tests/Unit/Css/CssDocumentTest.php
 
 		-
-			message: '#^Offset 0 might not exist on array\<int, string\>\|null\.$#'
-			identifier: offsetAccess.notFound
-			count: 1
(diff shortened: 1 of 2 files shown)


config/phpstan-baseline.neon
```diff
@@ -158,12 +158,6 @@ parameters:
 			count: 1
 			path: ../tests/Unit/Css/CssDocumentTest.php
 
-		-
-			message: '#^Offset 0 might not exist on array\<int, string\>\|null\.$#'
-			identifier: offsetAccess.notFound
-			count: 1
-			path: ../tests/Unit/Css/CssDocumentTest.php
-
```
JakeQZ, line 166: I think I've worked out why this error disappears, even though offset `0` still might not exist.  The variable previously could be an array or `null`.  Stan is warning about the `null` case, with a slightly misleading message, and isn't caring about the case when it is an array.  `assertIsArray()` eliminates the `null` case.  (From memory, I think the `null` case results in a fatal error, whereas the array-key-missing case merely generates a PHP warning or notice.)

JakeQZ (approved):

PR #1483, opened 2025-09-28
[TASK] Use safe preg functions in `CssConstraint`

oliverklee:
Part of #919


tests/Support/Constraint/CssConstraint.php
```diff
@@ -167,8 +167,17 @@ abstract class CssConstraint extends Constraint
      */
     protected static function getCssRegularExpressionMatcher(string $css): string
     {
-        $callback = \Closure::fromCallable([self::class, 'getCssRegularExpressionReplacement']);
-        $matcher = (new Preg())->replaceCallback(self::CSS_REGULAR_EXPRESSION_PATTERN, $callback, $css);
+        $callback = static function (array $matches): string {
+            /** @var array<int<0, max>, string> $matches $matches */
```
JakeQZ: Duplicate `$matches`.

```suggestion
            /** @var array<int<0, max>, string> $matches */
```

JakeQZ (changes requested):
There's a duplicate parameter name in a DocBlock.

oliverklee:
> This doesn't seem to have included the correction.

Very strange … possibly I forgot to click the button or something.

I've re-applied your suggestion (again) now.

PR #1485, opened 2025-09-28
[CLEANUP] Refactor preg functional with callback

oliverklee:
This is in preparation for using the safe version of `preg_replace_callback` and `preg_callback`.


JakeQZ (commented):
I looked briefly at the issue in #1483.

It seems like a bug in PHPStan.  A `Closure` should be synonymous with a `callable`, and Stan doesn't complain when they are untyped.  The types prescribed in the 'Safe' version appear to be consistent with the `Closure` wrapper.

I would prefer to keep the private method private, live with the PHPStan warning for now, and perhaps log an issue with PHPStan.  A very brief internet search did not turn up relevant results, but I only looked at the first page.

The `Closure` wrapper was needed because it is only possible to pass a private method as a `callable` to in-built functions.  Perhaps a later version of PHP has resolved that problem (IDK).

WDYT?

oliverklee:
Okay, I've changed this back to `Closure::fromCallable` for now. Let's see how this pans out when I switch the calls to use the safe preg functions.

PR #1486, opened 2025-10-12

oliverklee:


oliverklee:
I'd like to add https://github.com/phpstan/phpstan-deprecation-rules, and for the latest version of that tool, we'll need to update PHPStan first.

JakeQZ (approved):

PR #1488, opened 2025-10-25
[CLEANUP] Tighten type annotations in `CssDocumentTest`

oliverklee:


tests/Unit/Css/CssDocumentTest.php
```diff
@@ -380,7 +375,7 @@ public function rendersValidNonConditionalAtRule(string $atRuleCss, string $cssB
     }
 
     /**
-     * @return array<non-empty-string, list<non-empty-string>>
+     * @return array<non-empty-string, array{0: non-empty-string}>
```
JakeQZ: This can provide one or two non-empty strings.  I guess there's a syntax like `?1:` for the optional element but don't know offhand.

tests/Unit/Css/CssDocumentTest.php
```diff
@@ -566,7 +571,7 @@ public function provideInvalidNonConditionalAtRuleWhichCausesException(): array
     }
 
     /**
-     * @return array<string, array<int, string>>
+     * @return array<non-empty-string, array{0: non-empty-string}>
```
JakeQZ: This also may return an array with one or two strings.

JakeQZ (changes requested):
Two of the data providers return datasets with either one or two strings.  (The tests methods using them have a default value for the second parameter.)

JakeQZ (approved):

PR #1490, opened 2025-10-26
[CLEANUP] Tighten type annotations in `StyleRuleTest`

oliverklee:


JakeQZ (approved):
The rendered version of the selectors given by `provideSelectors` does not appear to be used in any test (there isn't a method `getSelectorsAsText` akin to `getDeclarationAsText` - which should probably have been named `getDeclarationsAsText`).

oliverklee:
I've taken care of these two findings in #1493 and #1494.

PR #1495, opened 2025-10-29
[TASK] Use safe preg functions in `CssDocument`

oliverklee:
Part of #919


JakeQZ (approved):
I concur that a changelog entry is warranted even though the class itself is `@internal`.

oliverklee:
> I concur that a changelog entry is warranted even though the class itself is `@internal`.

Yes - as it's in production code.

PR #1498, opened 2025-11-02
[CLEANUP] Tighten type annotations in `AbstractHtmlProcessorTest`

oliverklee:


tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -718,7 +718,7 @@ public function addsMissingHtml5DocumentType(): void
     }
 
     /**
-     * @return array<string, array<int, string>>
+     * @return array<non-empty-string, array{0: non-empty-string, 1?: non-empty-string}>
```
JakeQZ: The datasets have only one string element.  (Some are concatenated over two lines because they are too long for the code formatting.)


```suggestion
     * @return array<non-empty-string, array{0: non-empty-string}>
```

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -996,15 +996,15 @@ public function selfClosingTagDataProvider(): array
...
              *
-             * @return array{0: string, 1: string, 2:string}
+             * @return array{0: non-empty-string, 1: non-empty-string, 2:non-empty-string}
              */
             static function (array $dataset): array {
                 $updatedDataset = $dataset;
                 \array_unshift($updatedDataset, \str_replace('/>', '>', $dataset[0]));
 
-                /** @var array{0: string, 1: string, 2:string} $updatedDataset */
+                /** @var array{0: non-empty-string, 1: non-empty-string, 2:non-empty-string} $updatedDataset */
```
JakeQZ: While we're here, we should probably add a space after `2:` for consistency.


```suggestion
                /** @var array{0: non-empty-string, 1: non-empty-string, 2: non-empty-string} $updatedDataset */
```

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -1015,12 +1015,12 @@ static function (array $dataset): array {
...
-     * @param array<string, array<int, string>> $rightDatasets
+     * @param array<non-empty-string, list<non-empty-string>> $leftDatasets
+     * @param array<non-empty-string, list<non-empty-string>> $rightDatasets
      *
-     * @return string[][] The new datasets comprise the first dataset from the left-hand side with each of the datasets
-     *         from the right-hand side, and the each of the remaining datasets from the left-hand side with the first
-     *         dataset from the right-hand side.
+     * @return array<non-empty-string, list<non-empty-string>> The new datasets comprise the first dataset from the
+     *         left-hand side with each of the datasets from the right-hand side, and each of the remaining datasets
+     *         from the left-hand side with the first dataset from the right-hand side.
```
JakeQZ, line 1023: As this is a generic method, I don't think there should be a requirement that the string values in the datasets are non-empty (even if in the actual use cases they all happen to be).

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -1042,7 +1042,7 @@ public static function joinDatasets(array $leftDatasets, array $rightDatasets):
     }
 
     /**
-     * @return string[][]
+     * @return array<non-empty-string, list<non-empty-string>>
```
JakeQZ, line 1045: This will in fact return datasets which are an array of exactly three non-empty strings.  I don't know if it would be possible to use an `assert` on the return value of the generic method in order to further tighten the return type here...

JakeQZ (changes requested):
I found a couple of minor issues, and have a couple of queries.  Otherwise all good.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
oliverklee: If we change this, methods relying on this method returning non-empty strings will break. I propose we widen these type annotations when we need to.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
oliverklee: I've tried, but failed …

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
JakeQZ: **If** and when we need to.  I suspect we won't, particularly as you were looking at replacing this with functionality from the Data Providers library.

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
JakeQZ: I can't think of and don't know of a way to do so either.  Just wondered if it might be possible with a flash of inspiration :)

tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -474,8 +474,8 @@ public function provideContentWithHtmlTag(): array
     }
 
     /**
-     * @return array<string, list<non-empty-string>> The second element of each dataset is optional, and is the expected
-     * normalization of the `<html>` tag, if different.
+     * @return array<non-empty-string, array{0: non-empty-string, 1?: non-empty-string}> The second element of each
+     * dataset is optional, and is the expected normalization of the `<html>` tag, if different.
```
JakeQZ, line 478: For future reference, when changing DocBlock types with a description that spans multiple lines, we can move the description to start on the next line, so that if we end up changing the type again later, there'll be less 'diff', and we won't have to further adjust the decription to fit the line length limitation.  But I don't see much point in re-editing this now for this PR :)

PR #1500, opened 2025-11-05
Enhance regex to match custom properties

joostdekeijzer:
CSS custom property names must start with "--" and contain extra characters which may be any letter, number, underscore or hyphen.

This fixes eg. custom property name "--base-size-16: 1rem;".

There are other options for the regex. Eg. `'/^([A-Za-z\\-][A-Za-z0-9_\\-]*)\\s*:\\s*(.+)$/s'` should work (will be less precise) and https://github.com/MyIntervals/emogrifier/blob/cd4d1724c337dc19eb768bc026c6a0be89b4c079/src/CssInliner.php#L452 has a different take.

### DIFF
--- a/src/Utilities/DeclarationBlockParser.php
+++ b/src/Utilities/DeclarationBlockParser.php
@@ -82,7 +82,7 @@ final class DeclarationBlockParser
         $properties = [];
         foreach ($declarations as $declaration) {
             $matches = [];
-            if (preg_match('/^(-?[a-zA-Z_][a-zA-Z_0-9\-]++|--[a-zA-Z_0-9\-]++)\\s*+:\\s*(.++)$/s', \trim($declaration), $matches) === 0) {
+            if (preg_match('/^(-?+[a-zA-Z_][a-zA-Z_0-9\\-]++|--[a-zA-Z_0-9\\-]++)\\s*+:\\s*(.++)$/s', \trim($declaration), $matches) === 0) {
                 continue;
             }
 
--- a/tests/Unit/Utilities/DeclarationBlockParserTest.php
+++ b/tests/Unit/Utilities/DeclarationBlockParserTest.php
@@ -156,6 +156,38 @@ final class DeclarationBlockParserTest extends TestCase
                 'string' => '--Base_size-4u: normal;',
                 'array' => ['--Base_size-4u' => 'normal'],
             ],
+            'specification test underscore allowed' => [
+                'string' => '_allowed: normal;',
+                'array' => ['_allowed' => 'normal'],
+            ],
+            'specification test hyphen underscore allowed' => [
+                'string' => '-_allowed: normal;',
+                'array' => ['-_allowed' => 'normal'],
+            ],
+            'specification test double hyphen underscore allowed' => [
+                'string' => '--_allowed: normal;',
+                'array' => ['--_allowed' => 'normal'],
+            ],
+            'specification test double underscore allowed' => [
+                'string' => '__allowed: normal;',
+                'array' => ['__allowed' => 'normal'],
+            ],
+            'specification test number not allowed' => [
+                'string' => '2not-allowed: unset;',
+                'array' => [],
+            ],
+            'specification test hyphen number not allowed' => [
+                'string' => '-2not-allowed: unset;',
+                'array' => [],
+            ],
+            'specification test double hyphen number allowed' => [
+                'string' => '--2allowed: normal;',
+                'array' => ['--2allowed' => 'normal'],
+            ],
+            'specification test backslash not allowed' => [
+                'string' => 'not\\allowed: unset;',
+                'array' => [],
+            ],
             'declaration using custom property' => [
                 'string' => 'color: var(--text-color);',
                 'array' => ['color' => 'var(--text-color)'],


oliverklee:
Hi @joostdekeijzer, thank you very much for this contribution!

Could you please add a unit test (or multiple tests) that test your change, i.e., that fail without your change and that pass with your change? This ensures that your changes work as intended, and that further changes down the road won't break the changed functionality.

joostdekeijzer:
Sorry for the messy commits, but tests added that fail on current main and succeed on PR

src/Utilities/DeclarationBlockParser.php
```diff
@@ -82,7 +82,7 @@ public function parse(string $declarationBlock): array
         $properties = [];
         foreach ($declarations as $declaration) {
             $matches = [];
-            if (preg_match('/^([A-Za-z\\-]+)\\s*:\\s*(.+)$/s', \trim($declaration), $matches) === 0) {
+            if (preg_match('/^([A-Za-z\\-]+|--[A-Za-z0-9_\\-]+)\\s*:\\s*(.+)$/s', \trim($declaration), $matches) === 0) {
```
JakeQZ: As we're changing this regex, and introducing alternation, I think it would be prudent at this time to also add the [possessive quantifier](https://www.regular-expressions.info/possessive.html).  This tells the regex engine not to store backtracking information (when we know it is not needed) which can grow exponentially to catastrophic failure, and is an optimization anyway.  (The alternation I think would increase the amount of backtracking information otherwise.)

I have not tested this, but I think it will work.  Let us know if otherwise:


```suggestion
            if (preg_match('/^([A-Za-z\\-]++|--[A-Za-z0-9_\\-]++)\\s*+:\\s*+(.++)$/s', \trim($declaration), $matches) === 0) {
```

tests/Unit/Utilities/DeclarationBlockParserTest.php
```diff
@@ -30,6 +30,10 @@ public function providePropertyNameAndExpectedNormalization(): array
                 'name' => '--text-color',
                 'expect' => '--text-color',
             ],
+            'custom property with numbers' => [
+                'name' => '--base-size-16',
+                'expect' => '--base-size-16',
+            ],
```
JakeQZ, line 36: The code change also allows underscores in the property name, so I think that should be tested too.

tests/Unit/Utilities/DeclarationBlockParserTest.php
```diff
@@ -140,6 +144,10 @@ public function provideDeclratationBlockAsStringAndArray(): array
                 'string' => '--text-color: green;',
                 'array' => ['--text-color' => 'green'],
             ],
+            'custom property with numbers definition' => [
+                'string' => '--base-size-16: 1rem;',
+                'array' => ['--base-size-16' => '1rem'],
+            ],
```
JakeQZ, line 154: And a test with underscores here as well.

JakeQZ:
> (We should make those common - good find, will log issue for that.)

#1501

joostdekeijzer:
I'll test your suggestion, I did not know the possessive quantifier.

About the regex: there is an overlap: eg. "--just-letters" matches both alternations. Not really a problem I guess.

Since I don't really know what the original should match I've left it as-is. If it's "only" regular properties the draft says:

From: https://drafts.csswg.org/css-syntax-3/#example-0ffd6a46
>Property names and [at-rule](https://drafts.csswg.org/css-syntax-3/#at-rule) names are always [ident sequences](https://drafts.csswg.org/css-syntax-3/#ident-sequence), which have to start with an [ident-start code point](https://drafts.csswg.org/css-syntax-3/#ident-start-code-point), two hyphens, or a hyphen followed by an ident-start code point, and then can contain zero or more [ident code points](https://drafts.csswg.org/css-syntax-3/#ident-code-point). You can include any [code point](https://infra.spec.whatwg.org/#code-point) at all, even ones that CSS uses in its syntax, by [escaping](https://drafts.csswg.org/css-syntax-3/#escape-codepoint) it.

(which is hard to read) but I do find that it should be illegal to start with a number (see https://drafts.csswg.org/css-syntax-3/#ident-start-code-point)

Properties are: `-?[a-zA-Z_][a-zA-Z_0-9\-]++` (I must test :)
Custom properties are: `/--[a-zA-Z_0-9\-]++/`

The (default)properties regex skips most of the https://drafts.csswg.org/css-syntax-3/#non-ascii-ident-code-point defined.

src/Utilities/DeclarationBlockParser.php
```diff
@@ -82,7 +82,7 @@ public function parse(string $declarationBlock): array
         $properties = [];
         foreach ($declarations as $declaration) {
             $matches = [];
-            if (preg_match('/^([A-Za-z\\-]+)\\s*:\\s*(.+)$/s', \trim($declaration), $matches) === 0) {
+            if (preg_match('/^(-?[a-zA-Z_][a-zA-Z_0-9\-]++|--[a-zA-Z_0-9\-]++)\\s*+:\\s*(.++)$/s', \trim($declaration), $matches) === 0) {
```
JakeQZ: The hypens (`-`) need escaping with a double backslash, to be escaped for the PHP string and then for the regex.  The possessive quantifier can also be applied to the optional vendor prefix.


```suggestion
            if (preg_match('/^(-?+[a-zA-Z_][a-zA-Z_0-9\\-]++|--[a-zA-Z_0-9\\-]++)\\s*+:\\s*(.++)$/s', \trim($declaration), $matches) === 0) {
```

Also, if changing the left-hand-side of the alternation to improve matching of standard and vendor property names, there should probably be another test added to confirm that underscores and numbers are now allowed in these (though I'm not aware of any real-world examples that use them).

JakeQZ (changes requested):
The regex looks good, though the backslashes need escaping for the PHP string, and an additional possessive quantifier could be added.

We should have an additional test to cover the change for matching standard or vendor properties (i.e. now allowing underscores and numbers).

src/Utilities/DeclarationBlockParser.php
joostdekeijzer: I'm a bit confused regarding the double backslash before the hyphen. I thought this was not needed in single-quoted text strings?
I've added the second backslashes but I've also added some extra unit tests (including backslash property names) and both with and without extra backslash give the same results?

JakeQZ (changes requested):
I like the additional tests, particularly that they cover rejection of invalid property names as well as acceptance of valid ones.

However, the array keys for the test data need to be unique (otherwise only the last one with the same name is included).  Names for keys like "standard property starting with underscore", "standard property starting with two underscores", etc., should suffice.

> I'm a bit confused regarding the double backslash before the hyphen. I thought this was not needed in single-quoted text strings?

I think you are right that they are not strictly necessary.  However, having them helps avoid ambiguity, and makes the intention clear.  It may amount to personal preference, but elsewhere in the project we always escape literal backslashes.  (They're not necessary in double-quoted strings either, but with so many escape sequences defined, and possibly more in the future, it becomes risky leaving them out.)

JakeQZ (approved):
Looks good now.  Thanks for contributing <3

tests/Unit/Utilities/DeclarationBlockParserTest.php
```diff
@@ -140,6 +148,50 @@ public function provideDeclratationBlockAsStringAndArray(): array
...
+                'array' => ['--2allowed' => 'normal'],
+            ],
+            'specification test backslash not allowed' => [
+                'string' => 'not\allowed: unset;',
+                'array' => [],
+            ],
+            'specification test double backslash not allowed' => [
+                'string' => 'not\\allowed: unset;',
+                'array' => [],
+            ],
```
JakeQZ, line 190: As discussed, these equate to the same, but not escaping the `\` also fails an integration check.  I suggest simply removing the one that breaks the CI:

```suggestion
            'specification test backslash not allowed' => [
                'string' => 'not\\allowed: unset;',
                'array' => [],
            ],
```

JakeQZ (changes requested):
The string syntax with the unescaped backslash in the test data is failing an integration check.  I suggest simply removing it, as it is a effectively a duplicate.

Otherwise all looks good.

joostdekeijzer:
> The string syntax with the unescaped backslash in the test data is failing an integration check. I suggest simply removing it, as it is a effectively a duplicate.
> 
> Otherwise all looks good.

I was wondering myself what an escaped "a" would be in single quote PHP string 😄 PHP itself did not seem to mind but removed it now.

JakeQZ (approved):
> I was wondering myself what an escaped "a" would be in single quote PHP string 😄 PHP itself did not seem to mind but removed it now.

[The documentation for strings](https://www.php.net/manual/en/language.types.string.php) says that unrecognized escape sequences will result in the backslash being included.  (Other languages behave differently: e.g. in C/C++ an unrecognized escape sequence results in a compilation error.)

Thanks again for contributing.

JakeQZ:
I just realized there should have been a changlog entry for this change.  I've created #1502 for that, which @oliverklee will review, so there's nothing more you need to do, @joostdekeijzer (though you can now delete the branch in your fork).

PR #1503, opened 2025-11-11
[TASK] Use safe preg functions in `CssVariableEvaluator`

oliverklee:
Part of #919

### DIFF
--- a/config/phpstan-baseline.neon
+++ b/config/phpstan-baseline.neon
@@ -37,13 +37,13 @@ parameters:
 			path: ../src/HtmlProcessor/CssToAttributeConverter.php
 
 		-
-			message: '#^Parameter \#2 \$callback of function Safe\\preg_replace_callback expects callable\(array\)\: string, array\{\$this\(Pelago\\Emogrifier\\HtmlProcessor\\CssVariableEvaluator\), ''getPropertyValueRep…''\} given\.$#'
+			message: '#^Parameter \#2 \$callback of function Safe\\preg_replace_callback expects callable\(array\)\: string, Closure\(array\<int, string\>\)\: string given\.$#'
 			identifier: argument.type
 			count: 1
 			path: ../src/HtmlProcessor/CssVariableEvaluator.php
 
 		-
-			message: '#^Parameter \#2 \$callback of function preg_replace_callback expects callable\(array\<string\>\)\: string, array\{\$this\(Pelago\\Emogrifier\\HtmlProcessor\\CssVariableEvaluator\), ''getPropertyValueRep…''\} given\.$#'
+			message: '#^Parameter \#2 \$callback of function preg_replace_callback expects callable\(array\<string\>\)\: string, Closure\(array\<int, string\>\)\: string given\.$#'
 			identifier: argument.type
 			count: 1
 			path: ../src/HtmlProcessor/CssVariableEvaluator.php
--- a/src/HtmlProcessor/CssVariableEvaluator.php
+++ b/src/HtmlProcessor/CssVariableEvaluator.php
@@ -54,10 +54,8 @@ final class CssVariableEvaluator extends AbstractHtmlProcessor
      * Callback function for {@see replaceVariablesInPropertyValue} performing regular expression replacement.
      *
      * @param array<int, string> $matches
-     *
-     * @internal
      */
-    public function getPropertyValueReplacement(array $matches): string
+    private function getPropertyValueReplacement(array $matches): string
     {
         $variableName = $matches[1];
 
@@ -125,7 +123,7 @@ final class CssVariableEvaluator extends AbstractHtmlProcessor
                 \\)
             /x';
 
-        $callable = [$this, 'getPropertyValueReplacement'];
+        $callable = \Closure::fromCallable([$this, 'getPropertyValueReplacement']);
         // The safe version is only available in "thecodingmachine/safe" for PHP >= 8.1.
         if (\function_exists('Safe\\preg_replace_callback')) {
             $result = preg_replace_callback($pattern, $callable, $propertyValue);


JakeQZ:
We can't use the `Safe` version of `preg_replace_callback` universally until we have PHP 8.1 as a minimum version.  It was [added late to the library](https://github.com/thecodingmachine/safe/issues/223) and [not included until the 3.0 release](https://github.com/thecodingmachine/safe/releases/tag/v3.0.0).

What you have done elsewhere is use a `function_exists()` test, falling back to the standard PHP function, as in #1457.  People using PHP<8.1 will have to live with a possible failure.  Those PHP versions are EOL anyway, and it would be good to eliminate the `Preg` class since we have found a maintained alternative.

It might be worth our adding a `Pelago\Emogrifier\preg_replace_callback()` function in the meantime to handle the situation, so that the main code isn't cluttered with `function_exists()` tests.  WDYT?

oliverklee:
> It might be worth our adding a `Pelago\Emogrifier\preg_replace_callback()` function in the meantime to handle the situation, so that the main code isn't cluttered with `function_exists()` tests. WDYT?

I'd prefer living with these `function_exists` for some more time, avoiding the work of creating and maintaining a utility class, and then feel even more satisfied when we can drop these switches after dropping support for PHP < 8.1 hopefully quite soon.

Would you be okay with that approach?

src/HtmlProcessor/CssVariableEvaluator.php
```diff
@@ -120,8 +122,20 @@ private function replaceVariablesInPropertyValue(string $propertyValue): string
                     )?+
                 \\)
             /x';
-        $callable = \Closure::fromCallable([$this, 'getPropertyValueReplacement']);
-        return (new Preg())->replaceCallback($pattern, $callable, $propertyValue);
+
+        $callable = function (array $matches): string {
+            /** @var array<int, string> $matches */
+            return $this->getPropertyValueReplacement($matches);
+        };
```
JakeQZ: Having this extra function seems redundant.  Is it not possible to simply have

```php
$callable = [$this, 'getPropertyValueReplacement'];
```

I can't recall offhand why I used `Closure::fromCallable`, but there must have been a reason, and it probably related to Stan.  And I guess that the `Safe` functions not accepting a `Closure` also relates to Stan.

CHANGELOG.md
```diff
@@ -19,7 +19,8 @@ Please also have a look at our
 ### Fixed
 
 - Allow numbers and underscores in CSS property names (#1500)
-- Use safe version of some PHP functions (#1457, #1492, #1495, #1496, #1499)
+- Use safe version of some PHP functions
+  (#1457, #1492, #1495, #1496, #1499, #5103)
```
JakeQZ: Typo:

```suggestion
  (#1457, #1492, #1495, #1496, #1499, #1503)
```

JakeQZ (changes requested):
I have a query about adding an additional wrapper function for a method that already has the correct signature to be passed to `preg_replace_callback`.

There's a typo in the changelog.

JakeQZ:
> > It might be worth our adding a `Pelago\Emogrifier\preg_replace_callback()` function in the meantime to handle the situation, so that the main code isn't cluttered with `function_exists()` tests. WDYT?
> 
> I'd prefer living with these `function_exists` for some more time, avoiding the work of creating and maintaining a utility class,

It wouldn't be a utility class, it would be a utility function, and it would simply encapsulate the `function_exists()` test to avoid it being duplicated throughout the code.  I.e.:

```php
namespace Pelago\Emogrifier;

function preg_replace_callback($pattern, callable $callback, $subject, int $limit = -1, int &$count = null, int $flags = 0)
{
    if (\function_exists('Safe\\preg_replace_callback')) {
        $result = \Safe\preg_replace_callback($pattern, $callable, $propertyValue);
    } else {
        $result = \preg_replace_callback($pattern, $callable, $propertyValue);
    }
}
```

I doubt it would require any maintenance.

> and then feel even more satisfied when we can drop these switches after dropping support for PHP < 8.1 hopefully quite soon.

As you know, I'm keen to retain compatibility with the EOL PHP versions unless we have a compelling reason to drop them.  It is up to the end user whether they continue to use them; in some situations upgrading is problematic.  So I don't see dropping support for PHP < 8.1 as "quite soon" at all.

> Would you be okay with that approach?

I don't mind having these branches in each piece of code that uses `preg_replace_callback()`.  I just thought that it could be tidied up for now, as outlined above.

oliverklee:
> It wouldn't be a utility class, it would be a utility function,

I see.

I don't know how Composer autoloading would works for functions instead of classes. Do you have any experience with that?

src/HtmlProcessor/CssVariableEvaluator.php
oliverklee: I tried … this would break `Safe\preg_replace_callback`:

```
TypeError : Safe\preg_replace_callback(): Argument #2 ($callback) must be of type callable, array given, called in /home/klee/src/php/emogrifier/src/HtmlProcessor/CssVariableEvaluator.php on line 133
```

With the native `preg_replace_callback`, this works. I don't quite understand why this doesn't work with `Safe\preg_replace_callback`, which uses `callable` as a native type declaration.

oliverklee: Ah, it works if we change the method from `private` to `public`. I'd prefer that as it would allow us to drop the anonymous function.

Would you be okay if I changed that in a follow-up PR (and also made the callback methods `@internal`)?

oliverklee: > Ah, it works if we change the method from `private` to `public`.

I think I understand why: When we are using the native `preg_replace_callback`, we are within the scope of the class with the method, whereas with ` Safe\preg_replace_callback`, the callback is used (or passed on) outside of the scope of the original class, where `private` methods are not available.

JakeQZ:
> I don't know how Composer autoloading would works for functions instead of classes. Do you have any experience with that?

Me neither.  But the `Safe` library manages it, so we could just replicate whatever they do.

src/HtmlProcessor/CssVariableEvaluator.php
JakeQZ: > Ah, it works if we change the method from `private` to `public`. I'd prefer that as it would allow us to drop the anonymous function.

Yes, I remember now.  `private` methods can be passed as `callable`s to internal PHP functions, but not to user or library functions or methods.  It's one of the quirks of PHP that will hopefully be resolved in a future release.

`Closure::fromCallable()` solved this issue when using the internal `Preg` wrapper class, but I guess isn't working with `Safe` for some reason.  I think I got that from a Stack Overflow Q/A, but can't immediately find the source.

> Would you be okay if I changed that in a follow-up PR (and also made the callback methods `@internal`)?

I'd actually rather you made that change in this PR.  I don't see the two as particularly separable.

JakeQZ (changes requested):
I think we should change the actual callback method from `private` to `public` `@internal` as part of this PR.  Niether change stands up alone.

src/HtmlProcessor/CssVariableEvaluator.php
oliverklee: Done and repushed.

config/phpstan-baseline.neon
```diff
@@ -36,6 +36,18 @@ parameters:
...
+			identifier: argument.type
+			count: 1
+			path: ../src/HtmlProcessor/CssVariableEvaluator.php
+
+		-
+			message: '#^Parameter \#2 \$callback of function preg_replace_callback expects callable\(array\<string\>\)\: string, array\{\$this\(Pelago\\Emogrifier\\HtmlProcessor\\CssVariableEvaluator\), ''getPropertyValueRep…''\} given\.$#'
+			identifier: argument.type
+			count: 1
+			path: ../src/HtmlProcessor/CssVariableEvaluator.php
+
```
JakeQZ, line 50: I recall now: `Closure::fromCallable` served two purposes when using the `Preg` utility class:
1. Allow a private method to be passed as the `callable` parameter;
2. Silence PHPStan warnings that an array comprising object instance (or class name) and method name was not a `callable`.

It seems that there are no PHPStan warnings passing a `Closure` when the parameter is defined simply as a `callable`.  But the internal and `Safe` functions define the parameter as `callable(array):string`, and PHPStan considers a `Closure` with the same type pattern not to be compatible.

I think there are two bugs in PHPStan:
1. `Closure(array):string` not considered suitable for `callable(array):string` parameter;
2. `array{0:object,1:string}` not considered to be a `callable`.

That the warnings are currently in the baseline should be sufficient to keep track of this issue.

JakeQZ (changes requested):
Apologies for requesting a further change.  I dug into the PHPStan warnings (again).

I think it would actually be better to keep `Closure::fromCallable()` to allow the method to stay `private`.  This will give a different PHPStan warning.  But as we're gaining a PHPStan warning anyway, we may as well keep private what should be private.

JakeQZ:
> Will look into logging issue(s) with PHPStan or seeing if there are any already existing...

I found https://github.com/phpstan/phpstan/issues/7544 where the callback had incorrect PHPDoc.  After [a bit of playing around](https://phpstan.org/r/53645ca3-184b-43ad-b77c-375d8e75199d), I think I found that the callback parameter needs to be `array<string>`, not `array<int, string>`.  It seems that the internal functions are not as tightly-typed as they should be.  Watch this space (or another, if it comes as a PR)...

JakeQZ:
> Watch this space (or another, if it comes as a PR)...

I found that using `array<array-key, string>` as the callback parameter type resolves the PHPStan warning for the internal function.

However, it will always be `list<string>`.  This is an issue with PHPStan and/or the PHP documentation, particularly if PHPStan lifts the types from the online PHP docs (which I know 'Safe' does).

The above is not sufficient not resolve the warning for 'Safe':

```cmd
 ------ ---------------------------------------------------------------------------------------------------------------
  Line   emogrifier\src\HtmlProcessor\CssVariableEvaluator.php
 ------ ---------------------------------------------------------------------------------------------------------------
  129    Parameter #2 $callback of function Safe\preg_replace_callback expects callable(array): string, Closure(array<
         string>): string given.
         🪪  argument.type
         💡  Type array<string> of parameter #1 $matches of passed callable needs to be same or wider than parameter ty 
         pe array<int|string, mixed> of accepting callable.
 ------ ---------------------------------------------------------------------------------------------------------------
```

The 'Safe' library has only determined the type to be a generic array.  I am sure that changing the parameter of the callback method to that will get rid of the warning, but don't think it is the approach to take.

We should not change our code to cater for bugs in third-party tools (unless absolutely necessary).  Rather, we should report the issues found.  That's for tomorrow...

(Note to self: Issues are with Safe, PHPStan, and PHP documentation.)

JakeQZ:
> However, it will always be `list<string>`. This is an issue with PHPStan and/or the PHP documentation, particularly if PHPStan lifts the types from the online PHP docs (which I know 'Safe' does).

The array syntax for a `callable` is all-of-a-sudden now working in the PHPStan Playground, though [it is still not working for a `Closure`](https://phpstan.org/r/9c2b141c-fc1e-443a-a96d-8618e35deefe).

JakeQZ:
> > However, it will always be `list<string>`. This is an issue with PHPStan and/or the PHP documentation, particularly if PHPStan lifts the types from the online PHP docs (which I know 'Safe' does).
> 
> The array syntax for a `callable` is all-of-a-sudden now working in the PHPStan Playground, though [it is still not working for a `Closure`](https://phpstan.org/r/9c2b141c-fc1e-443a-a96d-8618e35deefe).

https://github.com/phpstan/phpstan/issues/13824

PR #1504, opened 2025-11-13
[CLEANUP] Reformat the code with the PER-2 configuration

oliverklee:


src/Utilities/DeclarationBlockParser.php
```diff
@@ -82,7 +82,11 @@ public function parse(string $declarationBlock): array
         $properties = [];
         foreach ($declarations as $declaration) {
             $matches = [];
-            if (preg_match('/^(-?+[a-zA-Z_][a-zA-Z_0-9\\-]++|--[a-zA-Z_0-9\\-]++)\\s*+:\\s*(.++)$/s', \trim($declaration), $matches) === 0) {
+            if (preg_match(
+                '/^(-?+[a-zA-Z_][a-zA-Z_0-9\\-]++|--[a-zA-Z_0-9\\-]++)\\s*+:\\s*(.++)$/s',
```
JakeQZ, line 86: I've just noticed that the first part of this regex should probably have `*+` not `++` so that it can match a single character...

oliverklee:
> Did an autoformat remove the indentation in the DocBlocks? 

Well, kind of … it changed the indentation in _one_ line in the CSS in the DocBlock, and then I manually aligned the indentation of the other lines so that things were consistent again and autoformatting wouldn't change anything anymore. :see_no_evil:

src/Utilities/DeclarationBlockParser.php
oliverklee: Are you going to create the ticket for that, or should I do it?

JakeQZ: > Are you going to create the ticket for that, or should I do it?

I'll just create a PR for it now.  Planned to ealier but got bogged down with the Lint issue...

JakeQZ: I also note that the line length exceeded 120 characters.  We used to use PHP_CodeSniffer to check that.  See #1505.

JakeQZ: > I'll just create a PR for it now.

#1506

PR #1507, opened 2025-11-15
[TASK] Replace deprecated PHP-CS-Fixer rules

oliverklee:


JakeQZ (approved):
I had also noticed the deprecation warnings.  Not sure when the replacements were added, but as long as all CI checks pass, it should be fine.  There are already some development tools which don't run on the lowest supported PHP version of the package.

PR #1508, opened 2025-11-15
[FEATURE] Add line length code check and fixer

### LINKED ISSUE #1505: Add a CI check for line length
The 120 character limit was enforced by PHP_CodeSniffer, but that was removed from the toolset in #1253 because it was not well-maintained and did not support PER-2.

Perhaps an alternative can be found...

oliverklee:
Fixes #1505

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,8 @@
         "phpunit/phpunit": "9.6.29",
         "rawr/phpunit-data-provider": "3.3.1",
         "rector/rector": "1.2.10 || 2.2.2",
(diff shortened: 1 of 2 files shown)


JakeQZ (approved):
All arguments and properties look correct to me.  It doesn't seem clear what the default is if `extensions` is not specified, but we don't want `.css` or other test data files being processed.  The default for `colors` (coloured output) is disabled (which would be `<arg name="no-colors"/>` - i.e. the argument does not have a value).

PR #1509, opened 2025-11-16
[FEATURE] Add support for Symfony 8.0

oliverklee:


oliverklee:
I've verified that at least in the [PHP 8.5 with highest Composer dependencies CI job](https://github.com/MyIntervals/emogrifier/actions/runs/19730975319/job/56531797879?pr=1509), we're indeed using symfony/css-selector v8.0.0.

PR #1514, opened 2025-11-22
[TASK] Drop the now-unused `Preg` utility class

oliverklee:


JakeQZ (approved):
It provided similar functionality to the `Safe` library, though only for the PCRE functions.  It's better to use a well-maintained third-party library, so that we don't have to do the maintenance.

A couple of notes:
- It provided details of the last PCRE error upon failure, which `Safe` does not;
- It had a default option to fallback on failure, which would allow the system to carry on as best as possible, though potentially lead to problems being missed if the error log was not checked.

PR #1515, opened 2025-11-23
[TASK] Use safe preg functions in `CssInliner`

oliverklee:
This fixes a missing import. (Once we've added the `thecodingmachine/phpstan-safe-rule` package, PHPStan is going to warn us about these.)

Part of #919


JakeQZ (approved):
Missed in both development and review.

PR #1517, opened 2025-11-23
[TASK] Add PHPStan rules for Safe-PHP

oliverklee:
These PHPStan rules also add some type information that allow
us to drop quite a lot of redundant type checks.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -10,6 +10,7 @@ use Pelago\Emogrifier\Utilities\CssConcatenator;
 use Pelago\Emogrifier\Utilities\DeclarationBlockParser;
 use Symfony\Component\CssSelector\CssSelectorConverter;
 use Symfony\Component\CssSelector\Exception\ParseException;
+
 use function Safe\preg_match;
 use function Safe\preg_replace;
 use function Safe\preg_replace_callback;
@@ -898,7 +899,7 @@ final class CssInliner extends AbstractHtmlProcessor
     private function removeImportantAnnotationFromNodeInlineStyle(\DOMElement $node): void
     {
         $style = $node->getAttribute('style');
-        $inlineStyleDeclarations = (new DeclarationBlockParser())->parse((bool)$style ? $style : '');
+        $inlineStyleDeclarations = (new DeclarationBlockParser())->parse((bool) $style ? $style : '');
         /** @var array<string, string> $regularStyleDeclarations */
         $regularStyleDeclarations = [];
         /** @var array<string, string> $importantStyleDeclarations */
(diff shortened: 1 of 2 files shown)


src/Utilities/DeclarationBlockParser.php
```diff
@@ -90,12 +90,7 @@ public function parse(string $declarationBlock): array
                 continue;
             }
 
-            \assert(\is_array($matches));
             $propertyName = $matches[1];
-            if ($propertyName === '') {
-                // This cannot happen since the regular expression matches one or more characters.
-                throw new \UnexpectedValueException('An empty property name was encountered.', 1727046409);
-            }
```
JakeQZ, line 98: Has it really delved into the regex to determine this?  If so, that's impressive.

Noting a couple of other changes, it seems it does.

tests/Unit/Css/CssDocumentTest.php
```diff
@@ -529,7 +529,7 @@ public function discardsInvalidNonConditionalAtRuleNotInDebugMode(string $atRule
         $result = $subject->renderNonConditionalAtRules();
 
         preg_match('/@[\\w\\-]++/', $atRuleCss, $atAndRuleNameMatches);
-        self::assertIsArray($atAndRuleNameMatches);
+        self::assertArrayHasKey(0, $atAndRuleNameMatches);
```
JakeQZ, line 532: I assume from this that when `preg_match` does not find a match, it will nonetheless set the `$matches` argument (to an empty array).  (It's not clear from the documentation, and I have not empirically tested.)

config/phpstan-baseline.neon
```diff
@@ -42,12 +42,6 @@ parameters:
 			count: 1
 			path: ../src/HtmlProcessor/CssVariableEvaluator.php
 
-		-
-			message: '#^Parameter \#2 \$callback of function preg_replace_callback expects callable\(array\<string\>\)\: string, Closure\(array\<int, string\>\)\: string given\.$#'
-			identifier: argument.type
-			count: 1
-			path: ../src/HtmlProcessor/CssVariableEvaluator.php
-
```
JakeQZ, line 50: Looks like the `thecodingmachine/phpstan-safe-rule` library has a solution for https://github.com/phpstan/phpstan/issues/13824 when using the `Safe` functions.

JakeQZ (approved):
Good job.  Pleased to see some of the redundant `assert`s eliminated.  I'm guessing that there's a warning if they are left in.

oliverklee:
>  I'm guessing that there's a warning if they are left in.

Yes, exactly. All warnings were of the "this will always be true" or "this will always be false" type.

src/Utilities/DeclarationBlockParser.php
oliverklee: Yes, that's what I thought as well. :tada:

tests/Unit/Css/CssDocumentTest.php
oliverklee: Exactly.

Issue #1511, opened 2025-11-20
PCRE regex execution error PREG_BACKTRACK_LIMIT_ERROR


timboomerang:
Hello,

I use your CSS inliner library to generate PDF files from HTML.

Everything is working fine for small PDF (up to 20 pages), but when I try to apply the inline CSS on larger files (100 pages), I've got the `PCRE regex execution error PREG_BACKTRACK_LIMIT_ERROR` exception in `src/Utilities/Preg.php` on line 200.

It looks like the regex system uses a generated regex built with every matching rule which in my case has too many parameters.

Is there a way to parse the content chunk by chunk to avoid this issue?

oliverklee:
Hi @timboomerang, thanks for reporting this!

We're currently moving away from our handcrafted `Preg` wrapper class.

Could you please give the current `main` branch a try and report back whether the problem still occurs with that version? Thanks!

timboomerang:
Hi @oliverklee,

I just gave a try to the latest `main` with this code.

```php
<?php
$html = 'here is my very large html data';
$css = file_get_contents( 'style.css' );
$cssInliner   = CssInliner::fromHtml( $html )->inlineCss( $css );
$html = $cssInliner->render();
```

To give you an idea, the generated HTML before applying style roughly weights 50Mo.
It contains inlined image (base64) which are responsible for the most of the weigth.

oliverklee:
Hi @timboomerang, thanks for trying `main` out. Does the problem still occur with `main`?

JakeQZ:
Hi @timboomerang,

On the assumption that the issue still occurs with the latest `main`...

Are you able to provide a stack trace of the error, so that we can pinpoint the particular regular expression that is failing with your content?  (Some regexes can probably still be improved by using the [possessive quantifier](https://www.regular-expressions.info/catastrophic.html) to limit backtracking.)

Ideally, to reproduce, we'd like your 50Mb files, but they may be beyond upload limit allowed by GitHub.

timboomerang:
Hi,

Message ID: ***@***.***>

The issue still occurs on the latest main version.
I cannot share the exact content with you for privacy reason. What I can see from my tests is that when applying the same style to the content without the base64 encoded image the library can complete its task without issue.

At this stage, the content is roughly 1Mb.
As the style is the same, I assume that the generated regex is the same? The issue must come from the size of the final content.

I’ll try to get the full stack trace when I’ll get back to the office.

JakeQZ:
[...]

PR #1518, opened 2025-11-25
[BUGFIX] Support very large HTML files

JakeQZ:
Part of #1511.

This solves the main issue, but a few side issues have been found and noted, either in the original bug report, or as comments in the code for now.

Thus I will not close the OBR with this PR.

### DIFF
--- a/src/HtmlProcessor/AbstractHtmlProcessor.php
+++ b/src/HtmlProcessor/AbstractHtmlProcessor.php
@@ -341,7 +341,29 @@ abstract class AbstractHtmlProcessor
     private function hasContentTypeMetaTagInHead(string $html): bool
     {
         preg_match(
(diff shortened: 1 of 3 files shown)


src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -341,7 +341,29 @@ private function addContentTypeMetaTag(string $html): string
...
+                        # todo: this backtracks on `[^>]` which is not ideal
+                        meta(?=\\s)[^>]*\\shttp-equiv=(["\']?+)Content-Type\\g{-1}[\\s/>]
+                    )
+                )
+                # start of `subject`
+                ^
+                # one or other of these
+                (?:
+                    # one or more characters other than `<`
+                    [^<]++
```
JakeQZ, line 357: Has to be one or more to avoid infinite loop.

JakeQZ:
I checked the tests for addition of `Content-Type` or not, and they seemed satisfactory.

PR #1519, opened 2025-11-25
[TASK] Improve `Content-Type` matcher performance

JakeQZ:
Avoid unnecessary backtracking in the regular expression.


JakeQZ:
As an aside, I checked what tests we have for this, and they do seem to be quite complete.

PR #1524, opened 2025-11-28
[DEVELOPER] Correct and expand the release checklist

JakeQZ:
In the GitHub UI, it is not currently possible to create a 'tag' without at least creating a draft release -
see https://github.com/orgs/community/discussions/48462 - so it cannot be a separate pre-step.

Also add more guidance for newbies when filling in the release form.

### DIFF
--- a/docs/release-checklist.md
+++ b/docs/release-checklist.md
@@ -8,7 +8,11 @@
 1. Update the target milestone in the Dependabot configuration.
 1. Create a pull request "Prepare release of version x.y.z" with those changes.
 1. Have the pull request reviewed and merged.
-1. Tag the new release.
 1. In the [Releases tab](https://github.com/MyIntervals/emogrifier/releases),
-   create a new release and copy the change log entries to the new release.
+   click 'Draft a new release'.
+1. Select to create a new tag, e.g. `v8.2.0`, and the branch to use (or `main`).
+1. Give the release a title, e.g. "V8.2.0: New features and bug fixes".
+1. Copy the change log entries to the 'Release notes' box.
+1. Check or uncheck 'Set as the latest release' accordingly.
+1. Double-check everything is correct before clicking 'Publish release'.
 1. Post about the new release on social media.


JakeQZ:
I've used the '[DEVELOPER]' title prefix for a change that is developer-specific.  If you don't think that works, you're welcome to change it to '[TASK]'...

oliverklee:
> In the GitHub UI, it is not currently possible to create a 'tag' without at least creating a draft release -
see https://github.com/orgs/community/discussions/48462 - so it cannot be a separate pre-step.

I usually create the tags locally and then push the tags. This looks like this:

```bash
git tag -a v8.2.0 -m "Tag version 8.2.0"
git push --tags
```

I also usually sign the tag with GPG:

```bash
git tag -a -s v8.2.0 -m "Tag version 8.2.0"
git push --tags
```

So I don't think we need the extra step using the GitHub UI (which feels more clunky to me than using the command line, and which also does not allow a person to GPG-sign the release).

So I'd prefer to not mention the GitHub UI for this in the release checklist.

But maybe we can list how to create and push a tag on CLI?

WDYT?

JakeQZ:
> I usually create the tags locally and then push the tags.

[This suggestion](https://github.com/orgs/community/discussions/48462#discussioncomment-14367717) got 7 downvotes and just one upvote.  But that was on a bug report complaining that GitHub UI was inadequate and failed to resolve isues on a timely basis.

> a person to GPG-sign the release

I have no idea how this works.

> But maybe we can list how to create and push a tag on CLI?

This may be the way forward for now.  If only there was a UI that could make the job simpler...

oliverklee:
> [This suggestion](https://github.com/orgs/community/discussions/48462#discussioncomment-14367717) got 7 downvotes and just one upvote. But that was on a bug report complaining that GitHub UI was inadequate and failed to resolve isues on a timely basis.

Yes, that's how I understand the downvotes, too. (The command itself is perfectly fine - it just doesn't solve the original UI problem).

> If only there was a UI that could make the job simpler...

It seems like we feel differently about the Git CLI being simple. :wink:

JakeQZ:
> It seems like we feel differently about the Git CLI being simple. 😉

It's easy if you're familiar with it and know what to do.  It's less straightforward otherwise if you have to pore over the documentation to figure out precisely what command line options you need, and worry that if you get it wrong you'll muck things up irrevocably.  (E.g. commit messages on a protected branch like `main` can't easily be changed, and c.f. https://github.com/MyIntervals/PHP-CSS-Parser/pull/1220#discussion_r2016325885.)

> But maybe we can list how to create and push a tag on CLI?

I'll put this on hold for now, until the next time I create a new release (using the CLI) - I'd like to document the steps as I carry them out, to confirm the guidance is correct.

oliverklee:
On GPG-signing:

Git docs:
https://git-scm.com/book/ms/v2/Git-Tools-Signing-Your-Work

GitHub then will display the corresponding commits as verified (with the green bubble tag):
https://docs.github.com/en/authentication/managing-commit-signature-verification

oliverklee:
Should we close this in favor of #1532?

JakeQZ:
> Should we close this in favor of #1532?

Next time I create a release I will try the command line for creating the tag.  If I find any issues with the documented procedure, I can open a new ticket,  So, yes.

PR #1525, opened 2025-11-28
[CLEANUP] Improve type declarations and annotations in some tests

oliverklee:


oliverklee:
This PR includes the changes suggested by the latest Rector version.

tests/Unit/Support/Constraint/IsEquivalentCssTest.php
```diff
@@ -67,10 +60,7 @@ public function provideNonEquivalentCss(): array
         return $datasets + $transposedDatasets;
     }
 
-    /**
-     * @return Constraint
-     */
-    protected function createSubject(): Constraint
+    protected function createSubject(): IsEquivalentCss
```
JakeQZ, line 63: This is replacing a placeholder method in a trait which has the return type `Constraint`.

tests/Unit/Support/Constraint/StringContainsCssCountTest.php
```diff
@@ -164,10 +139,7 @@ public function notMatchesHaystackContainingExactlyOneNeedleWithCountOfTwo(strin
         self::assertFalse($result);
     }
 
-    /**
-     * @return Constraint
-     */
-    protected function createSubject(): Constraint
+    protected function createSubject(): StringContainsCssCount
```
JakeQZ, line 142: As is this.

tests/Unit/Support/Constraint/StringContainsCssTest.php
```diff
@@ -54,10 +47,7 @@ public function notMatchesHaystackNotContainingNeedle(string $needle, string $ha
         self::assertFalse($result);
     }
 
-    /**
-     * @return Constraint
-     */
-    protected function createSubject(): Constraint
+    protected function createSubject(): StringContainsCss
```
JakeQZ, line 50: And this.

JakeQZ (commented):
I'm not sure about changing the return type for `createSubject()`.

These methods replace a placeholder in a trait, which is called from methods in the trait.

Unlike C++, PHP does not support multiple-inheritance.  Traits allow for pseudo-multiple-inheritance.

If the language allowed full multiple-inheritance, we would not be allowed to make the return type in the subclass more specific.

On that basis, I don't think we should we should do so when reimplementing a method in a trait.

A future version of PHP may enforce that reimplented methods from traits must have a compatible signature (and IMO that should happen).

I'm open to arguments to the contrary... ;)

(I've not played around with Rector at all, so I don't know what it's saying.)

Other changes all look fine - they're just removing redundant DocBlock entries.

oliverklee:
I think the dirty part here is that in the classes, we're overwriting a method from the trait. I think I have seen warnings about this somewhere (in some PHP version, or with PHPStan - I don't remember exactly where).

This is something I'd like to get rid of.

I see two possible ways to solve this:

a) create an interface (and mark the trait as requiring that interface):

```php
protected function createSubject(): IsEquivalentCss;
```

b) In the trait, move the subject from the method `createSubject` to a `protected` property `subject`

I'd prefer b).

WDYT?

JakeQZ:
> a) create an interface (and mark the trait as requiring that interface):

That would work.
 
> ```
> protected function createSubject(): IsEquivalentCss;
> ```

Though it would need to be the more generic

```php
protected function createSubject(): Constraint;
```

> b) In the trait, move the subject from the method `createSubject` to a `protected` property `subject`

I'm not sure this would work.  Doesn't the subject need to be created afresh for each invokation of a test method (akin to `setUp` and `tearDown`) whilst the test class object remains?  So having it as a property would mean that if it is manipulated by one test method, it would be in a potentially different (and unexpected) state when another test method is called.  Or perhaps I haven't got a clear grasp on how PHPUnit works...

> I'd prefer b).
> 
> WDYT?

I'm verging on a), but not sure...

oliverklee:
> Or perhaps I haven't got a clear grasp on how PHPUnit works...

That's my guess in that case. :wink: 

PHPUnit creates a [new instance of the test class](https://speakerdeck.com/oliverklee/test-driven-development-with-phpunit-915a5f2c-3d37-4e41-b5bc-ef1efca9c01e?slide=21) for each test method run.

JakeQZ:
> PHPUnit creates a [new instance of the test class](https://speakerdeck.com/oliverklee/test-driven-development-with-phpunit-915a5f2c-3d37-4e41-b5bc-ef1efca9c01e?slide=21) for each test method run.

IIRC it doesn't create new instances for each dataset from a data provider, though does call `setUp()` and `tearDown()` before and after each call to the test method.  So if a data provider somehow depended on or manipulated state, results may be unexpected.  Anyway, the change in #1527 uses `setUp()` so that shouldn't be an issue.

oliverklee:
> IIRC it doesn't create new instances for each dataset from a data provider, 

Yes, it does: #1529

JakeQZ:
> > IIRC it doesn't create new instances for each dataset from a data provider,
> 
> Yes, it does: #1529

Thank you so much for taking the time to demonstrate the behaviour.  I could probably have determined it myself, rather than speculating, but was too lazy ;)).  I guess you already knew, whereas I didn't.  I feel educated now :)).  Really appreciated <3.

Back to the main topic: is this PR now superseded by #1526 and #1527, or is there anything else remaining here to merge?

oliverklee:
> Back to the main topic: is this PR now superseded by https://github.com/MyIntervals/emogrifier/pull/1526 and https://github.com/MyIntervals/emogrifier/pull/1527, or is there anything else remaining here to merge?

There's still some type cleanup. I'll create a dedicated PR for that right away.

PR #1543, opened 2026-01-03
[CLEANUP] Tighten type annotations in `CssConcatenatorTest`

oliverklee:
Also drop a comment that does not add anything useful.

### DIFF
--- a/tests/Unit/Utilities/CssConcatenatorTest.php
+++ b/tests/Unit/Utilities/CssConcatenatorTest.php
@@ -108,7 +108,7 @@ final class CssConcatenatorTest extends TestCase
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tests/Unit/Utilities/CssConcatenatorTest.php
```diff
@@ -111,7 +108,7 @@ public function appendInsertsSemicolonCombiningRulesWithoutTrailingSemicolon():
     }
 
     /**
-     * @return array<string, array<int, array<int, string>>>
+     * @return array<non-empty-string, array{0: non-empty-list<non-empty-string>, 1: non-empty-list<non-empty-string>}>
```
JakeQZ:
```suggestion
     * @return array<non-empty-string, array{0: non-empty-list<non-empty-string>, 1: non-empty-list<non-empty-string>, 2: non-empty-list<non-empty-string>}>
```

PR #1545, opened 2026-01-04
[TASK] Relax a pathologic testcase

oliverklee:
This test is failing on my machine, and fixing an opening HTML tag that is missing its closing angle bracket is not part of the functionality that our library promises.


JakeQZ (commented):
The difference in behaviour probably relates to the version of `libxml` that is installed.

It seems better that `<b` is not changed to `<b></b>` (whether it is removed or changed to `&lt;b`).

We can however still test that HTML containing such a broken tag is nonetheless parsed.  The second item in the dataset is something expected in the rendered HTML, so we could just check for `<body>` in this instance.

JakeQZ (approved):

PR #1546, opened 2026-01-04
[CLEANUP] Tighten type annotations in `CssInlinerTest`

oliverklee:


oliverklee:
I'm still trying to find out why PHP-CS-Fixer is crashing (or reporting an unspecific error) …

tests/Unit/CssInlinerTest.php
```diff
@@ -3461,7 +3505,7 @@ public function inlineCssNotInDebugModeKeepsInvalidOrUnrecognizedSelectorsInMedi
 
     /**
      * @return array<non-empty-string, array{before: string, '@import': non-empty-string, after: string}>
-     */
+    */
```
JakeQZ:
```suggestion
     */
```

tests/Unit/CssInlinerTest.php
```diff
@@ -1813,7 +1819,7 @@ public function orderedRulesAndSurroundingCssDataProvider(): DataProvider
             // broken: invalid @charset (unquoted value)
         ];
 
-        /** @var array<string, array<int, string>> $datasetsSurroundingCss */
+        /** @var array<non-empty-string, non-empty-list<string>> $datasetsSurroundingCss */
```
JakeQZ, line 1822: It may be possible to remove this - Stan should be able to work it out.

tests/Unit/CssInlinerTest.php
```diff
@@ -1987,7 +1993,7 @@ public function inlineCssKeepsExistingStyleElementWithMediaOutOfBody(): void
     /**
      * Valid media query which need to be preserved
      *
-     * @return string[][]
+     * @return array<non-empty-string, non-empty-list<non-empty-string>>
```
JakeQZ: Each dataset has just one string, so I think

```suggestion
     * @return array<non-empty-string, array{0: non-empty-string}>
```

tests/Unit/CssInlinerTest.php
```diff
@@ -2229,10 +2251,11 @@ public function inlineCssNotKeepsUnneededMediaRuleAfterEmptyMediaRule(string $em
     }
 
     /**
-     * @param array<string, string> $precedingSelectorComponents selectors to which each type of pseudo-component is
-     *        appended to create a selector for a CSS rule. Keys are human-readable descriptions.
+     * @param array<non-empty-string, string> $precedingSelectorComponents
+     *        selectors to which each type of pseudo-component is appended to create a selector for a CSS rule.
+     *        Keys are human-readable descriptions.
```
JakeQZ, line 2256: I don't like the non-sentence description followed by a full-stop and an actual sentence.  But that was there before :/

tests/Unit/CssInlinerTest.php
```diff
@@ -3132,7 +3171,9 @@ public function inlineCssNotInDebugModeIgnoresOnlyInvalidExcludedSelector(): voi
     /**
      * @test
      *
-     * @param array<non-empty-string> $excludedSelectors
+     * @param non-empty-string $css
+     * @param array{0: non-empty-string} $excludedSelectors
```
JakeQZ, line 3175: It seems like this was designed with the idea of being able to provide multiple selectors for exclusion, but none of the datasets have that.  Maybe `non-empty-list<non-empty-string>` would be more appropriate, and we should add a dataset with multiple selectors (as a separate PR).

tests/Unit/CssInlinerTest.php
```diff
@@ -3155,7 +3196,7 @@ public function addExcludedCssSelectorCanExcludeCssSelectors(
     }
 
     /**
-     * @return array<array{non-empty-string, array<non-empty-string>, non-empty-string}>
+     * @return array<non-empty-string, array{0: non-empty-string, 1: array{0: non-empty-string}, 2: non-empty-string}>
```
JakeQZ, line 3199: Ditto.

tests/Unit/CssInlinerTest.php
```diff
@@ -3744,7 +3800,7 @@ public function getMatchingUninlinableSelectorsThrowsExceptionIfInlineCssNotCall
     }
 
     /**
-     * @return string[][][]
+     * @return array<non-empty-string, array{0: array{0: non-empty-string, 1?: non-empty-string}}>
```
JakeQZ: I think `list` would be more appropriate here:


```suggestion
     * @return array<non-empty-string, array{0: non-empty-list<non-empty-string>}>
```

tests/Unit/CssInlinerTest.php
```diff
@@ -3757,7 +3813,7 @@ public function matchingUninlinableSelectorsDataProvider(): array
     /**
      * @test
      *
-     * @param string[] $selectors
+     * @param array{0: non-empty-string, 1?: non-empty-string} $selectors
```
JakeQZ: Ditto:


```suggestion
     * @param non-empty-list<non-empty-string> $selectors
```

oliverklee:
> Must have been some sloggish gruntwork to go through all this ;)

It was indeed … :exploding_head: At least that was the final file to clean up. After this PR has been merged, we'll have a sound foundation to hopefully fix the remaining PHPStan warnings.

I really appreciate you digging through this massive PR. :pray:

tests/Unit/CssInlinerTest.php
oliverklee: #1549

tests/Unit/CssInlinerTest.php
oliverklee: #1550

tests/Unit/CssInlinerTest.php
oliverklee: #1551

tests/Unit/CssInlinerTest.php
oliverklee: #1551

PR #1547, opened 2026-01-04
[CLEANUP] Split overlong line

oliverklee:


JakeQZ:
> We should run `php:codesniffer` in the GitHub action.

#1548 (and https://github.com/MyIntervals/PHP-CSS-Parser/pull/1460)

PR #1554, opened 2026-01-05
[CLEANUP] Move a comment to avoid confusing PHPStan

oliverklee:


JakeQZ (approved):
Strange.  It seems that warning disappears if either
- The comment is moved (to anywhere else in the DocBlock), or
- The comment is changed to not start with the word "Invalid".

I can't find any documentation on this curious behaviour.

Issue #1555, opened 2026-01-16
Support heavily nested documents


aikar:
Encountered an error where CssVariableEvaluator::evaluateVariablesInElementAndDescendants blew up for exceeding recursive stack trace depth.

Issue appears to stem due to the recursive call, so that if an HTML document exceeds 256 depth layers, it will fail.

This issue can be resolved by converting the recursion to an explicit tail call optimization

```php
    private function evaluateVariablesInElementAndDescendants(
        \DOMElement $element,
        array $ancestorVariableDefinitions
    ): self {
        $elementsToEvaluate = [$element];

        while (($element = array_pop($elementsToEvaluate)) != null) {
            $style = $element->getAttribute('style');

            // Avoid parsing declarations if none use or define a variable
            if (preg_match('/(?<![\\w\\-])--[\\w\\-]/', $style) !== 0) {
                $declarations = (new DeclarationBlockParser())->parse($style);
                $variableDefinitions =
                    $this->getVariableDefinitionsFromDeclarations($declarations) + $ancestorVariableDefinitions;
                $this->currentVariableDefinitions = $variableDefinitions;

                $newDeclarations = $this->replaceVariablesInDeclarations($declarations);
                if ($newDeclarations !== null) {
                    $element->setAttribute('style', $this->getDeclarationsAsString($newDeclarations));
                }
            } else {
                $variableDefinitions = $ancestorVariableDefinitions;
            }

            foreach ($element->childNodes as $child) {
                if ($child instanceof \DOMElement) {
                    $elementsToEvaluate[] = $child;
                }
            }
        }

        return $this;
    }
```

oliverklee:
@aikar Thanks! Would you be willing to provide a PR for this?

aikar:
> [@aikar](https://github.com/aikar) Thanks! Would you be willing to provide a PR for this?

Sorry I do not have the time to go through the process for a PR at the moment.

JakeQZ:
The suggested code change looks good; thank you for that <3.  (I'd prefer not to overwrite the `$element` argument, but instead have a differently-named local variable for the loop.)

> Sorry I do not have the time to go through the process for a PR at the moment.

No worries; we can work from what you have provided.

But I'd like to understand...

> blew up for exceeding recursive stack trace depth.

How exactly did it fail?  Was this when using a debugger?  Or what was the error?

> HTML document exceeds 256 depth layers

Why 256?  Might there be a setting somewhere that controls this?

[...]

PR #1559, opened 2026-01-18
[TASK] Run a CI check with Xdebug

JakeQZ:
Relates to #1555.

A couple of tests have been disabled, as they will now fail until the fix for #1555 is applied.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,3 +202,62 @@ jobs:
 
       - name: Run Tests
         run: composer check:tests:unit
+
+  unit-tests-xdebug:
+    name: Unit tests with Xdebug
+
+    runs-on: ubuntu-22.04
+
+    needs: [ php-lint, composer-validate ]
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
+          - '8.4'
+        dependencies:
+          - highest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -202,3 +202,59 @@ jobs:
...
+
+      - name: Install Composer dependencies
+        run: |
+          if [[ "${{ matrix.dependencies }}" == 'lowest' ]]; then
+            DEPENDENCIES='--prefer-lowest';
+          else
+            DEPENDENCIES='';
+          fi;
+          composer install --no-progress;
+          composer require xdebug/xdebug:*;
```
oliverklee: Those are some Composer packages. We'll want to have the PHP module, though.

.github/workflows/ci.yml
```diff
@@ -202,3 +202,59 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
+          ini-file: development
+          tools: composer:v2
+          coverage: none
```
oliverklee: We could try `coverage:xdebug`.

https://github.com/shivammathur/setup-php?tab=readme-ov-file#xdebug

JakeQZ: I tried it.  It does not seem to work.  The test should fail because of the stack depth issue, but it does not.

JakeQZ: Finally sussed it.  We need `ini-values: xdebug.mode=develop`.

JakeQZ:
I've kept the matrix in place in the copied action, even though in that it is merely 1x1, as I think it makes it more maintainable.

PR #1561, opened 2026-01-24
[TASK] Add some type checks in the tests

oliverklee:


tests/Unit/HtmlProcessor/AbstractHtmlProcessorTest.php
```diff
@@ -1264,7 +1265,9 @@ public function getDomDocumentWithNormalizedHtmlRepresentsTheGivenHtml(): void
 
         $domDocument = $subject->getDomDocument();
 
-        self::assertEqualsHtml($html, $domDocument->saveHTML());
+        $normalizedHtml = $domDocument->saveHTML();
+        self::assertIsString($normalizedHtml);
+        self::assertEqualsHtml($html, $normalizedHtml);
```
JakeQZ: I'm struggling to see what normalized could mean in this test, other than perhaps that the HTML provided is well-formed.  I think the test method name could be improved, to perhaps `getDomDocumentReturnsDomDocumentRepresentingTheHtmlProvidedToFromHtml`, but that's beyond the scope of this PR.

It seems that the 'normalized' HTML is what is provided to `fromHtml()`.  To avoid further confusion, I'd suggest a different name for the local variable:

```suggestion
        $resultHtml = $domDocument->saveHTML();
        self::assertIsString($resultHtml);
        self::assertEqualsHtml($html, $resultHtml);
```

PR #1601, opened 2026-03-17

oliverklee:


JakeQZ (approved):

JakeQZ:
I can't see any way to silence the new PHPStan warnings.  I think they may be due to PHPStan bugs introduced with the  update.  They don't even make any sense.  2.1.46 doesn't have these issues; 2.1.47 does.  I can't see anything in the release note indicative of a relevant change, though there are a couple of changes relating to type narrowing.

oliverklee:
Yes, those are two rules from [Rector TypePerfect](https://github.com/rectorphp/type-perfect) that somehow gets impacted by the PHPStan update. I think it would make sense for us to disable the `narrow_param` and `narrow_return` rulesets. I'll ask my working student Anna to do this.

PR #1603, opened 2026-04-11
[CLEANUP] Replace a condition with an assertion

oliverklee:
The way we construct HTML processors, this exception will never be thrown. As PHPStan does not know this, state this knowledge using an assertion.

### DIFF
--- a/src/HtmlProcessor/AbstractHtmlProcessor.php
+++ b/src/HtmlProcessor/AbstractHtmlProcessor.php
@@ -115,10 +115,7 @@ abstract class AbstractHtmlProcessor
      */
     public function getDomDocument(): \DOMDocument
     {
-        if (!$this->domDocument instanceof \DOMDocument) {
-            $message = self::class . '::setDomDocument() has not yet been called on ' . static::class;
-            throw new \UnexpectedValueException($message, 1570472239);
-        }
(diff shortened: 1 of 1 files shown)


JakeQZ (approved):
The conditional exception throwing was to guard against any programmer error by maintainers/contributors of this library.  An `assert()` is a more suitable guard for this (and may also be in other similar cases), as it will compile to nothing in a production environment, saving a few CPU cycles.

As the constructor is final and private, and the class is abstract, the only (currently) possible means of instantiating objects of this class are by using `::fromHtml()` or `::fromDomDocument()`.  If either of these could instantiate without setting the `$domDocument` property, we would expect some tests to fail:
- Before this change, with an exception being thrown;
- After this change, with an assertion failure (trusting that the CI checks are actually run with development settings).

Double-checking the construction methods by inspection to confirm they always set `$domDocument`:
- [x] `::fromHtml()`
  - Always calls `::setHtml()` -> `::createUnifiedDomDocument()` -> `::createRawDomDocument()` -> `::setDomDocument()` with `new \DOMDocument()`, unless an exception is thrown, in which case it won't complete and won't return an object (with or without a `$domDocument`);
- [x] `::fromDomDocument()`
  - Has a non-nullable `DOMDocument` parameter which is immediately assigned to `$domDocument`

All boxes ticked :)

PR #1606, opened 2026-04-14
[CLEANUP] Move PHPUnit config file to `Build/`

AnReiter:


oliverklee (changes requested):
We'll also need to add `/Build/` to `.gitattributes` so it doesn't get packaged.

The file needs to stay sorted. In PhpStorm: Edit > Sort line

oliverklee (changes requested):
We'll also need to drop the `phpunit.xml` line from `.gitattributes`.

oliverklee (approved):
LGTM. Thanks!

.gitattributes
```diff
@@ -3,9 +3,9 @@
 /.github/ export-ignore
 /.gitignore export-ignore
 /.phive/ export-ignore
+/Build/
```
JakeQZ, line 6: Does this need `export-ignore`?

JakeQZ (commented):
Hi @AnReiter, welcome to the project :).

I think `export-ignore` might be needed for the `/Build/` directory to exclude it from the release package...

.gitattributes
oliverklee: Indeed it does, and I missed this during review. I've created a follow-up PR: #1607

PR #1609, opened 2026-04-15
[TASK] Also check `Build/` with static code analysis

oliverklee:
We now have PHP files in `Build/`, and we should check and autofix those as well.

We omit these two tools:

- PHPStan (as it cannot find the PHP-CS-Fixer classes from the PHAR)
- PHPMD (as it only can check one directory)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -93,9 +93,9 @@
         "check:dynamic": [
             "@check:tests"
         ],
-        "check:php:codesniffer": "phpcs --standard=./Build/phpcs/config.xml config src tests",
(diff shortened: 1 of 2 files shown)


JakeQZ (approved):
> * PHPMD (as it only can check one directory)

Actually it can check multiple directories, from a comma-separated list on the command line, e.g. `phpmd src,tests text config/phpmd.xml`.  We were previously only checking `src`; expanding the checking should be the scope of a separate PR.

oliverklee:
> expanding the checking should be the scope of a separate PR.

Ah, I see. Thanks!

I think we don't need to - as far as code metrics are concerned, I think those are only relevant for the production code.

PR #1613, opened 2026-04-16
[CLEANUP] Move Rector configuration to `Build/`

AnReiter:


oliverklee (changes requested):
This is the last PHP file from `config/`. So we also need to drop `config` from all Composer scripts that do something with PHP code, and also from all tool config files that check PHP code and that reference `config`.

Build/rector/config.php
```diff
@@ -10,10 +10,10 @@
 return RectorConfig::configure()
     ->withPaths(
         [
-            __DIR__ . '/../Build',
-            __DIR__ . '/../config',
-            __DIR__ . '/../src',
-            __DIR__ . '/../tests',
+            __DIR__ . '/../../Build',
+            __DIR__ . '/../../config',
```
oliverklee: We can drop this line as well.

oliverklee (changes requested):

Build/rector/config.php
AnReiter: no, when i do composer check, terminal says: "[ERROR] The path "/emogrifier/Build/rector/../Build" does not exist.  "

AnReiter: aha! it was about config

oliverklee (approved):

PR #1615, opened 2026-04-19
[TASK] Disable some rules from Rector Type Perfect

oliverklee:
The two type-narrowing rules tend to have too many false positives at the moment.


JakeQZ (approved):
c.f. https://github.com/MyIntervals/PHP-CSS-Parser/pull/1575#pullrequestreview-4136296256

PR #1616, opened 2026-04-19
[CLEANUP] Replace some exceptions with assertions

oliverklee:
These type-checking exceptions will never be thrown. So convert these checks into assertions stating what the types are.

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -415,16 +415,12 @@ final class CssInliner extends AbstractHtmlProcessor
      * Returns a list with all DOM nodes that have a style attribute.
      *
      * @return \DOMNodeList<\DOMElement>
-     *
-     * @throws \RuntimeException
      */
     private function getAllNodesWithStyleAttribute(): \DOMNodeList
     {
         $query = '//*[@style]';
         $matches = $this->getXPath()->query($query);
-        if (!$matches instanceof \DOMNodeList) {
-            throw new \RuntimeException('XPatch query failed: ' . $query, 1618577797);
-        }
+        \assert($matches instanceof \DOMNodeList);
         /** @var \DOMNodeList<\DOMElement> $matches */
 
         return $matches;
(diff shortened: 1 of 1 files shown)


oliverklee:
Marking as draft. We should get #1615 in first.

src/CssInliner.php
```diff
@@ -529,10 +525,7 @@ private function querySelectorAll(string $selectors, array $options = []): \DOMN
     {
         try {
             $result = $this->getXPath()->query($this->getCssSelectorConverter()->toXPath($selectors));
-
-            if ($result === false) {
-                throw new \RuntimeException('query failed with selector \'' . $selectors . '\'', 1726533051);
-            }
+            \assert($result instanceof \DOMNodeList);
             /** @var \DOMNodeList<\DOMElement> $result */
```
JakeQZ: We should remove this annotation as part of this change.

JakeQZ (changes requested):
I checked that `XPath::query()` would only return `false` if the query string is malformed, which should never happen unless there is a bug in 'CssSelectorConverter'.  Thus `assert()` is more appropriate.

#1615 is now merged.

But there's an annotation that should be removed as part of this change.

src/CssInliner.php
oliverklee: No, that's necessary as we have `@return \DOMNodeList<\DOMElement>` and hence need the part within the `<>`.

src/CssInliner.php
JakeQZ: We should also remove this (second) `@throws` annotation, since a `RuntimeException` will no longer be thrown.

JakeQZ: This (second) `catch` block should be removed, since a `RuntimeException` will no longer be thrown.

JakeQZ (changes requested):
Apologies that my original review was not thorough (partly because the PR was in 'draft' status at the time).

Since we'll no longer throw a `RuntimeException`, we should remove the DocBlock annotation and `catch` block for it.

JakeQZ (approved):
Neat.

PR #1618, opened 2026-04-26
[TASK] Drop some unnecessary type checks

oliverklee:
The DOM nodes in the touched code parts are already guaranteed to be `DOMElement`s, and PHPStan can already deduce this. (So we don't need to have `assert`s for this.)

### DIFF
--- a/src/CssInliner.php
+++ b/src/CssInliner.php
@@ -163,8 +163,6 @@ final class CssInliner extends AbstractHtmlProcessor
      * @throws \RuntimeException
      *         in debug mode, if an internal PCRE error occurs
      *         or `CssSelectorConverter::toXPath` returns an invalid XPath expression
-     * @throws \UnexpectedValueException
-     *         if a selector query result includes a node which is not a `DOMElement`
      */
     public function inlineCss(string $css = ''): self
     {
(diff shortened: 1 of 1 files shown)


src/CssInliner.php
```diff
@@ -496,15 +494,12 @@ private function getCssFromAllStyleNodes(): string
...
-     * @throws \UnexpectedValueException if the selector query result includes a node which is not a `DOMElement`
      */
     private function getNodesToExclude(): array
     {
         $excludedNodes = [];
         foreach (\array_keys($this->excludedSelectors) as $selectorToExclude) {
-            foreach ($this->querySelectorAll($selectorToExclude) as $node) {
-                $excludedNodes[] = $this->ensureNodeIsElement($node);
-            }
+            \array_push($excludedNodes, ...$this->querySelectorAll($selectorToExclude));
```
JakeQZ, line 502: Neat.

JakeQZ (approved):
`assert()` rules.

I came from C/C++, where we always used `assert()`.  Maybe 10 years ago PHP's incarnation didn't work quite the same (or I failed to research properly).  But it seems perfect now.

src/CssInliner.php
oliverklee: Now that I finally (think that I) understand the unpack/splat operator, I'm starting to quite like it. :slightly_smiling_face:

PR #1619, opened 2026-04-26
[CLEANUP] Autoformat the code

oliverklee:
This avoids unrelated changes when someone changes the code and then autoformats it with PhpStorm.

### DIFF
--- a/Build/php-cs-fixer/config.php
+++ b/Build/php-cs-fixer/config.php
@@ -97,5 +97,5 @@ return (new Config())
             // string notation
             'single_quote' => true,
             'string_implicit_backslashes' => ['single_quoted' => 'escape'],
-        ]
+        ],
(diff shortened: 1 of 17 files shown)


JakeQZ (approved):
After some brief research, I think I am very much sold on having the [trailing comma in function calls](https://wiki.php.net/rfc/trailing-comma-function-calls) - if an additional argument needs to be added in future, it will simplify the diffs.

The other changes are whitespace only.  I don't know how they ended up in the mess they were, but it's probably my fault, using Windows ;)).

JakeQZ:
> The other changes are whitespace only. I don't know how they ended up in the mess they were, but it's probably my fault, using Windows ;)).

Actually, they're just different indentation levels that don't match `.editorconfig`.

PR #1621, opened 2026-05-07
[CLEANUP] Make `AbstractHtmlProcessor::domDocument` private

oliverklee:
Also make it non-nullable as it will never be read before initialization.

Access to the property already is guarded by the `getDomDocument()` accessor, which we'll need to keep as we access it from the outside.

We cannot guard the getter's code with an `assert` as PHPStan would complain about a useless assert there. So rely on PHP throwing an error if the getter is called before the property is initialized.

Also improve some related comments.

### DIFF
--- a/src/HtmlProcessor/AbstractHtmlProcessor.php
+++ b/src/HtmlProcessor/AbstractHtmlProcessor.php
@@ -45,9 +45,9 @@ abstract class AbstractHtmlProcessor
         = '%<template[\\s>][^<]*+(?:<(?!/template>)[^<]*+)*+(?:</template>|$)%i';
 
     /**
-     * @var \DOMDocument|null
+     * @var \DOMDocument
      */
-    protected $domDocument = null;
+    private $domDocument;
 
     /**
      * @var \DOMXPath|null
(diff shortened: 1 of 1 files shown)


src/HtmlProcessor/AbstractHtmlProcessor.php
```diff
@@ -453,9 +451,7 @@ private function ensurePhpUnrecognizedSelfClosingTagsAreXml(string $html): strin
     }
 
     /**
-     * Checks that $this->domDocument has a BODY element and adds it if it is missing.
-     *
-     * @throws \UnexpectedValueException
```
JakeQZ, line 458: `getHtmlElement()` currently throws this exception if there is no `<html>` element, which would only be the case if `loadHTML()` failed.  But [the documentation](https://www.php.net/manual/en/domdocument.loadhtml.php#refsect1-domdocument.loadhtml-returnvalues) doesn't indicate under which circumstances it would fail, merely stating "malformed HTML should load successfully".

So perhaps we can assume that there will always be an `<html>` element, and change `getHtmlElement()` to `assert()` so as a separate PR.  (Also, calling methods aren't currently declared with this `@throws` annotation.)

JakeQZ (approved):

JakeQZ:
> We cannot guard the getter's code with an `assert` as PHPStan would complain about a useless assert there.

We could set [`treatPhpDocTypesAsCertain`](https://phpstan.org/config-reference#treatphpdoctypesascertain) to `false` in `phpstan.neon` to allow us to `assert()` complete construction.  WDYT?

oliverklee:
> We could set [`treatPhpDocTypesAsCertain`](https://phpstan.org/config-reference#treatphpdoctypesascertain) to `false` in `phpstan.neon` to allow us to `assert()` complete construction.  WDYT?

Same as https://github.com/MyIntervals/emogrifier/pull/1620#issuecomment-4413149677.

PR #1629, opened 2026-06-16
[BUGFIX] Add missing `use` in PHPStan extension test

JakeQZ:
PHPStan is only run on a recent PHP version, so the extension only needs testing against that version.  Older versions of PHPStan (which support older PHP versions) do not support the extension.  But the unit tests are run on all supported PHP versions, so those for the extension need to be skipped on versions for which it is not supported.  This is done via an `interface_exists()` check.

But without the `use` statement, that check was resulting in the test being skipped for all versions.


oliverklee (approved):

JakeQZ:
Ref: #1624, the change adding the PHPStan extension.

PR #1631, opened 2026-06-21

oliverklee:


JakeQZ (approved):
Looks like the newer `thecodingmachine/phpstan-safe-rule` package is now able to determine when `preg_replace()` cannot return an empty string.  (I did a quick check by changing the replacement to an empty string, and confirmed that the warning is not generated in that case.)  So we can remove the redundant `assert()`s as a separate PR.

PR #1633, opened 2026-06-28
[CLEANUP] Drop some redundant `assert`s

oliverklee:


JakeQZ (approved):
Follows on from https://github.com/MyIntervals/emogrifier/pull/1631#pullrequestreview-4573624789.

PR #1638, opened 2026-07-10
[TASK] Do not scan the PHPStan unit tests with PHPStan

oliverklee:
This is broken with newer versions of PHPStan.


JakeQZ (approved):
We can re-enable analysis when we drop support for PHP<7.4.

PR #1641, opened 2026-08-12
[TASK] Update PHPStan

oliverklee:


JakeQZ (changes requested):
With the recent changes (#1646 and #1647) to keep the PHPStan extension working, we can and should retain the supposedly pointless asserts.  They are not pointless because the DocBlock alone does not enforce type safety - PHP itself completely ignores it.

JakeQZ (approved):

PR #1642, opened 2026-08-12
[TASK] Update to Rector 2.6.2

oliverklee:
https://github.com/rectorphp/rector/releases/tag/2.6.2

### DIFF
--- a/Build/rector/config.php
+++ b/Build/rector/config.php
@@ -3,8 +3,6 @@
 declare(strict_types=1);
 
 use Rector\Config\RectorConfig;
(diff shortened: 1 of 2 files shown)


JakeQZ (commented):
I think the `RectorConfig` changes should be in a separate PR from the version increment - surely they are not interdependent...

JakeQZ (approved):
> At least the removal of the PHPUnit set needs to be in the same PR as the class got removed in that Rector version.

Oh, OK.  Not ideal for what should be a bugfix-only release.

PR #1646, opened 2026-08-26
[TASK] Make PHPStan extension compatible with PHPStan 2.2.8

JakeQZ:
The extension merely silences warnings about impossible `instanceof` assertions, on a global basis, avoiding the need to clutter the code with `phpstan-ignore` comments.

But if PHPStan's behaviour keeps changing with every point release, the extension may need to be dropped in favour of cluttering the code with `phpstan-ignore` comments.

In PHPStan 2.2.8, the applicable `Node` parameter passed to `shouldIgnore()` is now a `FunctionCallExpressionNode`, wrapping `FuncCall` that was passed in previous versions.


oliverklee (changes requested):
I think we'll need to combine this with the PHPStan update in one PR to avoid the "class not found/unknown class" PHPStan warnings.

oliverklee (approved):

